PR #2, opened 2026-05-16
fix(flow/record_scope): handle gRPC envelopes in capture_scope url_prefix/method (USK-909)

usk6666:
## Summary
- Add `*envelope.GRPCStartMessage` arm to `extractScopeFields` so capture_scope `url_prefix` and `method` rules apply to native gRPC and gRPC-Web envelopes (previously silently no-op).
- Wire-faithful projection: synthesize `:path = "/<Service>/<Method>"` and `:method = "POST"` per gRPC HTTP/2 transport spec. Guarded against malformed `:path` (Service or Method empty) so partial paths are not synthesized.
- Add `grpcStartEnv` test helper and tests covering URLPrefix and Method rules for both gRPC and gRPC-Web.

## Test plan
- [x] `make lint` passes
- [x] `make build` passes
- [x] `make test` passes
- [x] `TestRecordScope_URLPrefix_GRPC` covers positive and negative service match
- [x] `TestRecordScope_Method_GRPC` covers POST match and non-POST non-match
- [x] `TestRecordScope_URLPrefix_GRPCWeb` covers gRPC-Web envelope
- [x] Pre-existing `TestRecordScope_NonHTTP_URLPrefixRule_DoesNotMatch` continues to pass

## Background
Discovered during gRPC Phase 1 manual testing (test case GRPC-P1-05). Design review confirmed: live bug, all 6 MITM Implementation Principles upheld, single case arm covers both gRPC and gRPC-Web because both layers emit the same envelope type.

Resolves USK-909
Linear: https://linear.app/usk6666/issue/USK-909

### DIFF
--- a/internal/flow/record_scope.go
+++ b/internal/flow/record_scope.go
@@ -186,6 +186,18 @@ func extractScopeFields(env *envelope.Envelope) (method, hostname, path string) 
 				hostname = h
 			}
 		}
+	case *envelope.GRPCStartMessage:
+		// gRPC over HTTP/2 wire-literally carries `:method = POST` and
+		// `:path = /<Service>/<Method>`. Exposing these to capture_scope
+		// is wire-faithful, not normalization (MITM Principle 1). Both
+		// native gRPC and gRPC-Web emit *envelope.GRPCStartMessage. The
+		// 3-field guard suppresses synthesis when a malformed `:path`
+		// envelope left Service or Method empty, so a partial path like
+		// "/Service/" cannot spuriously prefix-match a rule.
+		if m != nil && m.Service != "" && m.Method != "" {
+			path = "/" + m.Service + "/" + m.Method
+			method = "POST"
+		}
 	}
 	return method, hostname, path
 }
--- a/internal/flow/record_scope_test.go
+++ b/internal/flow/record_scope_test.go
@@ -19,6 +19,25 @@ func httpEnv(method, authority, path string) *envelope.Envelope {
 	}
 }
 
+// grpcStartEnv builds a Direction=Send GRPCStartMessage envelope shaped
+// like what the gRPC / gRPC-Web Layers emit. The proto argument selects
+// envelope.ProtocolGRPC or envelope.ProtocolGRPCWeb — both layers produce
+// the same Message type, so a single capture_scope arm must cover both.
+func grpcStartEnv(t *testing.T, proto envelope.Protocol, authority, service, method string) *envelope.Envelope {
+	t.Helper()
+	return &envelope.Envelope{
+		Protocol:  proto,
+		Direction: envelope.Send,
+		Message: &envelope.GRPCStartMessage{
+			Service: service,
+			Method:  method,
+		},
+		Context: envelope.EnvelopeContext{
+			TargetHost: authority,
+		},
+	}
+}
+
 func nonHTTPEnvWithCtx(targetHost string, sni string) *envelope.Envelope {
 	env := &envelope.Envelope{
 		Protocol: envelope.ProtocolWebSocket,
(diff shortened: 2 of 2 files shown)


usk6666 (commented):
## Code Review: APPROVED

The PR is a minimal, surgically-scoped fix for a live bug (USK-909). The diff adds a single `*envelope.GRPCStartMessage` arm to `extractScopeFields` so `capture_scope.url_prefix` and `.method` rules apply to native gRPC and gRPC-Web envelopes — closing the silent no-op documented in the Issue.

**Correctness**

- Wire-faithful projection: `:path = "/<Service>/<Method>"` and `:method = "POST"` match what HTTP/2 actually carries on a gRPC stream (RFC-001 §3.2.3, gRPC HTTP/2 transport spec). This is observation, not normalization — MITM Principle 1 upheld.
- The 3-field guard (`m != nil && m.Service != "" && m.Method != ""`) correctly suppresses synthesis when a malformed `:path` envelope left either field empty. Without this guard, a `/Service/` rule could spuriously prefix-match a partial path. The new tests exercise both the Service-empty and Method-empty branches.
- Single case arm covers both native gRPC and gRPC-Web because both Layers emit `*envelope.GRPCStartMessage` (verified in `internal/layer/grpc/channel.go:1037` and `internal/layer/grpcweb/channel.go:331`). No duplication.
- Decision propagation: `decisionCache` in `internal/pipeline/record_step.go` keys by StreamID and evaluates on `Direction=Send && Sequence=0`. The first envelope on a gRPC stream is `GRPCStartMessage`, so the cached decision correctly carries through downstream `GRPCDataMessage`/`GRPCEndMessage` envelopes.

**Architecture compliance**

- Matches the existing `*envelope.HTTPMessage` arm pattern directly above it (nil-check, set local fields, fall back to context for hostname when absent).
- No new dependencies, no exported API change, no `net/http` types in the data path.
- Test helper `grpcStartEnv(t, proto, authority, service, method)` accepts a `Protocol` arg so one helper drives both `ProtocolGRPC` and `ProtocolGRPCWeb` cases — deliberate parameterization, not duplication.

**Test quality**

- New tests cover: positive match, negative match (different service), empty-Service guard, empty-Method guard, POST method match, non-POST non-match, gRPC-Web parity. All `t.Error` messages name the expectation.
- Pre-existing `TestRecordScope_NonHTTP_URLPrefixRule_DoesNotMatch` (WS envelope) continues to pass — the new arm narrowly handles GRPCStartMessage only, no regression to other non-HTTP envelopes.

**Code health**

- Comment explicitly cites MITM Principle 1 and explains the 3-field guard rationale — saves future readers (and reviewers) the same investigation.
- No dead code, no TODOs, no magic numbers.

FINDINGS: None.

---

usk6666 (commented):
## Security Review: APPROVED

The PR is a pure field-extraction change in `internal/flow/record_scope.go` with no new network parsing, no goroutines, no TLS/cert handling, no `unsafe`, no `crypto/rand`, no SQL, and no new dependencies. The attack surface is narrow.

**TLS / Certificates**: N/A — no TLS code touched.

**Network**: N/A — no listener / dial / timeout / SSRF surface touched.

**Input validation**

- Inputs (`Service`, `Method`) come from `*envelope.GRPCStartMessage`, which is populated upstream by the gRPC parser at `internal/layer/grpc/channel.go`. The values are already attacker-controlled by design (a MITM proxy processes untrusted traffic), and the consumer in this PR uses them only for `strings.HasPrefix` / `strings.EqualFold` matching — no format-string injection, no SQL, no shell, no CRLF.
- The synthesized path `"/" + Service + "/" + Method` is not emitted to the wire; it is consumed only by the in-process matcher. No CRLF / header-injection vector.
- The 3-field guard (`m != nil && m.Service != "" && m.Method != ""`) prevents an attacker from crafting a malformed gRPC `:path` (e.g., empty Method) to spuriously match an `add_excludes:[{url_prefix:"/Service/"}]` rule and bypass recording. This is correct defensive posture — better than the original Issue body's `m.Service != ""` minimum.

**Go security**

- No concurrency added. RecordScope's existing `sync.RWMutex` already protects the rule slices.
- No goroutine leak risk — `extractScopeFields` is a pure synchronous function.
- No slice-bounds risk — string concatenation only; no indexing.
- `gofmt` / `go vet` clean; `go test -race ./internal/flow/...` passes.

**MCP / API**: N/A — control-plane untouched.

**Dependencies / Licenses**: No new dependencies.

FINDINGS: None.

---

PR #6, opened 2026-05-17
feat(proxybuild,connector): L7 TCP forward — http2 (h2c) + gRPC dispatch (USK-914)

usk6666:
## Summary
- BuildConnectionStackWithTarget HTTP2 + GRPC arms; clones BuildPlainH2CStack shape (SETTINGS_ENABLE_CONNECT_PROTOCOL mirror, ServerRole MaxConcurrentStreams)
- New file `internal/proxybuild/tcp_forward_h2.go` owns `runTCPForwardH2Loop` + `tcpForwardH2SessionOpts` (sibling to existing `tcpForwardSessionOpts`)
- Per-stream dispatch via `DispatchH2StreamFull` (native gRPC / gRPC-Web / HTTPMessage routing)
- Per-stream wire-record callbacks for `h2-frame` / `grpc-lpm-frame` / `grpc-h2-frame` parity with MITM path
- `Protocol="grpc"` filter: content-type validator on first envelope; mismatch → RST_STREAM(REFUSED_STREAM) + error flow with `Tags["forward_protocol_mismatch"]`
- gRPC-Web rejected under `Protocol="grpc"` (crisper semantic; gRPC-Web routes via `Protocol="http2"`)
- h2c preface read deadline added (defence vs slow-handshake DoS — the MITM path gets this via the inner-byte-peek timeout; forward callers need explicit floor)
- e2e: `tcp_forward_h2_integration_test.go` (HEADERS/DATA, 10 parallel streams with recording isolation) + `tcp_forward_grpc_integration_test.go` (unary, server-stream, bidi, filter reject)
- Race-flake checklist applied; OnComplete classification reused unchanged

## Out of scope
- `TLS=true` (USK-915), `UpstreamTLS=true` (USK-916), gRPC-Web explicit selector, h2 pool sharing, graceful drain

## Note: parallel with USK-913
USK-913 owns http / ws / sse / auto arms (same files). This PR will need to rebase on USK-913 before merge. Conflict surface is bounded — only the dispatch switches in `BuildConnectionStackWithTarget` and `startTCPForwardListener`.

## Test plan
- [x] `make lint`
- [x] `make build`
- [x] `make test`
- [x] `make test-e2e-smoke` (no regression)
- [x] `go test -tags e2e -race -v -run 'TestProxybuild_TCPForward' ./internal/proxybuild/...`
- [x] Concurrent-stream test run 5x under -race (zero flakes)
- [x] Existing `TestProxybuild_TCPForward_RawBytesFlowRecording` still passes (raw path regression check)
- [x] Existing `TestH2C_BasicRoundtrip` + `TestMultipleConcurrentStreams_RecordingIsolation` still pass (MITM h2 regression check)

Resolves USK-914
Linear: https://linear.app/usk6666/issue/USK-914

### DIFF
--- a/internal/connector/h2_dispatch.go
+++ b/internal/connector/h2_dispatch.go
@@ -294,6 +294,27 @@ func extractContentType(evt *http2.H2HeadersEvent) string {
 	return ""
 }
 
+// IsGRPCContentType is the exported wrapper for isGRPCContentType. It
+// reports whether the given Content-Type value is a native-gRPC media
+// type (application/grpc[+suffix]; does NOT match application/grpc-web*).
+// Used by forward callers (USK-914 proxybuild.runTCPForwardH2Loop) that
+// must apply a Protocol="grpc" content-type filter on the first envelope
+// of each accepted stream — the filter has to fire BEFORE
+// DispatchH2StreamFull (which consumes the peek), so an exported helper
+// is needed rather than re-using the dispatcher's inline check.
+func IsGRPCContentType(ct string) bool {
+	return isGRPCContentType(ct)
+}
+
+// ExtractH2ContentType is the exported wrapper for extractContentType.
+// Returns the first content-type header value found on evt (case-
+// insensitive name match), or "" if none is present. Used by forward
+// callers that peek the first envelope of an h2 stream before
+// dispatching — see IsGRPCContentType for the use case.
+func ExtractH2ContentType(evt *http2.H2HeadersEvent) string {
+	return extractContentType(evt)
+}
+
 // isGRPCContentType reports whether ct is a native-gRPC media type:
 // exactly application/grpc, or application/grpc with a structured-syntax
 // suffix (application/grpc+proto, application/grpc+json, ...). Parameters
--- a/internal/connector/stack_builder_target_override.go
+++ b/internal/connector/stack_builder_target_override.go
@@ -5,11 +5,13 @@ import (
 	"crypto/tls"
 	"fmt"
 	"net"
+	"time"
 
 	"github.com/google/uuid"
 
 	"github.com/usk6666/yorishiro-proxy/internal/envelope"
 	"github.com/usk6666/yorishiro-proxy/internal/layer/bytechunk"
+	"github.com/usk6666/yorishiro-proxy/internal/layer/http2"
 )
 
 // ForwardProtocol selects the L7 wire shape a caller of
@@ -43,12 +45,19 @@ const (
 	// USK-913 (L7 dispatch); currently returns a "not yet wired" error.
 	ForwardProtocolHTTP ForwardProtocol = "http"
 
-	// ForwardProtocolHTTP2 requests an h2c stack. Reserved for USK-913;
-	// currently returns a "not yet wired" error.
+	// ForwardProtocolHTTP2 requests an h2c stack. Wired by USK-914 —
+	// builds [http2 ServerRole client → http2 ClientRole upstream] with
+	// scheme="http" and SETTINGS_ENABLE_CONNECT_PROTOCOL mirrored from
+	// upstream. Both gRPC and gRPC-Web auto-route via DispatchH2Stream at
+	// the per-stream proxybuild dispatch layer (no filter).
 	ForwardProtocolHTTP2 ForwardProtocol = "http2"
 
-	// ForwardProtocolGRPC requests an h2c + gRPC stack. Reserved for
-	// USK-914; currently returns a "not yet wired" error.
+	// ForwardProtocolGRPC requests an h2c + gRPC content-type filter
+	// stack. Wired by USK-914 — assembles the same h2c stack as
+	// ForwardProtocolHTTP2; the per-stream dispatch layer additionally
+	// validates each stream's content-type against application/grpc[+...]
+	// and rejects non-gRPC traffic (including gRPC-Web) with
+	// RST_STREAM(REFUSED_STREAM).
 	ForwardProtocolGRPC ForwardProtocol = "grpc"
 
 	// ForwardProtocolWebSocket requests an HTTP/1.x stack that may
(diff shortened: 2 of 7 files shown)


usk6666 (commented):
## Code Review: APPROVED with LOW findings

The implementation faithfully mirrors `BuildPlainH2CStack` (the CONNECT-inner h2c sibling) and `buildOnHTTP2Stack` (the live h2 dispatcher), making the diff easy to validate against the proven precedents. Per-stream dispatch, wire-record callback wiring, race-flake checklist, and ctx-cancel watcher all match the established patterns. Lint, build, and `make test` all pass.

### Findings

| F-1 | MEDIUM | internal/proxybuild/tcp_forward_h2.go | 402-433 | Architecture | `tcpForwardH2SessionOpts` does not install `OnPipelineDrop = buildPipelineDropRecorder(...)`. The raw-forward sibling (`tcpForwardSessionOpts`, tcp_forward.go:431) does. Pipeline-Drops (scope / safety / intercept BlockedBy attributions, USK-782) on h2 forward traffic will not produce audit Streams, creating a recording-fidelity gap relative to the raw path. Add the same `buildPipelineDropRecorder` wiring (with the per-conn `blockedStreamSet`, and the eviction in the `OnComplete` path) so audit attribution is symmetric across forward modes. |
| F-2 | LOW | internal/proxybuild/tcp_forward_h2.go | 368-392 | Correctness | `recordGRPCFilterReject` always stamps `Protocol="grpc"` on the rejected Stream, even when the rejected stream was actually gRPC-Web. The `mismatchKind` tag does carry the distinction, but readers filtering on Protocol cannot distinguish "client sent non-gRPC under grpc filter" from "client sent grpc-web under grpc filter". Consider passing the observed content-type or a `rejectedProtocol` discriminant to record more faithfully (e.g. `Protocol="grpc-web"` when `IsGRPCWebContentType(ct)` matched). Minor — current behaviour is consistent and traceable via the tag. |
| F-3 | NIT | internal/proxybuild/tcp_forward_h2.go | 374 | CodeHealth | `streamFlowCtx` in `recordGRPCFilterReject` is consumed only for `ConnID`. Either reduce the parameter to `connID string` for clarity, or document that the wider type is reserved for future fields (Scheme, ClientAddr) the reject Stream might want to carry. |
| F-4 | NIT | internal/proxybuild/tcp_forward_h2.go | 102 | Architecture | The `default` arm of the `switch fc.Protocol` in `handleTCPForwardH2Conn` (line 98-104) is unreachable per the upstream gate in `startTCPForwardListener`. The comment correctly calls it a defensive guard, but `slog.Warn` here is appropriate only if it can fire; consider downgrading to a code comment + early return, or to `slog.Error` (since reaching it indicates a logic invariant violation). |

### MITM Compatibility

I applied the MITM Compatibility Triage. None of the findings touch header normalisation, dedup, Host=URL enforcement, or any other rejected pattern. The h2c forward stack preserves wire shape (`scheme="http"`, `SETTINGS_ENABLE_CONNECT_PROTOCOL` mirror) and the gRPC content-type filter is an opt-in selector behaviour that does not mutate forwarded traffic.

### Strengths worth calling out

- The h2c preface 30s read deadline (`targetOverrideH2CPrefaceReadDeadline`) closes a real slow-handshake DoS gap that the MITM path gets via the CONNECT inner-byte peek. Good catch.
- The `replayChannel` wrapper is the right minimal abstraction for the peek-then-replay shape the filter needs; the single-use semantic is correctly enforced under `consumeMu`.
- Per-stream wire-record callbacks (h2-frame / grpc-lpm-frame / grpc-h2-frame / grpc-web-base64) achieve recording parity with `buildOnHTTP2Stack` verbatim, so L4-capable principle holds.
- `OnComplete` classification reuses the `connCtx.Err()` gate from the raw sibling, correctly distinguishing listener-shutdown (`complete`) from upstream-RST (`error`).
- Test coverage is strong: HEADERS/DATA basic, 10-stream concurrent recording isolation, gRPC unary/server-stream/bidi, plus a filter-reject e2e that asserts the `forward_protocol_mismatch` tag.

---

usk6666 (commented):
## Security Review: APPROVED

The PR adds an h2c forward stack and per-stream gRPC content-type filter. Both are defensive additions that close existing gaps (slow-handshake DoS on raw forward h2c; non-gRPC traffic under a gRPC-tagged forward port) rather than introduce attack surface. No CRITICAL / HIGH / MEDIUM exploitable findings.

### Threat-model coverage

- The 30s preface read deadline (`targetOverrideH2CPrefaceReadDeadline`) bounds the per-connection handshake window, matching the protection the MITM path already gets via the CONNECT inner-byte peek timeout. Mitigates Slowloris-style abuse of the new forward-h2c path.
- The gRPC content-type filter rejects non-gRPC streams with `RST_STREAM(REFUSED_STREAM)` and records an error Stream tagged `forward_protocol_mismatch`. The rejection message ("tcp forward grpc filter: non-gRPC content-type rejected") is local-only — the wire only sees the error code, no information leak.
- Upstream dial honours `EffectiveUpstreamProxyForCtx`, so per-listener upstream-proxy overrides (USK-826) reach the new path without bypass.

### Low-severity observations (NOT blocking)

| S-1 | LOW | internal/proxybuild/tcp_forward_h2.go | 316 | GoSecurity | CWE-400 | `applyGRPCFilter` calls `ch.Next(ctx)` with the connection context — a malicious client could open many HEADERS-less h2 streams to pin per-stream goroutines (the per-stream replayChannel + dispatchForwardH2Stream goroutine pair). The per-forward `maxConns` cap on TCP connections bounds the total but not stream count per connection. The Issue body explicitly defers stream-limit to a follow-up, so this is documented design; surfacing as LOW so the follow-up Issue can pick it up. The watcher does close the stack on ctx-cancel, which unblocks Next, so the resource is bounded by listener lifetime. |
| S-2 | LOW | internal/proxybuild/tcp_forward_h2.go | 234 | GoSecurity | CWE-362 | Wire-record callbacks (`lpmOpt`, `grpcH2FrameOpt`, `aggH2FrameOpt`, `grpcWebBase64Opt`) are wired AFTER `applyGRPCFilter` consumes the first envelope. The HEADERS frame consumed by the filter is NOT recorded via the per-stream wire callbacks. This is acceptable in the accept path because the post-filter DispatchH2StreamFull picks up the replayed envelope and the standard h2 layer recording captures the wire bytes upstream. In the reject path, no Flow rows are written (Stream skeleton only). L4-capable principle has a small gap here, but the rejected stream's wire bytes are by design not forwarded. Mitigation if you wanted to close this fully: emit a single `RawMessage`-style audit Flow for the rejected HEADERS. Optional. |

### Dependencies / Licenses

No new external dependencies added. `google.golang.org/grpc` + `golang.org/x/net/http2` are already in the module (test-only use in new files). No license concerns.

### TLS / Cert

PR is h2c-only. No TLS surface added; `TLS=true` and `UpstreamTLS=true` are explicitly gated to USK-915 / USK-916 with new error returns. No CA-key handling changes.

### MITM-specific exception confirmation

The PR does not propose any header dedup, Host=URL enforcement, header canonicalisation, or `net/http` types in the data path. Wire fidelity is preserved.

---

PR #8, opened 2026-05-17
feat(proxybuild,connector): TCP forward — client-side TLS terminate + cert.Issuer integration (USK-915)

usk6666:
## Summary

- New `internal/proxybuild/tcp_forward_tls.go` adds client-side TLS termination for TCP forward listeners when `ForwardConfig.TLS=true`. The handshake is wrapped via `tlslayer.Server` + an SNI-honoring per-listener `*tls.Config`; dispatch to the H1 / H2 arm is driven by the negotiated ALPN (not the operator-declared protocol, which only governs the ALPN advertise list and the per-arm post-dispatch filter).
- `alpnOffersForForwardProtocol` maps the operator-declared protocol to the ALPN advertise list: `auto→[h2, http/1.1]`, `http→[http/1.1]`, `http2→[h2]`, `grpc→[h2]`, `websocket→[http/1.1]`, `sse→[http/1.1]`, `raw→nil`.
- Server cert issuance honors wire SNI (`mitmServerConfigForForward` builds a `*tls.Config` with a `GetCertificate` callback that lowercases the `ClientHelloInfo.ServerName` and falls back to the operator-declared `Target` hostname when SNI is absent). The Issuer's cert LRU is shared, so per-hostname cert reuse is intact.
- SNI ≠ Target host emits one `slog.Warn` per connection (no dedup), post-handshake.
- `TargetOverrideParams` gains `Scheme` + `ClientTLSSnapshot` fields. Both H1 (`buildTargetOverrideHTTPStack`) and H2 (`buildTargetOverrideH2CStack`) builders thread them through to the underlying Layer options so TLS-terminated forwards record `Scheme="https"` and stamp `EnvelopeContext.TLS` for plugin / recording parity with the live MITM CONNECT path.
- Defensive listener-start guard rejects `tls=true` when no CA Issuer is configured (`startTCPForwardListener`); the per-conn handler additionally fails-soft into `buildTLSStackBuildErrorRecorder` for CA regen races.
- TLS handshake failures are recorded as `state="error"` Streams with `FailureReason="client_tls_error"` via the existing `buildTLSStackBuildErrorRecorder` (wrapped with `connector.ErrClientTLSMITMHandshake` — same classification as the live MITM CONNECT branch, USK-858). This covers `no_application_protocol` ALPN rejection, CA regen races, and any underlying `tlslayer.Server` failure.
- `(tls, on_handshake, side=\"server\")` plugin hook fires post-handshake for parity with the live MITM CONNECT path's `connector.fireTLSHandshakeHook` site.
- A `forwardConnOverride` value type carries per-connection rewrites from the TLS handler to the downstream H1 / H2 dispatch arms so the listener-owned `*tcpForwardEntry` is never copied (which would copy its embedded `sync.WaitGroup` + atomic counters).

## Out of scope (separate Issues)
- Upstream-side TLS dial (`UpstreamTLS=true`) — USK-916
- MCP `proxy_start` schema / docs / RFC-001 §3.3 / WebUI Settings update — USK-917
- Recording the server-side `TLSSnapshot` into `Stream.ConnInfo` — deferred per design review (the `(tls, on_handshake)` plugin hook + the per-Layer `EnvelopeContext.TLS` stamp is the parity surface)

## Test plan
- [x] `make lint` clean (gofmt + go vet + staticcheck + ineffassign + gocyclo)
- [x] `make build` succeeds
- [x] `make test` passes (unit tier, including the 4 new unit tests in `tcp_forward_tls_test.go`)
- [x] `make test-e2e-smoke` clean (smoke tier — new e2e file is exhaustive, `//go:build e2e && !e2e_smoke`)
- [x] `make test-e2e` passes (full tier — 7 new e2e cells in `tcp_forward_tls_integration_test.go`)
- [ ] Manual: start a listener with `{\"443\": {\"target\": \"127.0.0.1:8443\", \"protocol\": \"auto\", \"tls\": true}}`, hit it with `curl -k --http2` and `curl -k --http1.1`, confirm both work and recorded Streams show `Scheme=\"https\"`

## Behaviour note: Go crypto/tls ALPN-fallback special-case

Go's `crypto/tls.negotiateALPN` has a deliberate special-case fallback for (`server=\"h2\"`, `client=\"http/1.1\"`) — the handshake succeeds with empty negotiated ALPN instead of sending `no_application_protocol` (Go issue 46310). The `TestProxybuild_TCPForward_TLS_HTTP2_AlpnMismatch_Rejected` test exercises the real `no_application_protocol` path by offering an unknown synthetic protocol (`unknown/9.9`) with no overlap; the (h2, http/1.1) combo gracefully falls through to the H1 dispatch arm by design (a comment in `chooseH1ForwardProtocol` documents the exact behaviour).

Resolves USK-915
Linear: https://linear.app/usk6666/issue/USK-915


usk6666 (commented):
## Code Review: APPROVED

The PR cleanly wires client-side TLS termination for TCP forward listeners using the existing tlslayer + cert.Issuer + plugin hook seams. The TargetOverrideParams threading (Scheme + ClientTLSSnapshot) into both H1 and H2 builders is well-scoped and additive. The fail-soft handshake-error recording reuses ErrClientTLSMITMHandshake for FailureReason="client_tls_error" parity with the live MITM CONNECT path. Tests cover all 7 dispatch cells (TLS+http1, TLS+h2, auto→h1/h2, ALPN reject, WSS, SNI mismatch) plus 8 unit cells. `make vet` and `go build ./...` succeed.

### LOW Findings (not blocking)

| F-1 | LOW | internal/proxybuild/tcp_forward.go | 113-118 | CodeHealth | `forwardConnOverride.fc` field is set in `dispatchForwardTLSByALPN` (both h2 and h1 branches) but never read anywhere downstream. The h2 handler receives `&fcOverride` as a separate `fc` arg; the h1 handler reads only `override.protocol` / `override.tlsTerminated` / `override.clientTLSSnapshot`. The field's godoc claims "fc overrides entry.fc so the gRPC content-type filter sees the correct Protocol declaration" but `applyGRPCFilter` reads content-type from the channel envelope, not from `override.fc`. Suggest: drop the field entirely, or actually use it where the doc claims it's used. |
| F-2 | LOW | internal/proxybuild/tcp_forward_tls.go | 99-103 | CodeHealth | `mitmServerConfigForForward` sets `ClientSessionCache: tls.NewLRUClientSessionCache(64)` on a *server-side* `*tls.Config`. `ClientSessionCache` only governs the client side of TLS session resumption (the local cache a tls.Client maintains for its own sessions) — on a tls.Server config it is unused. The comment ("ClientSessionCache mirrors what crypto/tls auto-initialises on MITMServerConfig") is also incorrect: `cert.Issuer.MITMServerConfig` (internal/cert/issuer.go:197) does NOT set this field. Server-side session ticket resumption is governed by `SessionTicketsDisabled` (default false) and an auto-generated ticket key, both of which work without this field. Suggest: remove the field + comment, or replace with a comment explaining that server-side ticket key is auto-managed. |
| F-3 | NIT | internal/proxybuild/tcp_forward_tls.go | 286-301 | CodeHealth | `dispatchForwardTLSByALPN` passes `&fcOverride` as both the `fc` argument to `handleTCPForwardH2ConnWithOverride` AND embedded inside the `override` value. Given F-1 (override.fc is dead), the redundancy is moot once F-1 is fixed; flagged as a NIT because the duplication makes the call site harder to read than it needs to be. |

### Suggestions

For F-1 + F-3: collapse the duplication by removing `forwardConnOverride.fc`. The h2 handler signature already takes `fc *config.ForwardConfig` explicitly, so the override only needs to carry (protocol, tlsTerminated, clientTLSSnapshot).

For F-2: replace the misleading comment + dead `ClientSessionCache` line with a one-line note that server-side ticket keys are auto-generated by crypto/tls and persist for the lifetime of the *tls.Config.

---

usk6666 (commented):
## Security Review: APPROVED

The PR adds client-side TLS termination on the TCP forward path. TLS posture is consistent with the live MITM CONNECT path: MinVersion=TLS1.2 on the proxy-as-server side (CWE-757), SNI normalized via `strings.ToLower(strings.TrimSpace(...))` for cert-cache key consistency (CWE-770), defensive copy of NextProtos before tls.Config retention, listener-start guard rejects fc.TLS=true with nil Issuer, per-conn defensive guard for CA regen race, handshake error wrapped with the canonical `ErrClientTLSMITMHandshake` sentinel and recorded via the existing buildTLSStackBuildErrorRecorder. Plugin hook (tls, on_handshake, side="server") fires post-handshake for parity with the live CONNECT path. tlslayer.Server uses HandshakeContext so ctx-cancel terminates slow handshakes (no DoS via half-open TLS).

### LOW Findings

| S-1 | LOW | internal/proxybuild/tcp_forward_tls.go | 204, 213 | Network | n/a | The forward-path TLS handshake error recorder is constructed with `nil` scope (`buildTLSStackBuildErrorRecorder(parentStack.FlowStore, nil, parentListenerName, connLogger)`), bypassing capture_scope filtering. The live MITM CONNECT path threads `deps.RecordScope` (internal/proxybuild/builder.go:378) so out-of-scope hosts (HSTS-pinned services the browser dials but the operator didn't include in scope) get filtered out at slog.Debug instead of recorded. Forward path always records, regardless of capture_scope. Not exploitable — failure modes are bounded by the LRU and the rate at which the listener accepts conns; an out-of-scope leak is operationally noisy at worst. The fix would require Stack to expose RecordScope or threading it through the Manager; defer until the rest of the forward path needs scope filtering. |

### Threat Model Notes

- **CA key**: not touched by this PR — issuer.GetCertificate closure captures the existing *Issuer; no new CA-key surface area.
- **SNI**: any SNI value is accepted and triggers a per-hostname cert issuance via the existing LRU. This is the intended pentest-MITM behaviour; CWE-770 is already handled by the bounded LRU.
- **ALPN advertise list**: operator-declared per ForwardConfig.Protocol — clients with no overlap correctly trigger `no_application_protocol` (RFC 7301 §3.2) and are recorded as state="error" with FailureReason="client_tls_error".
- **Fail-soft for CA regen race**: defensive nil-Issuer check at per-conn handshake path is correct; the error is wrapped with ErrClientTLSMITMHandshake so the recorder classifies it consistently.
- **Goroutine safety**: entry.tlsServerCfg is written at listener start (single goroutine, before accept loop spawn) and read concurrently by per-conn handlers — safe via happens-before. mitmServerConfigForForward closure captures immutable values (fallback string, *Issuer pointer); cert.Issuer.GetCertificate is documented thread-safe.

No CRITICAL / HIGH / MEDIUM findings.

---

PR #5, opened 2026-05-17
feat(connector): stack builder を target-override 化して TCP forward から再利用可能にする

usk6666:
## Summary

- New `BuildConnectionStackWithTarget` in `internal/connector/` — target-fixed sibling of `BuildConnectionStack`. Takes a `TargetOverrideParams` struct (Target, Protocol, ALPNOffers, TLSTerminate, UpstreamTLS).
- New `ForwardProtocol` enum (auto / raw / http / http2 / grpc / websocket / sse) mirrors `config.ForwardConfig.Protocol`.
- `ForwardProtocolRaw` is fully wired (bytechunk client + bytechunk upstream). Every other selector and both TLS flags return a clear "not yet wired" error that cites the responsible follow-up Issue (USK-913 / USK-914 / USK-915 / USK-916), locking the API surface today.
- New `PerformClientMITM` and `DialUpstreamWithALPN` exported wrappers for the existing private helpers. Minimum-viable export scope — `ResolvePerHostTLS`, `ALPNRoute`, `BuildStackFromRoute` stay private until USK-913/914/915 actually need them.
- godoc on every exported surface documents the operator-declared-target contract: target fixed by caller, no SNI derivation, caller-supplied ALPN advertise list (empty = no ALPN extension on the wire), callers retain dial control.
- No changes to `handleTCPForwardConn`. The new builder is stand-alone with a unit test exercising it, per design review decision U4. L7 callsite wiring belongs to USK-913+.

## Design decisions (from Phase 1.5 design review)

- U1 — Naming: `BuildConnectionStackWithTarget` (caller-agnostic).
- U2 — ALPN API: `[]string` verbatim offers, matching `dialUpstreamWithALPN`.
- U3 — Helper export scope: minimum-viable (PerformClientMITM + DialUpstreamWithALPN only).
- U4 — Callsite shape: stand-alone with unit test, `handleTCPForwardConn` unchanged.

`BuildConnectionStack` signature is unchanged; `BuildPlainHTTPStack` / `BuildPlainH2CStack` / `BuildBytechunkStack` are unchanged.

## Test plan

- [x] `make lint` green (gofmt, go vet, staticcheck, ineffassign, gocyclo)
- [x] `make build` green
- [x] `make test` green — all existing tests pass, new tests pass
[...]

PR #4, opened 2026-05-17
feat(config): ForwardConfig に UpstreamTLS 追加 + TLS/UpstreamTLS 意味論の確定

usk6666:
## Summary
- Add `UpstreamTLS bool` (`json:"upstream_tls,omitempty"`) to `ForwardConfig` (foundation #1 of the L7 TCP forward re-implementation split).
- Lock semantics in godoc + MCP help: `TLS` = client-side termination, `UpstreamTLS` = upstream-dial encryption. Independent fields; never inferred from `Target` scheme/port. Both default to `false`, preserving the existing raw-forward behavior (USK-711).
- MCP control-plane wiring: `parseTCPForwardsAny` extracts `upstream_tls`; `proxyStartInput` jsonschema, `schema_proxy_start.json`, and `help_proxy_start.md` document the new field + the 4-combination matrix.
- `validateTCPForwardsConfig` Warns on `UpstreamTLS=true && Protocol="raw"` (mirroring the existing TLS+raw Warn) and emits a per-entry Info log whenever `TLS=true` explaining the meaning shift ("TLS now means client-side termination only; for upstream TLS use upstream_tls=true").
- Data-path wiring (upstream-TLS-aware dial) is deliberately deferred to USK-916; `internal/proxybuild/tcp_forward.go` remains untouched and continues to runtime-reject non-raw + TLS exactly as USK-711 left it.

## TLS × UpstreamTLS matrix
| tls   | upstream_tls | Client wire | Upstream wire | Use case                                      |
|-------|--------------|-------------|---------------|-----------------------------------------------|
| false | false        | plaintext   | plaintext     | Raw L4/L7 forwarding (default)                |
| true  | false        | TLS         | plaintext     | TLS termination only (downstream offload)     |
| false | true         | plaintext   | TLS           | Plaintext client → TLS-only upstream          |
| true  | true         | TLS         | TLS           | Full MITM: terminate then re-encrypt upstream |

## Test plan
- [x] `make lint` (gofmt + go vet + staticcheck + ineffassign + gocyclo)
- [x] `make build`
- [x] `make test` — all packages pass
- [x] New `TestForwardConfig_UnmarshalJSON_TLSUpstreamTLSMatrix` (4 cases)
- [x] New `TestForwardConfig_MarshalJSON_OmitEmptyDefaults`
- [x] New `TestForwardConfig_MarshalJSON_RoundTrip_AllCombinations` (4 cases)
- [x] Extended `TestValidateForwardConfig` with upstream_tls cases
[...]

PR #9, opened 2026-05-17
feat(proxybuild,connector): TCP forward — UpstreamTLS dial + ALPN propagation (USK-916)

usk6666:
## Summary
- proxybuild: new `dialForwardUpstream` + `upstreamALPNOffersForForward` helpers in `tcp_forward_tls.go` centralise the upstream dial (plain or TLS) for both the H1/raw/L7 arm and the H2 arm.
- proxybuild: H1 and H2 forward handlers perform upstream TLS handshake via `DialUpstreamRaw` when `fc.UpstreamTLS=true`; the `(tls, on_handshake, side="client")` plugin hook fires on success for parity with the live MITM dial path.
- proxybuild: remove the listener-start `UpstreamTLS=true not yet supported` reject; the field is wired end-to-end through the per-conn handler.
- proxybuild: upstream TLS dial failures (cert verify, handshake timeout, etc.) are recorded via `buildTLSStackBuildErrorRecorder` so `state="error"` Streams with `FailureReason="upstream_tls_error"` surface via MCP queries.
- connector: `TargetOverrideParams.UpstreamTLSSnapshot` stamps the upstream Layer's `EnvelopeContext.TLS` in HTTPStack + H2CStack + AutoStack. `UpstreamTLS=true` is now informational rather than a hard reject (existing unit test updated to match).

## ALPN propagation policy (Decision #4 / #5)
- Operator-declared Protocol → derive via `alpnOffersForForwardProtocol`.
- `"auto"` + client TLS terminated + client-negotiated ALPN → propagate the client's choice as a single-element list.
- `"auto"` + plaintext client (or no client ALPN) → offer only `["http/1.1"]` to preserve the single-protocol-per-stack invariant. Operators wanting h2 upstream MUST declare `Protocol="http2"` explicitly. **This refines the Issue body's stated `[h2, http/1.1]` per the design review record** — the proxy's Auto plaintext-peek path routes only to H1 or Raw (no h2c opportunistic upgrade), so advertising h2 upstream would create an unsupported h1↔h2 bridge.

## e2e test coverage (new file `tcp_forward_upstream_tls_integration_test.go`, exhaustive tier)
- 4 (TLS × UpstreamTLS) combinations across H1 and H2: HTTP→HTTPS, HTTPS→HTTPS, h2c→h2-TLS, h2-TLS→h2-TLS.
- 4 ALPN-policy cells: explicit `http2`, auto+client-h2, auto+client-h1, auto+plaintext-client.
- Upstream cert verification failure (`InsecureSkipVerify=false` + self-signed cert) → `state="error"` Stream with `FailureReason="upstream_tls_error"`.
- Plugin `(tls, on_handshake, side="client")` hook firing verified via Starlark `print()` sentinel captured through `testutil.CaptureLogger`.

## Test plan
- [x] `make lint` passes (gofmt + go vet + staticcheck + ineffassign + gocyclo)
- [x] `make build` passes
- [x] `make test` (fast tier) passes
- [x] `make test-e2e-smoke` (merge gate) passes
- [x] Full e2e tier (`go test -tags 'e2e' ./internal/proxybuild/... ./internal/connector/...`) passes including new test file
- [x] CLAUDE.md e2e Subsystem Verification Checklist items addressed for the new file (communication success, stream/flow recording, state transitions, plugin hook firing, error path, MCP-queryable failure reason)

## Reviewer focus points
1. **Decision #5 refinement** — the helper godoc on `upstreamALPNOffersForForward` documents the divergence from the Issue body; please confirm the rationale (single-protocol-per-stack invariant) lands cleanly.
2. **Seam discipline** — the builder now treats `UpstreamTLS` as informational and stamps the snapshot from `UpstreamTLSSnapshot`. TLS is performed in `proxybuild.dialForwardUpstream` (caller-side) mirroring the USK-915 client-side terminate seam.
3. **Error classification** — upstream TLS dial errors use the existing `buildTLSStackBuildErrorRecorder` (no new sentinel); non-`ErrClientTLSMITMHandshake` errors map to `FailureReason="upstream_tls_error"`. Verified via the cert-verify-failure e2e cell.
4. **No mTLS / uTLS / HostTLSRegistry changes** — per Issue scope, deferred. `DialRawOpts{InsecureSkipVerify, OfferALPN, TLSConfig, DialTimeout}` only; per-host TLS material is not consulted on the forward path.

Resolves USK-916
Linear: https://linear.app/usk6666/issue/USK-916

### DIFF
--- a/internal/connector/stack_builder_target_override.go
+++ b/internal/connector/stack_builder_target_override.go
@@ -148,11 +148,34 @@ type TargetOverrideParams struct {
 	// CLAUDE.md MITM Principle 2.
 	TLSTerminate bool
 
-	// UpstreamTLS, when true, indicates the caller wants the upstream
-	// dial to perform a TLS handshake. The upstream-TLS dial wire-up is
-	// owned by USK-916 and is NOT implemented by USK-912; setting
-	// UpstreamTLS=true currently returns a "not yet wired" error.
+	// UpstreamTLS, when true, indicates the caller dialed the upstream
+	// connection over TLS before invoking this builder. USK-916 changed
+	// the field semantics from "request TLS dial" to "TLS has happened
+	// upstream of this builder": the caller (proxybuild.dialForwardUpstream)
+	// owns the dial primitive (DialUpstreamRaw with a populated TLSConfig)
+	// and passes the negotiated *envelope.TLSSnapshot via
+	// UpstreamTLSSnapshot. Mirrors the seam discipline used by USK-915
+	// for client-side TLS terminate (TLS happens in the per-conn handler;
+	// the builder consumes the resulting snapshot).
+	//
+	// Setting UpstreamTLS=true without populating UpstreamTLSSnapshot is
+	// permitted (used by callers that elect to stamp the upstream Layer's
+	// EnvelopeContext.TLS later via a different mechanism) but yields the
+	// same lossy state as the cleartext path — the recorded envelopes
+	// will not reflect upstream TLS reality. The proxybuild forward
+	// handlers always populate the snapshot.
 	UpstreamTLS bool
+
+	// UpstreamTLSSnapshot, when non-nil, is plumbed into the upstream-side
+	// EnvelopeContext.TLS field of the assembled L7 Layers so plugin hooks
+	// and recordings observe the negotiated SNI / ALPN / cipher for the
+	// upstream TLS handshake. USK-916: forward callers that performed the
+	// upstream TLS dial themselves (proxybuild.dialForwardUpstream) own
+	// this field; callers that left UpstreamTLS=false leave it nil. Mirrors
+	// the per-Layer upstream TLS wiring in stack_builder.go::buildStackFromRoute
+	// (live MITM CONNECT path), preserving CLAUDE.md MITM Principle 3
+	// (lossless representations).
+	UpstreamTLSSnapshot *envelope.TLSSnapshot
 
 	// Scheme overrides the envelope Scheme stamped by the assembled L7
 	// Layers. Defaults to "http" when unset, matching the cleartext
@@ -294,9 +317,11 @@ func BuildConnectionStackWithTarget(
 	if params.TLSTerminate {
 		return nil, fmt.Errorf("connector: BuildConnectionStackWithTarget: TLSTerminate=true not yet wired (deferred to USK-915)")
 	}
-	if params.UpstreamTLS {
-		return nil, fmt.Errorf("connector: BuildConnectionStackWithTarget: UpstreamTLS=true not yet wired (deferred to USK-916)")
-	}
+	// USK-916: params.UpstreamTLS is now informational only — TLS is
+	// performed caller-side via proxybuild.dialForwardUpstream (which
+	// invokes connector.DialUpstreamRaw with a populated TLSConfig). The
+	// builder consumes UpstreamTLSSnapshot to stamp upstream Layer's
+	// EnvelopeContext.TLS. No reject gate.
 
 	protocol := params.Protocol
 	if protocol == "" {
@@ -312,18 +337,18 @@ func BuildConnectionStackWithTarget(
 		// proxybuild handler layer (it must observe the first request /
 		// first response, which is per-exchange Pipeline territory rather
 		// than per-stack-builder territory).
-		return buildTargetOverrideHTTPStack(ctx, clientConn, upstreamConn, params.Target, params.effectiveScheme(), params.ClientTLSSnapshot, cfg)
+		return buildTargetOverrideHTTPStack(ctx, clientConn, upstreamConn, params.Target, params.effectiveScheme(), params.ClientTLSSnapshot, params.UpstreamTLSSnapshot, cfg)
 	case ForwardProtocolAuto:
 		// USK-913: peek the first inner byte on clientConn to disambiguate
 		// HTTP/1.x → http arm, h2c → reject with USK-914 citation, TLS →
 		// warn + raw fallback, everything else → raw fallback.
-		return buildTargetOverrideAutoStack(ctx, clientConn, upstreamConn, params.Target, params.effectiveScheme(), params.ClientTLSSnapshot, cfg)
+		return buildTargetOverrideAutoStack(ctx, clientConn, upstreamConn, params.Target, params.effectiveScheme(), params.ClientTLSSnapshot, params.UpstreamTLSSnapshot, cfg)
 	case ForwardProtocolHTTP2, ForwardProtocolGRPC:
 		// USK-914: h2c forward stack. The GRPC selector reuses the same
 		// h2c stack assembly; the gRPC content-type filter is applied at
 		// the per-stream proxybuild dispatch layer (see
 		// internal/proxybuild/tcp_forward_h2.go).
-		return buildTargetOverrideH2CStack(ctx, clientConn, upstreamConn, params.Target, params.effectiveScheme(), params.ClientTLSSnapshot, cfg)
+		return buildTargetOverrideH2CStack(ctx, clientConn, upstreamConn, params.Target, params.effectiveScheme(), params.ClientTLSSnapshot, params.UpstreamTLSSnapshot, cfg)
 	default:
 		// Unreachable: params.validate already rejects unknown values.
 		return nil, fmt.Errorf("connector: BuildConnectionStackWithTarget: unhandled protocol %q", protocol)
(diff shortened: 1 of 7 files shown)


usk6666 (commented):
## Code Review: APPROVED

Implementation cleanly wires upstream-side TLS dial through a shared `dialForwardUpstream` helper used by both H1 and H2 forward arms. The seam discipline (caller-side dial; builder consumes snapshot via the new `UpstreamTLSSnapshot` field) mirrors USK-915 well, ALPN propagation policy is documented and tested for all 3 cases, error classification reuses `buildTLSStackBuildErrorRecorder`, and e2e coverage hits the 4 TLS×UpstreamTLS cells, ALPN cells, cert-verify failure, and plugin hook firing. Findings are all minor.

### Findings

| F-1 | LOW | internal/proxybuild/tcp_forward_tls.go | 459-466 | CodeHealth | `dialForwardUpstream` has an `entry == nil` guard at the start of the plain-dial branch but immediately dereferences `entry.target` on line 461 — a true nil-entry call would panic. Both production call sites pass a non-nil entry, so this is defensive-only. Either drop the `entry == nil` clause from the conjunction (leaving `entry.fc == nil || !entry.fc.UpstreamTLS`) or also guard the `entry.target` dereference. |
| F-2 | LOW | internal/proxybuild/tcp_forward_tls.go | 469 | Correctness | Upstream ALPN derivation reads `entry.fc.Protocol` (operator declaration). In the edge case `fc.Protocol="http2"` + `fc.TLS=true` + `fc.UpstreamTLS=true` + client negotiates ALPN="" (or http/1.1 if you tweak the server-side advertise), `chooseH1ForwardProtocol("http2")` already maps the client-facing dispatch down to `"http"`, but `upstreamALPNOffersForForward("http2", override)` still falls into Branch 1 and offers `["h2"]` upstream — the proxy then builds an http1 Layer on a conn that may negotiate h2. By the documented Decision #5 ("explicit declaration wins"), this is intentional; however, the asymmetry with the client-side fallback is worth noting in the helper godoc so a future reader does not assume Branch 1 is bug-free for the h2-declared-but-h1-negotiated client case. |
| F-3 | LOW | internal/proxybuild/tcp_forward_upstream_tls_integration_test.go | 30, 766-768 | CodeHealth | The `internal/flow` import is only kept alive by the bottom-of-file `var _ = (*flow.Stream)(nil)` "compile guard". The file does not actually use any `flow.*` types — `summarizeStreams` (which accepts a `[]*flow.Stream`) lives in `tcp_forward_tls_integration_test.go` and the test calls it with the result of `fix.store.Streams()` without naming the slice type. Drop the import and the dummy guard line. |
| F-4 | NIT | internal/proxybuild/tcp_forward_upstream_tls_integration_test.go | 697, 722-724 | CodeHealth | `startALPNObservingTLSUpstream` creates `stopCh := make(chan struct{})` and closes it inside the returned `stop` closure, but the accept goroutine never reads from it — termination is exclusively driven by `ln.Close()` returning an error from `Accept()`. The channel is dead state; either wire it into the accept loop's `select` or remove it entirely. |
| F-5 | NIT | internal/proxybuild/tcp_forward_upstream_tls_integration_test.go | 625 | GoConventions | `cap, logger := testutil.NewCaptureLogger()` shadows the `cap` builtin. Rename to e.g. `capLog` / `capture` to silence the shadow and read more clearly at the call site (`cap.Contains(...)` → `capture.Contains(...)`). |

### Suggestions

- F-1: simplest fix is to drop `entry == nil ||` from the early-exit conjunction since both call sites already pass non-nil entries; the nil-check is misleading because the very next line dereferences `entry.target`.
- F-2: add one sentence to `upstreamALPNOffersForForward`'s Branch 1 godoc clarifying that explicit-protocol declarations bypass any client-side fallback derived by `chooseH1ForwardProtocol`, so operators are responsible for the explicit-declaration/client-capability match (or defer to `Protocol="auto"`).
- F-3: remove the `flow` import on line 30 and the trailing guard on lines 766-768.
- F-4: either delete `stopCh` (3 lines) or add `case <-stopCh: return` to the accept loop's `select` for explicit termination signaling.
- F-5: `cap, logger := testutil.NewCaptureLogger()` → `capture, logger := testutil.NewCaptureLogger()`; update the two subsequent `cap.` references.

**STATS**: CRITICAL: 0  HIGH: 0  MEDIUM: 0  LOW: 3  NIT: 2

---

usk6666 (commented):
## Security Review: APPROVED

USK-916 wires upstream-side TLS dial for the L7 TCP forward path with sound MITM hygiene: SNI is derived from the operator-declared Target (not the client SNI), ALPN propagation is restricted to crypto/tls-validated strings, plugin hook payload exposes only summary fields, and the cert-verify-failure path closes the TCP socket and records `state="error"` with `FailureReason="upstream_tls_error"`. The implementation mirrors the USK-915 client-side terminate seam (TLS happens in the per-conn handler; the builder consumes the snapshot) and reuses the existing `DialUpstreamRaw` / `buildTLSStackBuildErrorRecorder` plumbing — no new attack surface.

### Verification highlights
- **SNI isolation (MITM principle)** — `tlsCfg.ServerName = targetHostOnly(entry.target)` uses operator-declared Target only; client-supplied SNI is not propagated. Combined with `connector.validateTarget` (rejects CRLF, enforces host:port) and TLS extension length-framing, no SNI injection vector.
- **ALPN propagation** — In the `auto + client TLS + client ALPN` branch, `clientTLSSnapshot.ALPN` is `crypto/tls.ConnectionState.NegotiatedProtocol` — constrained to values the proxy itself advertised via `alpnOffersForForwardProtocol` (hard-coded `{"h2","http/1.1"}` set). Not attacker-injectable.
- **TLS MinVersion** — `tls.VersionTLS12` floor matches USK-915 client-side terminate symmetry; MITM-exception note acknowledges legacy-TLS allowance is operator-configurable elsewhere.
- **InsecureSkipVerify source** — Only `BuildConfig.InsecureSkipVerify` (global, operator-controlled) is honoured. Per-host TLS material / mTLS / uTLS deferral is documented in the helper godoc.
- **Plugin hook payload** — `BuildTLSHandshakeDict` returns frozen Starlark dict of `{sni, alpn, version_name, cipher_name, peer_cert_subject (string DN only), client_fingerprint}`. No raw cert chain bytes; nil-safe via `PeerCertSubject()`.
- **Error path cleanup** — `DialUpstreamRaw` closes the underlying TCP conn on TLS handshake error (`dial_raw.go:89`); no caller-side leak. Failure is classified as `upstream_tls_error` by `buildTLSStackBuildErrorRecorder` (errors.Is check on `ErrClientTLSMITMHandshake` correctly routes upstream-direction failures to the right Stream FailureReason).
- **Timeouts** — `tcpForwardDialTimeout = 30s` covers TCP+TLS via `context.WithTimeout` inside `DialUpstreamRaw`; `tlslayer.Client` uses `HandshakeContext(ctx)` so cancellation propagates.
- **Concurrency** — `dialForwardUpstream` is stateless; reads only from immutable `BuildConfig` and `PluginV2Engine` after Manager init. No shared mutable state between per-conn handler goroutines.
- **Plugin fail-soft** — `fireForwardTLSHandshakeHook` logs and swallows hook errors (USK-671 policy); a noisy or buggy Starlark plugin cannot DoS the forward data path.
- **No new dependencies.**

### Low findings

  - ID: S-1
    Severity: LOW
    File: internal/proxybuild/tcp_forward_tls.go
    Line: 459-465
    Category: GoSecurity
    CWE: CWE-476 (NULL Pointer Dereference)
    Description: `dialForwardUpstream` guards with `if entry == nil || entry.fc == nil || !entry.fc.UpstreamTLS`, then dereferences `entry.target` on line 461 inside that branch. If `entry == nil` (first disjunct) were ever true, the plain-dial branch would nil-panic instead of failing cleanly. In production `entry` is always non-nil (sourced from validated forward listener entries via the accept loop), so this is not currently triggerable — but the defensive check is misleading.
    Impact: No exploitability under current call graph; cosmetic robustness concern only. Drop the `entry == nil` disjunct OR add an early-return `if entry == nil { return nil, nil, fmt.Errorf(...) }` before line 459 to make the contract explicit.
    Remediation: Either remove the `entry == nil` clause from the guard (current callers guarantee non-nil) or split the guard into two checks: `if entry == nil { return ... }` followed by `if entry.fc == nil || !entry.fc.UpstreamTLS { plain dial }`. Non-blocking.

STATS: CRITICAL: 0  HIGH: 0  MEDIUM: 0  LOW: 1

---

PR #12, opened 2026-05-17
feat(mcp,grpc): .proto schema-aware decode/encode + grpc_schema MCP tool (USK-923)

usk6666:
## Summary

- New package `internal/encoding/protoschema/` — `Decode(raw, msgDesc)` / `Encode(jsonStr, msgDesc)` via `protoreflect.DynamicMessage` + `protojson`, plus a lock-free `Registry` (atomic snapshot swap on register/unregister/clear).
- New MCP tool `grpc_schema` (actions: `register` / `list` / `unregister` / `clear`). Persistent via new `SchemaStore` interface + `grpc_schemas` SQLite table (sibling of `MacroStore`).
- `register` accepts `source: "descriptor_set"` only — `descriptor_set_b64` (≤16 MiB after base64 decode). `source: "file"` is explicitly rejected with a pointer to `protoc --include_imports --descriptor_set_out=...` (deferred to follow-up Issue).
- Query side (`computeDecodedGRPCBodyWithLimit`): schema lookup by `(grpc_service, grpc_method)` from `Flow.Metadata`. Direction (`send`/`receive`) picks input vs output descriptor. On hit → `body_decoded_encoding="proto-json"`, real field names; on miss → existing schemaless fallback; on schema-hit-but-parse-fail → schemaless fallback + `body_decode_anomaly{Type:"proto_schema_mismatch"}`.
- Resend side: new `body_encoding="proto-json"` next to `"proto-schemaless-json"`. Hard error if no schema for `service`/`method`. Non-fatal warning if the source flow had unknown bytes (lossy round-trip).
- Output Filter (PII masking) applies to protojson output identically. P5 guards: size cap pre-decode, `defer recover()` around protoreflect calls, `protojson.UnmarshalOptions{DiscardUnknown:true}` on encode.
- Help docs: new `help_grpc_schema.md` + `schema_grpc_schema.json`; updates to `help_query.md` + `help_resend_grpc.md`. CLI `client.go` / `client_params.go` updated to surface `grpc_schema` in the tool list.

## Fitness check (6 MITM principles)

- P1, P2, P4, P6 — PASS. Wire path untouched; gRPC-only gated on `grpc_event=="data"`.
- P3 — **PARTIAL**: `protojson.Marshal` drops unknown fields. Documented prominently in `help_grpc_schema.md`; non-fatal `warnings[]` emitted on resend when the source flow's bytes carry unknowns. Schemaless path remains lossless.
- P5 — PASS WITH GUARDS (16 MiB size cap before `protodesc.NewFiles`, `defer recover()` around every protoreflect entry point, `DiscardUnknown=true` on Encode).

## Deferred (will be filed as separate Issues by the orchestrator)

- `source="file"` + host `protoc` invocation
- WebUI schema management
- gRPC server-reflection-based auto-schema discovery
- Unknown-field preserving JSON round-trip (defer-pending-demand)

## Test plan

- [x] `make lint` clean (gofmt + go vet + staticcheck + ineffassign + gocyclo ≤15)
- [x] `make build` succeeds
- [x] `make test` (fast tier) passes
- [x] e2e exhaustive (`-tags 'e2e'`): register → MITM gRPC flow → query decoded → list/unregister/clear lifecycle
- [x] `-race` race-detector clean on Registry concurrent read/write (atomic snapshot test)

Resolves USK-923
Linear: https://linear.app/usk6666/issue/USK-923


usk6666 (commented):
## Code Review: CHANGES REQUESTED

The PR is well-structured: clear package boundaries (`internal/encoding/protoschema/` is a clean sibling of `internal/encoding/protobuf/`), excellent docstrings, atomic-snapshot Registry with a real concurrent-read-during-write test, and full lifecycle coverage of register/list/unregister/clear through the MCP transport. Lossy round-trip is documented and warned non-fatally. However, there is one HIGH correctness bug in the rehydrate-after-restart path that loses every sibling service in a multi-service descriptor set, plus a few minor issues.

### Findings

| F-1 | HIGH | internal/mcp/grpc_schema_tool.go | 318-347 | Correctness | `ensureGRPCSchemaRehydrated` loses sibling services from the same descriptor set after process restart. The `parsedByBytes` cache stores the result of `LoadFileDescriptorSet(rec.DescriptorSet, []string{rec.Service})` (filter=[first service only]). When the loop hits another DB row sharing the same `descriptor_set` BLOB (which is the normal case — `handleGRPCSchemaRegister` writes the SAME `raw` for every service inside the descriptor), the cached spec list contains only the FIRST service. `applyRehydrateLabel(cached, rec)` tries to update an entry whose `Service` does not match the new record, silently no-ops, and the second service is never loaded. Result: post-restart `list` returns only one of N services that were registered together. |
| F-2 | LOW | internal/mcp/server.go | 437-442 | Correctness | `grpcSchemaRegistry()` has a check-then-act data race. Two concurrent MCP tool handlers (e.g., `grpc_schema register` + `query`) running before the Registry was lazy-initialized can both observe `s.grpcSchemas == nil` and both create + assign a fresh Registry, dropping the loser's entries. The Registry itself is concurrency-safe (atomic snapshot), but the **publication** of the pointer is not. |
| F-3 | LOW | internal/encoding/protoschema/registry.go | 385-393 | Correctness | `isMissingDependencyErr` matches the substring `"not found"`, which is overly broad. Any future `protodesc` / `protoregistry` error containing those two words (e.g., a future "enum value X not found in type Y") will be re-wrapped with the misleading "run protoc --include_imports" hint. Tighten to `"could not resolve import"` / `"missing dependency"` only, or match on `protoregistry.NotFound` via `errors.Is`. |
| F-4 | NIT | internal/mcp/testdata/test.desc, internal/mcp/testdata/with_imports_missing.desc | — | CodeHealth | Test descriptors are duplicated across `internal/encoding/protoschema/testdata/` and `internal/mcp/testdata/`. The duplication is explained in `grpc_schema_tool_test.go` (go:embed cannot escape the package directory). Acceptable for now, but the `.desc` files can drift if `test.proto` is regenerated only in the protoschema package. Consider a tiny helper package that exports `var TestDescBytes []byte` via `//go:embed`. (NIT — keep as documented duplication is fine.) |

### Suggestions

**F-1 — Rehydrate cache fix**

The cleanest fix is to load each DB record independently (each `rec` carries its own service filter), and drop the `parsedByBytes` byte-cache entirely. The optimization was assuming "one payload → one service", which is wrong for the common register-all-services case.

```go
// Replace the parsedByBytes loop in ensureGRPCSchemaRehydrated with:
all := make([]*protoschema.ServiceSpec, 0, len(records))
var groupErr error
for _, rec := range records {
    specs, perr := protoschema.LoadFileDescriptorSet(rec.DescriptorSet, []string{rec.Service})
    if perr != nil {
        if groupErr == nil {
            groupErr = fmt.Errorf("rehydrate schema for service %q: %w", rec.Service, perr)
        }
        continue
    }
    applyRehydrateLabel(specs, rec)
    all = append(all, specs...)
}
if len(all) > 0 {
    s.grpcSchemaRegistry().Register(all)
}
return groupErr
```

If the per-record `protodesc.NewFiles` cost matters in practice (it shouldn't — a handful of services per operator), replace with a single full-set load + per-record service projection instead of the buggy byte-key cache. Either way, please add a unit test that calls `ensureGRPCSchemaRehydrated` against a SchemaStore populated with two rows from the same descriptor set and asserts both services appear in `ListServices()`.

**F-2 — Publication fix**

Use `sync.Once` (simpler) or change `grpcSchemas` to `atomic.Pointer[protoschema.Registry]`:

```go
// Option A: sync.Once
type Server struct {
    ...
    grpcSchemas     *protoschema.Registry
    grpcSchemasOnce sync.Once
}

func (s *Server) grpcSchemaRegistry() *protoschema.Registry {
    s.grpcSchemasOnce.Do(func() {
        s.grpcSchemas = protoschema.NewRegistry()
    })
    return s.grpcSchemas
}
```

Also update `lookupGRPCSchema` to read through the same accessor so the publication is consistent.

**F-3 — Narrow substring match**

```go
func isMissingDependencyErr(err error) bool {
    if err == nil {
        return false
    }
    s := err.Error()
    return strings.Contains(s, "could not resolve import") ||
        strings.Contains(s, "missing dependency")
}
```

The `"not found"` branch is the leaky one; the other two strings are specific enough to protodesc resolution failures.

---

usk6666 (commented):
## Security Review: APPROVED

The schema-aware decode/encode path is bounded by every guard documented in the design review:
- 16 MiB cap on the **post-base64-decode** descriptor_set size, enforced both in `handleGRPCSchemaRegister` (`internal/mcp/grpc_schema_tool.go:183-185`) and in `LoadFileDescriptorSet` (`internal/encoding/protoschema/registry.go:248-250`).
- `defer recover()` around every protoreflect entry point: `decodeSafe`, `encodeSafe`, `hasUnknownFieldsSafe`, `loadSafe` (`internal/encoding/protoschema/protoschema.go:62-67, 109-114, 144-150` and `registry.go:268-273`).
- `protojson.UnmarshalOptions{DiscardUnknown: true}` on the resend encode path (`internal/encoding/protoschema/protoschema.go:115-118`).
- SQL persistence uses parameterized queries throughout (`internal/flow/sqlite.go:980-983, 996, 1022, 1052, 1072`).
- Output Filter (PII masking) applies to the proto-json output before it leaves the MCP boundary (`internal/mcp/query_tool.go:1129-1141` `finalizeDecodedGRPC` → `s.filterOutputBody(plain)`).
- CRLF guards on user-supplied service/method/scheme/target_addr in `validateResendGRPCStringFields` (existing USK-922 hygiene; unchanged).
- Service/method names from descriptor sets feed only into in-memory map keys and parameterized SQL; no header injection / log injection surface.

The `protojson.Marshal` "drops unknown fields" lossy edge is a documented, accepted trade-off and the resend path emits a non-fatal warning when the source flow had unknown bytes — explicitly out of scope per the locked-in design review.

### Findings

| S-1 | LOW | internal/mcp/grpc_schema_tool.go | 179-185 | InputValidation | CWE-770 | `base64.StdEncoding.DecodeString(params.DescriptorSetB64)` is called before any size check on the input. A maliciously large base64 string (e.g., 1 GiB) allocates the decoded byte slice in full before the 16 MiB cap fires. Loopback-only MCP transport and the requirement for local MCP access make this LOW: the attack surface is "an MCP-authorized process can exhaust this proxy's memory", which is already true via many other tools. Recommend an upper bound on `len(params.DescriptorSetB64)` itself (e.g., `4 * MaxDescriptorSetBytes / 3` ≈ 22 MiB) before calling `DecodeString` for fail-fast cleanliness, but not blocking on this for v1. |

### Remediation

```go
// Add to handleGRPCSchemaRegister before DecodeString:
if len(params.DescriptorSetB64) > protoschema.MaxDescriptorSetBytes*4/3+16 {
    return nil, fmt.Errorf("descriptor_set_b64 length %d exceeds maximum encoded size", len(params.DescriptorSetB64))
}
```

Optional; the design is already within the project's MCP/loopback threat model.

---

usk6666 (commented):
## Code Review (Round 2): APPROVED

All four findings from round 1 are resolved cleanly in `a74472be`:

- **F-1 (HIGH)**: The byte-keyed `parsedByBytes` cache in `ensureGRPCSchemaRehydrated` was removed; each record now loads via its own `LoadFileDescriptorSet(rec.DescriptorSet, []string{rec.Service})`. The new regression test `TestGRPCSchema_Rehydrate_MultiServiceFromSameDescriptorSet` seeds two rows sharing one descriptor_set payload and asserts both services rehydrate.
- **F-2 (LOW)**: `Server.grpcSchemas` is now initialised under `sync.Once`. `lookupGRPCSchema` reads through `grpcSchemaRegistry()` so the publication race window is closed.
- **F-3 (LOW)**: `isMissingDependencyErr` dropped the overly broad `"not found"` substring; only the protodesc-specific markers remain.
- **S-1 (LOW)**: Pre-decode size cap on `descriptor_set_b64` (≈22 MiB encoded ceiling) closes the unbounded `base64.StdEncoding.DecodeString` allocation window.

Verification: `make lint` clean, `make test -race` passes including the new rehydrate regression test.

---

PR #13, opened 2026-05-18
fix(mcp,grpc): recover authority/service/method for listener-captured gRPC flows (USK-920)

usk6666:
## Summary
- Add `Authority` / `Scheme` / `Path` fields to `GRPCStartMessage`; populate from `H2HeadersEvent` in the gRPC Layer so the wire pseudo-headers reach the Pipeline.
- Project `Flow.URL` (and `Stream.Scheme`) in `record_step` for `GRPCStartMessage`, mirroring the existing HTTPMessage projection. Listener-captured gRPC flows now record a URL the same way HTTP flows do.
- Add a `Flow.Metadata` fallback in `extractResendGRPCStartFields` so already-recorded flows without `Flow.URL` still recover service/method.
- Scan for `grpc_event=="start"` in `loadResendGRPCFlows` instead of returning `sendFlows[0]` — the live data path records per-frame wire snapshots ahead of the semantic Start Flow.

## Test plan
- [x] Unit: `extractResendGRPCStartFields` covers URL-only / Metadata-only / both-present / URL-without-path / neither / nil-flow.
- [x] E2E (`e2e && !e2e_smoke`): listener-captured gRPC + `resend_grpc { flow_id }` with `body_encoding` ∈ {`base64`, `proto-schemaless-json`, `proto-json` with schema, `proto-json` without schema}.
- [x] `make lint`, `make build`, `make test`, `make test-e2e-smoke` all pass locally.
- [x] No regression on `resend_http` / `resend_ws` (unchanged code paths).

Resolves USK-920
Linear: https://linear.app/usk6666/issue/USK-920

### DIFF
--- a/docs/rfc/envelope.md
+++ b/docs/rfc/envelope.md
@@ -233,6 +233,16 @@ type GRPCStartMessage struct {
     Service string
     Method  string
 
+    // USK-920: the request-side pseudo-headers (:authority, :scheme,
+    // :path) are projected as derived L7 fields so RecordStep can build
+    // Flow.URL the same way it does for HTTPMessage. Empty on the
+    // response side (response HEADERS only carry :status). The pseudo-
+    // header wire bytes remain in Envelope.Raw — these fields are an
+    // overlay, not a replacement.
+    Authority string
+    Scheme    string
+    Path      string
+
     // gRPC metadata — custom and reserved. HTTP/2 pseudo-headers (:method,
     // :path, :status, etc.) are NOT included here; they belong to the
     // transport layer and are observable via Envelope.Context if needed.
--- a/internal/envelope/grpc.go
+++ b/internal/envelope/grpc.go
@@ -8,6 +8,14 @@ import "time"
 // One GRPCStartMessage envelope is emitted for the request side
 // (Direction=Send, Sequence=0) and one for the response side
 // (Direction=Receive, Sequence=0). Both share the RPC's Envelope.StreamID.
+//
+// USK-920: Authority / Scheme / Path are projected out of the HTTP/2
+// pseudo-headers on the request side so downstream consumers (notably
+// RecordStep → Flow.URL) can reconstruct the RPC's URL without re-parsing
+// the wire HPACK block. The pseudo-header wire bytes remain in
+// Envelope.Raw (untouched); these fields are a derived L7 overlay,
+// matching the existing HTTPMessage.Authority / Scheme / Path pattern.
+// Empty on the response side (response HEADERS only carry :status).
 type GRPCStartMessage struct {
 	// Service is the gRPC service name (derived from :path on the request
 	// side and mirrored on the response side).
@@ -15,6 +23,22 @@ type GRPCStartMessage struct {
 
 	// Method is the gRPC method name.
 	Method string
+
+	// Authority is the :authority pseudo-header from the request-side
+	// HEADERS frame. Empty on the response side. USK-920.
+	Authority string
+
+	// Scheme is the :scheme pseudo-header from the request-side HEADERS
+	// frame ("http" for h2c, "https" for TLS+ALPN h2). Empty on the
+	// response side. USK-920.
+	Scheme string
+
+	// Path is the :path pseudo-header from the request-side HEADERS
+	// frame ("/Service/Method"). Empty on the response side. USK-920.
+	// Service / Method are also parsed out individually; Path is retained
+	// verbatim so a malformed :path (which yields Service="" Method="")
+	// still leaves an inspectable trace.
+	Path string
 
 	// Metadata is the full gRPC metadata list. HTTP/2 pseudo-headers are
 	// NOT included here — they belong to the transport layer and are
@@ -53,6 +77,9 @@ func (m *GRPCStartMessage) CloneMessage() Message {
 	return &GRPCStartMessage{
 		Service:        m.Service,
 		Method:         m.Method,
+		Authority:      m.Authority,
+		Scheme:         m.Scheme,
+		Path:           m.Path,
 		Metadata:       cloneKeyValues(m.Metadata),
 		Timeout:        m.Timeout,
 		ContentType:    m.ContentType,
--- a/internal/layer/grpc/channel.go
+++ b/internal/layer/grpc/channel.go
@@ -1064,7 +1064,16 @@ func buildStartMessage(evt *http2.H2HeadersEvent, dir envelope.Direction) (*enve
 	// Service / Method on the request side derive from :path; on the
 	// response side they are mirrored from the request side by the
 	// channel.absorbHeaders caller (the response carries no :path).
+	//
+	// USK-920: also surface the request-side pseudo-headers (:authority /
+	// :scheme / :path) as derived L7 fields so RecordStep can project
+	// Flow.URL the same way it does for HTTPMessage. Response-side
+	// HEADERS frames only carry :status (no authority/scheme/path), so we
+	// leave the fields zero in that case.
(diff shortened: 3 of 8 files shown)


usk6666 (commented):
## Code Review: APPROVED

The PR cleanly closes USK-920 by mirroring the HTTPMessage projection pattern: gRPC Layer surfaces :authority/:scheme/:path on `GRPCStartMessage`, `record_step.projectGRPCStart` projects them into `Flow.URL` (and `Stream.Scheme`), `extractResendGRPCStartFields` reads URL first with a Metadata fallback for legacy flows, and `pickGRPCStartFlow` correctly skips the new wire-level frame envelopes that share `StreamID`. Build + lint pass on CI.

### Strengths

- Symmetric with the existing HTTPMessage projection (single read-site in `envelopeToFlow`, same `*url.URL` build pattern, no `Parse`/`String` round-trip — preserves the USK-672 CWE-918 lesson).
- `pickGRPCStartFlow` correctly disambiguates semantic Start envelopes from the USK-896/899 per-frame wire envelopes by scanning `Metadata["grpc_event"]=="start"` (the wire-level rows have `Message=nil` and never carry that key).
- `extractResendGRPCStartFields` is nil-safe (`if f == nil`) and gracefully degrades on every component (URL-only, Metadata-only, URL-without-path).
- USK-920 Linear addendum coverage in `resend_grpc_listener_capture_integration_test.go` is complete: all four cases — base64, proto-schemaless-json, proto-json with schema, proto-json without schema — are tested; the no-schema arm asserts the error string contains the recovered `service` and `method` per the addendum.
- Doc parity: `docs/rfc/envelope.md` is updated alongside `GRPCStartMessage` to reflect the new fields and overlay semantics.

### Low / Nit Findings (non-blocking)

| F-1 | LOW | internal/pipeline/record_step.go | 833-846, 1262-1275 | TestQuality | Neither the `createStream` Scheme branch nor `projectGRPCStart`'s URL projection has a direct unit test in `internal/pipeline/*_test.go`. The behaviour is reachable only via the `e2e && !e2e_smoke` integration test, so a CI regression in the projection (e.g. swapping `m.Authority` ↔ `m.Path`) would only surface in the nightly tier. A small table-driven test in `record_step_test.go` covering (Authority+Scheme+Path) / (Authority only) / (response-side empty) would close the gap. |
| F-2 | NIT | internal/pipeline/record_step.go | 833-846 | CodeHealth | The `switch msg := env.Message.(type)` over `*HTTPMessage` and `*GRPCStartMessage` runs the same `if msg.Scheme != "" { st.Scheme = msg.Scheme }` body in both arms. Either a tiny helper `schemeFrom(msg)` or a generic-method-set interface would dedupe, but the current shape makes the protocol-level intent more readable — fine to leave as is. |
| F-3 | NIT | internal/mcp/resend_grpc_helpers.go | 449-477 | Correctness | `pickGRPCStartFlow(recvFlows)` is called for the receive-direction Flow to read `Grpc-Accept-Encoding`. On a trailers-only D4 response (the receive direction starts AND ends in a single envelope), `recvFlows` may contain only the End / Data variant rather than a true Start. The function's fallback `return flows[0]` then returns the End Flow, whose `Headers` map may not carry `grpc-accept-encoding` — yielding the same empty result as today's prod code. This is not a regression (prior code was `recvFlows[0]` directly) and only affects the optional `AcceptEncoding` recovery, but a follow-up could tighten by either returning `nil` from the fallback when no `grpc_event=start` flow exists on the recv side, or by extending the projection to also match `grpc_event=end` headers (response trailers carry no AcceptEncoding by gRPC convention, so the nil-return option is cleaner). Out of scope for USK-920. |

### Suggestions

- Add a `TestProjectGRPCStart_URLProjection` / `TestRecordStepCreateStream_GRPCScheme` table-driven test in `internal/pipeline/record_step_test.go` (or a sibling file) covering F-1.

STATS:
  CRITICAL: 0
  HIGH: 0
  MEDIUM: 0
  LOW: 1
  NIT: 2

---

usk6666 (commented):
## Security Review: APPROVED

No new attack surface, no new external dependencies, no panic-risk in the new code paths. The fix recovers RPC identity from already-recorded wire data; values that originated from attacker-controlled traffic remain attacker-controlled but are not granted new pathways into the proxy's CA/control plane.

### Threat-Model Walkthrough

1. **CRLF / header injection (CWE-93)** — `validateResendGRPCStringFields` continues to gate every user-supplied URL component (`service`, `method`, `scheme`, `target_addr`) with `validateResendGRPCNoCRLF` before any dial. The new recovery path reads `service` / `method` / `authority` from `Flow.URL` and `Flow.Metadata`, which are recorded values from the prior wire observation; per the MITM Implementation Principle in CLAUDE.md, recorded data is not sanitized (preserves the pentester's intentional anomalies). The recovered values then flow into HPACK encoding via the gRPC Layer, which is the same path used by the live capture; there is no new bypass of input validation. Acceptable.

2. **SSRF (CWE-918)** — `extractResendGRPCStartFields` reads `f.URL.Host` directly into `authority`, never round-tripping through `url.Parse(str)`. `resendGRPCCanonicalURL` builds the canonical URL by assigning struct fields (no `Parse(String())`). This preserves the USK-672 lesson that motivated the existing canonical-URL build. The `checkResendGRPCScope` gate continues to validate both the canonical URL and the override `target_addr` (when distinct), so a recovered authority that points outside the configured scope is rejected before any dial.

3. **CWE-20 / panic on malformed input** — `extractResendGRPCStartFields` has an explicit `nil` guard on the input flow, and `strings.SplitN(strings.TrimPrefix(f.URL.Path, "/"), "/", 2)` is panic-safe on every input (empty / leading-slash-only / no-slash). The new `record_step.projectGRPCStart` URL build is unconditional on `m.Authority != "" || m.Scheme != ""`; the inner `path` fallback handles the edge case where Service+Method are set but Path is somehow empty without panicking.

4. **TLS / Certificates** — No changes to the proxy's TLS listener, CA key handling, or upstream dial TLS configuration. `dialResendGRPCUpstream` is unchanged. The recovered `scheme` ("http" / "https") only affects whether the resend dial does an explicit TLS upgrade; it cannot downgrade an existing listener to plaintext or affect the cert-issuance path.

5. **GoSecurity (race / leak)** — Two new field-write sites are inside existing locked sections (`channel.go`'s `buildStartMessage` runs under `c.mu` via `absorbHeaders`) or pure mutator helpers (`record_step.projectGRPCStart`), neither introduces goroutine state. No new `unsafe` / `math/rand` / `os.Exec`.

6. **MCP / API surface** — The `resend_grpc` tool's schema is unchanged; only the recovery logic behind `flow_id` is extended. Error messages for the no-schema arm now correctly include the recovered `service` / `method` (verified by the new integration test), which is the desired behaviour — operator gets actionable feedback. The exposed strings come from the recorded HEADERS frame, which is no more sensitive than what `query messages { id: <flow_id> }` already returns.

7. **Dependencies / licenses** — No additions to `go.mod`.

### MITM Wire-Fidelity Compliance

The PR adds `Authority` / `Scheme` / `Path` as *derived overlay fields* on `GRPCStartMessage` (documented as such in both godoc and RFC). `Envelope.Raw` continues to hold the HPACK wire bytes verbatim; the new fields project them into a higher-level shape consumed by `RecordStep` and `resend_grpc`. The original wire-observed casing/order of the metadata multimap is preserved (the existing `Metadata []KeyValue` slot is untouched). MITM Implementation Principles 1-4 hold.

### Findings

None.

STATS:
  CRITICAL: 0
  HIGH: 0
  MEDIUM: 0
  LOW: 0

---

PR #16, opened 2026-05-18
feat(tcp_forwards,config,mcp): upstream_insecure_skip_verify per-entry override (USK-918)

usk6666:
## Summary

- Adds `upstream_insecure_skip_verify *bool` to `ForwardConfig` with tri-state semantics: `nil` = inherit global `insecure_skip_verify`, `true` = skip verify (insecure), `false` = enforce verify.
- Threads the per-entry value through `dialForwardUpstream` (single dial site covers both H1 and H2 forward arms). Resolution extracted as a pure helper `resolveUpstreamInsecureSkipVerify` for unit testability.
- Mirrors the sibling `HostTLSEntry.TLSVerify *bool` pattern exactly.
- MCP wiring: `parseTCPForwardsAny` explicit field handling, `validateTCPForwardsConfig` warns on `skip=true && upstream_tls=false` and info-logs when the resolved value is `true` (security event).
- JSON Schema + `help_proxy_start.md` updated.

Use case: MITM-debugging a local h2-TLS gRPC server (e.g. `moul/grpcbin`) with a self-signed cert that lacks IP SANs. Scoped per-entry so the HTTP CONNECT MITM path (`tls_passthrough`) is unaffected.

## Test plan

- [x] Unit: `internal/config/forward_config_test.go` matrix extended (Unmarshal tri-state, Marshal omitempty + explicit true/false survival, RoundTrip, Validate)
- [x] Unit: `internal/proxybuild/tcp_forward_tls_test.go` `TestResolveUpstreamInsecureSkipVerify` covers 8 cross-product cells (per-entry × global)
- [x] Integration: `tcp_forward_upstream_tls_integration_test.go` `TestTCPForward_UpstreamTLS_PerEntryInsecureSkipVerify` global × per-entry matrix (nil/true/false against self-signed upstream)
- [x] `make lint` clean
- [x] `make build` succeeds
- [x] `make test` passes
- [x] `make test-e2e-smoke` passes

Resolves USK-918
Linear: https://linear.app/usk6666/issue/USK-918

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -400,6 +400,12 @@ func isValidProjectNameRune(r rune) bool {
 // Both fields default to false, which preserves the existing raw-forward
 // behavior. Target's scheme/port are advisory only; they do not change the
 // effective TLS / UpstreamTLS values.
+//
+// Upstream TLS certificate verification defaults to the global
+// Config.InsecureSkipVerify setting; a per-entry
+// UpstreamInsecureSkipVerify (*bool, nil/true/false) overrides that global
+// when set (USK-918). The override is scoped to the tcp_forwards entry and
+// is independent of the CONNECT-MITM tls_passthrough path.
 type ForwardConfig struct {
 	// Target is the upstream address (e.g. "api.example.com:50051").
 	Target string `json:"target"`
@@ -427,6 +433,19 @@ type ForwardConfig struct {
 	// independent of TLS and is not inferred from Target's scheme/port.
 	// See the type-level godoc for the full TLS × UpstreamTLS matrix.
 	UpstreamTLS bool `json:"upstream_tls,omitempty"`
+
+	// UpstreamInsecureSkipVerify is a per-entry override for upstream TLS
+	// certificate verification when UpstreamTLS=true. Tri-state semantics:
+	//   - nil   = inherit the global Config.InsecureSkipVerify (default)
+	//   - true  = skip verification on this entry (insecure; equivalent of
+	//             grpcurl -insecure / curl -k, but scoped per tcp_forwards
+	//             entry)
+	//   - false = enforce verification on this entry even when the global
+	//             flag is true
+	//
+	// Used for MITM-debugging local upstreams whose self-signed certs lack
+	// SAN entries for the dialled host/IP (e.g. moul/grpcbin). USK-918.
+	UpstreamInsecureSkipVerify *bool `json:"upstream_insecure_skip_verify,omitempty"`
 }
 
 // validForwardProtocols is the set of valid protocol values for ForwardConfig.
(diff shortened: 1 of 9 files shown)


usk6666 (commented):
## Code Review: APPROVED

The PR cleanly threads a tri-state `*bool` UpstreamInsecureSkipVerify field from `ForwardConfig` through `parseTCPForwardsAny` and a new pure helper `resolveUpstreamInsecureSkipVerify` to the single dial site `dialForwardUpstream`. The 4-cell global × per-entry integration matrix is exercised against a real self-signed httptest upstream in the exhaustive e2e tier. `make lint` clean, `make test` green locally.

### Findings (all LOW — auto-fixed in follow-up commit)

| ID | Severity | File | Category | Description | Status |
|----|----------|------|----------|-------------|--------|
| F-1 | LOW | internal/mcp/proxy_start_tool_test.go | TestQuality | `TestParseTCPForwardsAny_UpstreamTLS` did not extend the matrix to `upstream_insecure_skip_verify`, leaving the live MCP `map[string]any` parsing branch uncovered. (`TestForwardConfig_UnmarshalJSON_UpstreamInsecureSkipVerify` only covers the default JSON round-trip branch.) | FIXED |
| F-2 | NIT | internal/proxybuild/tcp_forward_tls.go | CodeHealth | Godoc claims the resolution "mirrors the sibling `HostTLSEntry.TLSVerify` pattern", but the semantic axis is inverted (TLSVerify is true=enforce; UpstreamInsecureSkipVerify is true=skip). The shape mirrors; the axis does not. | FIXED |

## Security Review: APPROVED

The bypass capability itself is the intended feature of USK-918 (MITM-debugging local upstreams with SAN-deficient self-signed certs); per the pentesting threat model and the orchestrate-level MITM rejection rules, the bypass is not flaggable. Around it:

- **Scope is correctly contained**: per-entry scoped to a `tcp_forwards` entry; the CONNECT-MITM `tls_passthrough` axis is unaffected.
- **`*bool` tri-state semantics** correctly distinguish "inherit global" from "explicitly enforce" — the existing `omitempty` test asserts the explicit `false` survives Marshal, which is the load-bearing property.
- **Security event logging** at Info per CLAUDE.md log-level guidelines (operator-visible by default): "TCP forward: upstream TLS certificate verification BYPASSED for this entry (insecure)".
- **Misconfig warn** when `skip=true && upstream_tls=false` (no-op combination): correctly emits Warn without rejecting.
- **No new attack surface**: pure config wiring; no parser, no untrusted-input handling.

### Findings: None

---
Review-gate completed — Code Review APPROVED, Security Review APPROVED, 1 fix round (LOW-only). All 8 changed files re-verified post-fix; `make lint` + `make test` clean.

Automated review by yorishiro-proxy review-gate.

PR #19, opened 2026-05-19
feat(grpc,config,mcp): grpc_schema source="file" host protoc invocation (USK-926)

usk6666:
## Summary

Adds the second source mode (`source="file"`) for the `grpc_schema` MCP tool's `register` action: the proxy invokes a host `protoc` binary on a list of absolute `.proto` paths and feeds the resulting FileDescriptorSet into the same persist + `Registry.Register` pipeline as the existing `source="descriptor_set"` path.

- USK-923's `source="descriptor_set"` stays the **recommended default** in docs and tool description (per user preference). This Issue exists for environments where invoking `protoc` server-side is acceptable.
- New `ProxyConfig.GRPCSchema.ProtocBinary` config knob with `YP_PROTOC_BINARY` env-var override; precedence: env → config → `"protoc"`.
- Security posture: absolute paths required, NUL-byte / non-canonical / `..` rejection, `EvalSymlinks` + allowed-roots check (proxy CWD or supplied `import_paths`), env minimised to `PATH` only, 30s timeout, 4 KiB stderr truncation in errors.
- New MCP param `import_paths` (optional `-I` roots); defaults to `filepath.Dir` of each `proto_paths` entry.
- Explicit, no auto-routing source × params validation per Resolved U3.
- Nightly CI installs `protobuf-compiler` so the host-protoc e2e test actually runs there.

## Resolves

Resolves [USK-926](https://linear.app/usk6666/issue/USK-926).

## Test plan

- [x] `make build` (UI + Go)
- [x] `make lint` (gofmt + go vet + staticcheck + ineffassign + gocyclo)
- [x] `make test` — fast tier passes (all existing tests + new unit tests).
- [x] `go test -race -tags e2e -run TestGRPCSchema_Register_FileSource ./internal/mcp/` — happy path, default import_paths, service filter, relative-path rejection, outside-allowed-root rejection all pass locally with `/usr/bin/protoc` (libprotoc 3.21.12).
- [x] No new go module dependencies.
- [x] Existing `source="descriptor_set"` tests continue to pass (back-compat verified).
- [ ] Code Review / Security Review focus areas: path validation order in `protoc_runner.go` (basics → EvalSymlinks → allowed-roots), `cmd.Env=[]string{"PATH="+os.Getenv("PATH")}` minimisation, nightly-e2e workflow change.

### DIFF
--- a/.github/workflows/nightly-e2e.yml
+++ b/.github/workflows/nightly-e2e.yml
@@ -37,6 +37,14 @@ jobs:
         with:
           go-version: "1.25.x"
 
+      - name: Install protoc
+        # protoc is required by the USK-926 grpc_schema source="file"
+        # e2e test (internal/mcp/grpc_schema_file_source_integration_test.go).
+        # Tests that need protoc gate on exec.LookPath and t.Skip when
+        # absent, so this step is a soft requirement: without it, the
+        # file-source happy-path tests skip silently.
+        run: sudo apt-get update && sudo apt-get install -y protobuf-compiler
+
       - name: Build UI
         run: make build-ui
 
--- a/cmd/yorishiro-proxy/client.go
+++ b/cmd/yorishiro-proxy/client.go
@@ -411,11 +411,17 @@ Parameters (key=value):
     clear                           Remove every registered schema
 
   Register parameters (key=value with params. prefix):
-    params.source=descriptor_set    Input shape (only "descriptor_set" is supported)
+    params.source=descriptor_set    Descriptor-set source (recommended). Pair with descriptor_set_b64.
+    params.source=file              File source. Pair with proto_paths; invokes a host protoc binary.
     params.descriptor_set_b64=<b64> Base64-encoded FileDescriptorSet payload (max 16 MiB decoded).
                                     Generate via:
                                       protoc --include_imports --descriptor_set_out=<file> <protos>
                                     then base64-encode the result.
+    params.proto_paths=<abs.proto>  Absolute path to a .proto file (repeatable). Required for source=file.
+                                    Each path must be canonical and fall under the proxy CWD or one of
+                                    import_paths.
+    params.import_paths=<abs-dir>   Absolute -I root for protoc (repeatable). Defaults to the parent
+                                    directory of each proto_paths entry.
     params.service_filter=<svc>     Restrict registration to a fully-qualified service name
                                     (repeatable). Empty means register every service.
     params.source_label=<label>     Free-form label preserved in list output.
@@ -426,6 +432,7 @@ Parameters (key=value):
 Examples:
   yorishiro-proxy client grpc_schema action=list
   yorishiro-proxy client grpc_schema action=register params.source=descriptor_set params.descriptor_set_b64=<base64>
+  yorishiro-proxy client grpc_schema action=register params.source=file params.proto_paths=/srv/protos/hello.proto
   yorishiro-proxy client grpc_schema action=unregister params.service=pkg.Greeter
   yorishiro-proxy client grpc_schema action=clear
 
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -692,6 +692,12 @@ type ProxyConfig struct {
 	// SSE configures the SSE Layer runtime limits (max event size).
 	SSE *SSELimits `json:"sse,omitempty"`
 
+	// GRPCSchema configures the grpc_schema MCP tool's runtime knobs
+	// (USK-926). Currently only the protoc binary path is exposed; see
+	// GRPCSchemaConfig for details. Nil/omitted = defaults (protoc on
+	// PATH).
+	GRPCSchema *GRPCSchemaConfig `json:"grpc_schema,omitempty"`
+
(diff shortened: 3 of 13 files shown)


usk6666 (commented):
## Code Review: APPROVED

The USK-926 implementation is tight: the source × params validation matrix is explicit (10 unit-test cases covering empty / descriptor_set / file × proto_paths × descriptor_set_b64 × import_paths combinations), the runner package keeps `os/exec` out of `internal/mcp/`, and the `(env → config → default)` protoc-binary resolution lives in `internal/config/grpc_schema.go` with table-driven tests for every precedence transition. Defensive double-validation in the runner (`validateRunOptions` mirrors the MCP-side checks) means `RunProtoc` is safe to call directly from tests and from any future caller. Test coverage spans happy-path, default `import_paths` derivation, service filter, relative-path rejection, non-canonical rejection, NUL-byte rejection, outside-allowed-root rejection, symlink-escape rejection (with platform skip), binary-missing (`errors.Is(err, ErrProtocNotFound())`), and protoc syntax error. The integration tests correctly route through the MCP client transport rather than calling handlers directly, matching the harness pattern.

### LOW Findings

| F-1 | LOW | internal/encoding/protoschema/protoc_runner.go | 313-321 | Correctness | `resolveAllowedRoots` silently skips roots whose `EvalSymlinks` fails. If every supplied root happens to fail (e.g. a config drift where the configured `import_paths` were since unmounted) the returned slice is empty and every input is rejected with "not under any allowed root []" — fail-closed is correct, but the error message that surfaces is the *original* `allowedRoots` slice (the unresolved input), which makes the failure mode confusing to debug. Suggestion: log the skipped roots at `slog.Debug` from inside `resolveAllowedRoots`, or include the count of resolved-vs-input roots in the rejection error. Not a blocker. |
| F-2 | NIT | internal/encoding/protoschema/protoc_runner_test.go | 117-140 | TestQuality | `TestPathUnderAnyRoot` is table-driven but loops without `t.Run` subtests, so a failing case reports only the inner-loop iteration index in the test log rather than a named subtest. The other table-driven tests in the same file use the `tc.name + t.Run` pattern — adding it here for consistency would tighten failure reporting. Not blocking. |

Neither finding affects correctness or security. Approving.

---

usk6666 (commented):
## Security Review: APPROVED

USK-926 introduces external-binary exec (`protoc`) into a user-callable MCP code path for the first time. The threat-surface mitigations match what the design review prescribed and survive concrete attack-scenario inspection:

- **Argv injection (CWE-88)**: `validatePathList` enforces `filepath.IsAbs(p)` so every proto/import path begins with `/`, structurally preventing a value that protoc would parse as a flag. Combined with `filepath.Clean(p) == p` (no `..`, no double-slashes) and NUL-byte rejection, the argv is fully sanitised before exec.
- **Path traversal / symlink attack (CWE-22 / CWE-59)**: `filepath.EvalSymlinks` runs **after** the syntactic checks (so a malformed path can never trigger a stat on attacker-controlled territory), and the resolved target must fall under at least one allowed root, enforced via `filepath.Rel` (not `strings.HasPrefix`, which has the classic `/srv/protos-other` false-positive). Allowed roots are the server CWD plus user-supplied `import_paths`, falling back to `filepath.Dir(proto)` when imports are empty — a clearly bounded set. The resolved (symlink-target) path is what gets passed to protoc, so any subsequent re-resolution by protoc itself hits the same inode.
- **Env handling**: `cmd.Env = []string{"PATH=" + os.Getenv("PATH")}` strictly minimises propagation. `cmd.Dir = tmpDir` per-call so protoc cannot read relative paths against an unexpected CWD. The 30-second `context.WithTimeout` caps per-call wall-clock.
- **Binary resolution**: The protoc path resolves only from `YP_PROTOC_BINARY` env → `ProxyConfig.GRPCSchema.ProtocBinary` → `"protoc"`. There is no per-MCP-call override (Resolved #6 explicitly rejected this on RCE-pivot grounds) — confirmed by code inspection of `loadDescriptorFromFiles`. `exec.LookPath` runs first; the resolved absolute path is what `exec.CommandContext` invokes, avoiding the PATH-reresolution race.
- **Resource exhaustion (CWE-770)**: Generated descriptor capped at 16 MiB (`MaxDescriptorSetBytes`), pre-read via `os.Stat`. Per-call tmp dir with `defer os.RemoveAll`. Combined stdout+stderr truncated to 4 KiB on the way into the error string.

### LOW Findings

| S-1 | LOW | internal/encoding/protoschema/protoc_runner.go | 188 | GoSecurity | CWE-770 | `cmd.CombinedOutput()` reads the entire subprocess output into memory before the 4 KiB truncation cap applies. If a misbehaving or malicious protoc emits gigabytes of stderr in under 30 s the proxy process can OOM. The protoc binary itself is operator-installed (so this requires either a tampered system protoc or a protoc bug producing runaway output), and the 30 s timeout bounds the worst case. A targeted fix (capture pipes with `io.LimitReader`) is non-trivial and was not in the design-review scope. Defer-with-issue acceptable; not blocking this PR. |

No CRITICAL, HIGH, or MEDIUM findings. Approving.

---

PR #24, opened 2026-05-20
fix(flow): order GetFlows by timestamp to match wire observation (USK-935)

usk6666:
## Summary
- `GetFlows` now orders by `timestamp ASC, sequence ASC, direction ASC`, matching wire-observed order across gRPC, gRPC-Web, and WebSocket. HTTP is unaffected because its sequence counter is shared across directions.
- Per-direction `Channel.nextSeq` semantics are unchanged at the Layer; the fix is purely at the read-time ORDER BY clause.
- New `schemaV16` adds `idx_flows_stream_id_timestamp` so the planner can stream rows in ORDER BY order without a filesort.
- Adds regression tests for the per-direction sequence collision case (gRPC reproduction from the Issue body) and the WS bidirectional case.

## Why this is a fix, not a feature
The pre-fix ORDER BY relied on `Sequence` alone, but gRPC / gRPC-Web / WebSocket Channels use a per-direction sequence counter (`internal/layer/{grpc,grpcweb,ws}/channel.go`). Both SEND and RECV start at 0, collide, and SQLite has no deterministic tie-break - it returns rows in an arbitrary rowid order. The real-DB reproduction in stream `58d99d96-99ab-4823-a81d-3ff76bafddcc` shows the WebUI rendered RECV(.120) above SEND(.118) for the same sequence value.

Output Filter / Input Filter / variant recording / overlay (h2-frame / grpc-lpm-frame) are all unaffected: every flow row gets a `time.Now()` stamp from `envelopeToFlow`, and the new ORDER BY uses that stamp as the primary key.

## Test plan
- [x] `make lint` passes
- [x] `make build` passes
- [x] `make test` passes (no regressions in `internal/flow/`, `internal/mcp/`, `internal/mcpserver/`)
- [x] `go test -race -tags 'e2e e2e_smoke' ./internal/flow/... ./internal/layer/grpc/... ./internal/layer/ws/... ./internal/layer/grpcweb/...` passes
- [x] New `TestSQLiteStore_GetFlows_WireOrderBeatsSequence` mirrors the Issue's real-DB reproduction and asserts wire-observed order
- [x] New `TestSQLiteStore_GetFlows_WireOrder_DeterministicTieBreak` pins the (sequence, direction) tie-break behavior
- [x] New `TestSQLiteStore_GetFlows_WireOrder_WebSocketInterleaved` exercises a symmetric WS chat scenario where send-seq and recv-seq collide and asserts per-direction filtering remains monotonic (no regression of the per-direction Sequence invariant)
- [x] New `TestMigrate_V16_Idempotent` re-opens an existing DB and confirms the schemaV16 index is created exactly once and `schema_version >= 16`

Resolves USK-935
Linear: https://linear.app/usk6666/issue/USK-935

### DIFF
--- a/internal/flow/schema.go
+++ b/internal/flow/schema.go
@@ -438,6 +438,22 @@ CREATE TABLE IF NOT EXISTS grpc_schemas (
 );
 `
 
+// schemaV16 adds a composite index on (stream_id, timestamp) for the
+// GetFlows wire-order ORDER BY introduced in USK-935. The fix changes
+// GetFlows from `ORDER BY sequence ASC` to
+// `ORDER BY timestamp ASC, sequence ASC, direction ASC` so the read-time
+// order reproduces wire-observed order for gRPC, gRPC-Web, and WebSocket
+// (whose per-direction sequence counters otherwise tie on collide).
+// Without this index, SQLite materializes a per-stream filesort on every
+// read; the index lets the planner stream rows directly in ORDER BY
+// order on the (stream_id, timestamp) prefix.
+//
+// Idempotent additive migration: CREATE INDEX IF NOT EXISTS, no table
+// recreation, no foreign-key toggle needed.
+const schemaV16 = `
+CREATE INDEX IF NOT EXISTS idx_flows_stream_id_timestamp ON flows(stream_id, timestamp);
+`
+
 var migrations = map[int]string{
 	1:  schemaV1,
 	2:  schemaV2,
@@ -454,6 +470,7 @@ var migrations = map[int]string{
 	13: schemaV13,
 	14: schemaV14,
 	15: schemaV15,
+	16: schemaV16,
 }
 
 func migrate(ctx context.Context, db *sql.DB) error {
--- a/internal/flow/sqlite.go
+++ b/internal/flow/sqlite.go
@@ -849,7 +849,26 @@ func (s *SQLiteStore) GetFlows(ctx context.Context, streamID string, opts FlowLi
 		args = append(args, opts.WireLevel)
 	}
 
-	query += " ORDER BY sequence ASC"
+	// USK-935: order by (timestamp, sequence, direction) so the read-time
+	// order matches the wire-observed order. gRPC, gRPC-Web, and WebSocket
+	// Channels use a per-direction sequence counter — sequence values for
+	// SEND and RECV collide (both start at 0), and SQLite's tie-break on
+	// "ORDER BY sequence ASC" alone is non-deterministic. Sorting by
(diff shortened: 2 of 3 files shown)


usk6666 (commented):
## Code Review: APPROVED

The ORDER BY change is correctly scoped to the read-time projection and does not touch wire fidelity (per-direction `Channel.nextSeq` semantics are preserved at the Layer). The new `idx_flows_stream_id_timestamp` is additive and idempotent (`CREATE INDEX IF NOT EXISTS`). The four new tests (wire-order beats-sequence, deterministic tie-break, WS interleaved, migration idempotency) cover the gRPC reproduction from the Issue body, the WebSocket symmetric case, and the migration upgrade path.

### Verified

- The lex-sortability concern around RFC3339Nano trailing-zero truncation was empirically rechecked. `time.Now()` on Linux populates 9-digit nanosecond fractions in >99.99% of calls in my stress test (0 inversions across 10000 back-to-back calls), and the (sequence, direction) tail tie-break correctly handles the residual same-nanosecond case.
- `make build`, `make lint`, and `make test` all pass on the PR head (114f02df) with no regressions.
- The new tests run with `-race` cleanly.

### Findings

None at HIGH/MEDIUM.

#### LOW / NIT

- `LOW (CodeHealth)`: The deterministic tie-break relies on the implicit fact that the string `"receive"` lex-sorts before `"send"` (because `r` (0x72) < `s` (0x73)). This is correct and documented in `TestSQLiteStore_GetFlows_WireOrder_DeterministicTieBreak` line 174, but a one-line comment on the ORDER BY itself (e.g., `// direction ASC: "receive" < "send" lexicographically — see test`) would make the dependency explicit at the point of use. Non-blocking; the implementer's existing comment block already discusses the tie-break behavior generically.

---

usk6666 (commented):
## Security Review: APPROVED

The change is a read-time ORDER BY clause and an additive schema-version-16 composite index. No new user-controlled input flows into SQL, no TLS / certificate / network code path changes, and no new dependencies. The ORDER BY string is a hardcoded constant appended to a parameterized query — SQL injection surface is unchanged.

### Verified

- SQL injection: ORDER BY is a static string literal; existing parameterized `args` array remains untouched.
- Resource handling: No new goroutines, file handles, or connections. Index creation runs inside the existing per-migration transaction.
- Migration safety: `CREATE INDEX IF NOT EXISTS` is idempotent and FK-safe (no table recreation); migration runner correctly omits this version from the `needsFKOff` list.
- Information disclosure: No new error message paths; no new logging.
- MCP / control plane: No tool surface changes. The MCP `query messages` resource benefits from the deterministic ordering without any change to input validation surface.

### Findings

None.

---

PR #33, opened 2026-05-22
feat(mcp): per-iteration upstream proxy rotation for fuzz_http / resend_http

usk6666:
## Summary

Enables residential-proxy IP switching: each `fuzz_http` variant (and each `resend_http` invocation) tunnels through a **distinct** upstream proxy URL, expanded per iteration from a `§var§`-style URL template.

MCP input gains an optional structured field:

```jsonc
{
  "upstream_proxy": {
    "url_template": "http://session-§__nonce§:pass@residential.example:8080"
  }
}
```

The template is re-expanded **per variant / per call** against runtime-reserved variables:
- `§__iteration§` — zero-based variant index
- `§__nonce§` — fresh UUID per iteration

The parsed URL drives the dial through HTTP CONNECT or SOCKS5 (existing `connector.DialViaUpstreamProxy` helper). Reserved variables (`__`-prefix) cannot be shadowed by user-supplied macro hooks (`internal/job/job.go` `mergeKVStore` guard).

## Architecture

| Layer | File | Purpose |
|---|---|---|
| ctx carrier | `internal/connector/upstream_proxy_ctx.go` | `ContextWithUpstreamProxyOverride` / `UpstreamProxyOverrideFromContext`. Distinguishes "present-but-nil = force direct dial" from "absent = fall through to per-listener / global". |
| resolution priority | `internal/connector/stack_builder.go` (`EffectiveUpstreamProxyForCtx`) | Per-flow override now resolves at priority **0**, ahead of per-listener (USK-826) / global (USK-734) / boot-time. |
| dial helper | `internal/connector/upstream_proxy.go` (`MaybeDialViaUpstreamProxy`) | Wraps `net.Dialer` for resend / fuzz; honors ctx override, falls through to direct dial when absent. |
| reserved keys | `internal/macro/reserved.go` + `internal/job/job.go` `mergeKVStore` | `__`-prefix is runtime-reserved. Hook KV results that try to set `__nonce` / `__iteration` are silently dropped. |
| per-iteration resolver | `internal/mcp/upstream_proxy_rotation.go` (`UpstreamProxyConfig.ResolveForIteration`) | Template expand → CR/LF guard (CWE-93) → `ParseUpstreamProxy` → `ContextWithUpstreamProxyOverride`. |
| fuzz_http wiring | `internal/mcp/fuzz_http*.go` | Per-variant resolution; malformed expansion records the variant with `row.Error` and continues (non-fatal — mirrors SafetyFilter violation semantics). |
| resend_http wiring | `internal/mcp/resend_http*.go` | Per-call resolution at `iteration=0`. |

### HTTP/2 connection pool

`internal/connector/h2_pool.go` already keys on the resolved upstream proxy URL string (line 84–97). Because each iteration's URL differs (`§__nonce§` substitution), pool key differs → fresh dial → fresh upstream IP. Verified in design discussion; no change required.

### Security

- **CWE-93 (CRLF injection)**: `ResolveForIteration` rejects expansion results containing `\r` or `\n` before parsing. This sits in front of `buildCONNECTRequest`'s base64-encoded Proxy-Authorization construction.
- **Reserved-key isolation**: `mergeKVStore` filters `__`-prefixed keys from `RunHookFunc` results. A malicious / careless macro cannot shadow `§__nonce§` to pin the run to a single residential IP.
- **Scheme allowlist**: existing `ParseUpstreamProxy` only accepts `http://` and `socks5://`; reused verbatim.

## Out of scope (follow-up issues)

The MVP intentionally targets only `fuzz_http` + `resend_http`. The remaining surface is mechanical to wire and should land as separate issues:

1. `fuzz_ws` / `fuzz_grpc` / `fuzz_raw` — per-variant resolution
2. `resend_ws` / `resend_grpc` / `resend_raw` — per-call resolution
3. `internal/job/job.go` Job runner — auto-inject `§__iteration§` / `§__nonce§` so non-MCP paths benefit
4. Error-driven rotation (`on_error` policy that rolls the nonce only after upstream failures)
5. Live MITM data path — per-listener rotation policy (orthogonal to USK-826 per-listener override)

## Test plan

- [x] **Unit** (`go test -race`):
  - ctx override roundtrip / absent-by-default / explicit-nil / nil-ctx safety
  - `EffectiveUpstreamProxyForCtx` resolution priority (ctx > per-listener > global > boot)
  - `MaybeDialViaUpstreamProxy` direct-dial fallback + HTTP CONNECT path (in-process proxy bridges to target listener; asserts `Proxy-Authorization` header)
  - `IsReservedKey` boundary cases
  - `mergeKVStore` drops `__`-prefixed keys
  - `ResolveForIteration`: nil config, empty template, expansion + ctx propagation, iteration substitution, nonce freshness across calls, scheme rejection, missing port, CRLF guard, SOCKS5 allowed
- [x] **e2e** (`make test-e2e`, build tag `e2e && !e2e_smoke`):
  - `TestFuzzHTTP_UpstreamProxyRotation_PerVariantDistinctNonce` — 3 variants, recording CONNECT proxy observes 3 distinct Basic-auth users with `session-<uuid>` prefix
  - `TestFuzzHTTP_UpstreamProxyRotation_IterationCounter` — usernames are `var-0`, `var-1`, `var-2`
  - `TestFuzzHTTP_UpstreamProxyRotation_MalformedTemplateNonFatal` — `ftp://` template records `row.Error` on every variant; `CompletedVariants == len(payloads)`
  - `TestResendHTTP_UpstreamProxyTunneledViaCONNECT` — single resend tunnels through recording CONNECT proxy with expected userinfo
  - `TestResendHTTP_UpstreamProxyMalformedTemplateRejected` — malformed scheme surfaces as MCP tool error
- [x] **Merge gate** (`make test-e2e-smoke`): no new failures introduced (3 pre-existing environmental failures unrelated: IPv6 loopback unavailable in sandbox, root-user file-permission tests in `internal/mcpserver/`)
- [x] **gofmt / vet** clean

https://claude.ai/code/session_01HMLDtMTzipWGzmpSr4mxgd

---
_Generated by [Claude Code](https://claude.ai/code/session_01HMLDtMTzipWGzmpSr4mxgd)_

### DIFF
--- a/internal/connector/stack_builder.go
+++ b/internal/connector/stack_builder.go
@@ -384,6 +384,11 @@ func (c *BuildConfig) UpstreamProxyForListener(name string) *url.URL {
 // (set by FullListener via ContextWithListenerName) (USK-826).
 //
 // Resolution order:
+//  0. Per-flow override installed via ContextWithUpstreamProxyOverride.
+//     Present-but-nil is honoured as an explicit "direct dial" — the
+//     resolver returns nil without consulting any lower-priority slot.
+//     Used by control-plane resend / fuzz per-iteration rotation
+//     (residential proxy IP switching).
 //  1. Per-listener override installed via SetUpstreamProxyForListener for
 //     the ctx's listener name.
 //  2. Process-global override installed via SetUpstreamProxy.
@@ -397,6 +402,9 @@ func (c *BuildConfig) UpstreamProxyForListener(name string) *url.URL {
 func (c *BuildConfig) EffectiveUpstreamProxyForCtx(ctx context.Context) *url.URL {
 	if c == nil {
 		return nil
+	}
+	if u, present := UpstreamProxyOverrideFromContext(ctx); present {
+		return u
 	}
 	if name := ListenerNameFromContext(ctx); name != "" {
 		c.upstreamProxyPerListenerMu.RLock()
--- a/internal/connector/upstream_proxy.go
+++ b/internal/connector/upstream_proxy.go
@@ -75,6 +75,24 @@ func RedactProxyURL(rawURL string) string {
 	return u.String()
 }
 
+// MaybeDialViaUpstreamProxy dials targetAddr either directly or through
+// the upstream proxy URL carried by ctx (set via
+// ContextWithUpstreamProxyOverride). When ctx carries no override OR an
+// explicit nil URL override, this falls through to a plain net.Dialer.
+//
+// Used by control-plane resend / fuzz dial paths that opt into the
+// per-flow override mechanism without participating in the live MITM
+// per-listener / global resolution chain — those paths historically
+// dialed direct via net.Dialer, so adding this wrapper preserves the
+// existing default behaviour while enabling per-iteration rotation.
+func MaybeDialViaUpstreamProxy(ctx context.Context, targetAddr string, timeout time.Duration) (net.Conn, error) {
+	if u, present := UpstreamProxyOverrideFromContext(ctx); present && u != nil {
+		return DialViaUpstreamProxy(ctx, u, targetAddr, timeout)
+	}
+	dialer := &net.Dialer{Timeout: timeout}
+	return dialer.DialContext(ctx, "tcp", targetAddr)
+}
+
 // DialViaUpstreamProxy dials the target address through the upstream proxy.
 // For HTTP proxies, it sends a CONNECT request and returns the tunneled connection.
 // For SOCKS5 proxies, it uses the golang.org/x/net/proxy package.
--- a/internal/connector/upstream_proxy.go
+++ b/internal/connector/upstream_proxy.go
@@ -39,16 +39,16 @@ func ParseUpstreamProxy(rawURL string) (*url.URL, error) {
 	}
 
 	if u.Host == "" {
-		return nil, fmt.Errorf("upstream proxy URL has no host: %s", rawURL)
+		return nil, fmt.Errorf("upstream proxy URL has no host: %s", RedactProxyURL(rawURL))
 	}
 
 	// Ensure host:port format.
 	host, port, splitErr := net.SplitHostPort(u.Host)
 	if splitErr != nil {
-		return nil, fmt.Errorf("upstream proxy URL must include a port (e.g. %s://host:port): %s", u.Scheme, rawURL)
+		return nil, fmt.Errorf("upstream proxy URL must include a port (e.g. %s://host:port): %s", u.Scheme, RedactProxyURL(rawURL))
 	}
 	if host == "" || port == "" {
-		return nil, fmt.Errorf("upstream proxy URL has empty host or port: %s", rawURL)
+		return nil, fmt.Errorf("upstream proxy URL has empty host or port: %s", RedactProxyURL(rawURL))
 	}
 
 	return u, nil
@@ -85,6 +85,11 @@ func RedactProxyURL(rawURL string) string {
 // per-listener / global resolution chain — those paths historically
 // dialed direct via net.Dialer, so adding this wrapper preserves the
 // existing default behaviour while enabling per-iteration rotation.
+//
+// Note: this intentionally does NOT consult EffectiveUpstreamProxyForCtx,
+// so resend/fuzz dials do not inherit per-listener / global upstream-proxy
+// configuration — they are opt-in via ContextWithUpstreamProxyOverride
+// only.
 func MaybeDialViaUpstreamProxy(ctx context.Context, targetAddr string, timeout time.Duration) (net.Conn, error) {
 	if u, present := UpstreamProxyOverrideFromContext(ctx); present && u != nil {
 		return DialViaUpstreamProxy(ctx, u, targetAddr, timeout)
--- a/internal/connector/upstream_proxy_ctx.go
+++ b/internal/connector/upstream_proxy_ctx.go
@@ -1,0 +1,59 @@
+// upstream_proxy_ctx.go carries the per-flow upstream-proxy URL override
+// for control-plane resend / fuzz dial paths. Used by per-iteration
+// rotation (residential proxy IP switching) where each fuzz variant /
+// resend invocation tunnels through a distinct upstream proxy URL.
+//
+// The override is consulted by EffectiveUpstreamProxyForCtx ahead of the
+// per-listener and process-global resolution chain, so a ctx-attached
+// override always wins. This is distinct from the per-listener override
+// (USK-826), which scopes a URL to a named listener for chained-MITM
+// setups but cannot vary per-iteration.
+package connector
+
+import (
+	"context"
+	"net/url"
+)
+
+type upstreamProxyOverrideCtxKey struct{}
+
+// upstreamProxyOverrideValue distinguishes "ctx carries an explicit nil
+// override (direct dial; skip upstream proxy)" from "ctx carries no
+// override (fall back to per-listener / global resolution)". The
+// wrapper presence is the signal; URL nullity carries the direct-dial
+// semantics.
+type upstreamProxyOverrideValue struct {
+	URL *url.URL
+}
+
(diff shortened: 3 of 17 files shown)


usk6666:
## Code Review: APPROVED

Solid implementation: the present-vs-nil ctx-override semantics are documented and tested, the priority-0 slot in `EffectiveUpstreamProxyForCtx` slots cleanly above USK-826 / USK-734, reserved-key (`__`-prefix) isolation is enforced at the `mergeKVStore` boundary, and the CRLF guard sits in the right place (post-expansion, pre-parse). The H2 pool-key claim in the PR body is correct in the live data path (`poolKeyForH2` reads `EffectiveUpstreamProxyForCtx`), though for resend/fuzz the rotation actually relies on each iteration getting a fresh `net.Dial` rather than the pool key — informational only, not a bug. Only LOW/NIT findings below.

### Findings

| F-1 | LOW | internal/connector/upstream_proxy.go | 88-94 | Architecture | `MaybeDialViaUpstreamProxy` reads `UpstreamProxyOverrideFromContext` directly instead of routing through `EffectiveUpstreamProxyForCtx`. This is documented as intentional (resend/fuzz historically dialed direct), but the consequence is that resend/fuzz cannot opt into a globally configured upstream proxy via the existing `SetUpstreamProxy` / per-listener slots — only the explicit per-call `upstream_proxy` field reaches them. Worth a one-line note in the godoc that this is deliberate asymmetry rather than an oversight; users coming from the live MITM path may otherwise expect the boot-time proxy URL to apply. |
| F-2 | NIT | internal/mcp/fuzz_http_upstream_proxy_rotation_integration_test.go | 21,24,290-292 | CodeHealth | The `var _ = httptest.NewServer`, `var _ = atomic.AddInt32`, `var _ = context.Background` anchors at EOF mask unused imports. `httptest` is genuinely unused; `context` and `sync/atomic` are also only referenced through the anchors. Remove the unused imports (and the anchors) instead of hiding them. |
| F-3 | NIT | internal/connector/upstream_proxy_test.go | 134-152 | CodeHealth | The inline `copy := func(dst, src net.Conn) {...}` shadows the builtin `copy`. Rename to e.g. `bridge` or `splice` to avoid the shadow. The integration test (`fuzz_http_upstream_proxy_rotation_integration_test.go` line 104) does this correctly with `pipe`; apply the same naming here. |
| F-4 | NIT | internal/connector/upstream_proxy_test.go | 200-210 | CodeHealth | `indexOf` re-implements `strings.Index` ostensibly to keep imports minimal, but the test file is already a test file in a `_test.go` that depends on multiple stdlib packages. Just `import "strings"` and use `strings.Index` / `strings.Contains` / `strings.Cut`. |
| F-5 | NIT | internal/connector/upstream_proxy_ctx.go | 1-10 | GoConventions | The new file leads with a multi-line file comment terminated by `package connector` on a separate line. By Go convention, a doc comment immediately preceding `package` becomes the package doc — which the connector package already has elsewhere. The current placement is technically valid (Go merges all comments preceding `package`), but consider either prefixing the comment with `// Package connector ...` style and acknowledging the duplication, or moving the rationale into the type/func godocs and keeping the file header as a one-liner. Aesthetic preference only. |

### Suggestions

- **F-1**: Add to `MaybeDialViaUpstreamProxy`'s godoc: "Note: this intentionally does NOT consult `EffectiveUpstreamProxyForCtx`, so resend/fuzz dials do not inherit per-listener / global upstream-proxy configuration — they are opt-in via `ContextWithUpstreamProxyOverride` only."
- **F-2**: Drop `net/http/httptest` and `sync/atomic` from imports (and `context` if unreferenced after removing the anchors); delete the three `var _ = ...` lines.
- **F-3**: `copy := func(...)` → `bridge := func(...)` or `tunnel := func(...)`.
- **F-4**: Replace `indexOf(string(header), needle)` with `strings.Index(string(header), needle)`, and `indexOf(rest, "\r\n")` similarly. Remove the `indexOf` helper.
- **F-5**: Optional — fold the file-header into per-type/func godocs.

### Strengths worth noting

- Comprehensive unit coverage of the new ctx override (round-trip, absent-by-default, explicit-nil, resolution priority across all four slots).
- e2e gate explicitly asserts **distinct** Proxy-Authorization headers across variants and counts variants under the malformed-template path — the rotation property is verified, not merely the auth-header-present property.
- Reserved-key guard placement at `mergeKVStore` rather than at the macro engine is the right boundary: reserved variables ride entirely on the runtime-controlled write path, so attempting to set them from a Starlark plugin's KV return is a no-op.
- CRLF guard is appropriately defensive even though the credential portion would be base64-encoded downstream — defense in depth.

---

---
_Generated by [Claude Code](https://claude.ai/code)_

usk6666:
## Security Review: APPROVED

The per-iteration upstream proxy rotation feature is well-defended. The CWE-93 guard chain is correctly layered: the post-expansion `strings.ContainsAny(expanded, "\r\n")` check catches literal CR/LF in the expanded template before `url.Parse`, `targetAddr` has its own independent CRLF guard in `dialViaHTTPProxy`, and even percent-decoded CR/LF in userinfo is contained by the base64 boundary in `buildCONNECTRequest` (base64 alphabet excludes CR/LF, so injection into the `Proxy-Authorization` header is structurally impossible). The scheme allowlist, port enforcement, reserved-key isolation, and per-flow ctx scoping all check out. Only minor LOW-severity informational notes below.

### Security Findings (informational)

| S-1 | LOW | internal/connector/upstream_proxy.go | 42, 48, 52 | MCP/API | CWE-209 | `ParseUpstreamProxy` error messages include the full raw URL (with cleartext credentials) on host/port validation failure. This error string is wrapped in `ResolveForIteration` and propagates into `fuzzHTTPVariantRow.Error`, which is persisted to the flow DB and returned through MCP query / export. Since the template is caller-supplied, this leaks the caller's own credentials back to themselves — but persisted records could surface in HAR exports or shared DB snapshots. Consider redacting via `RedactProxyURL` before formatting into error strings. |
| S-2 | LOW | internal/mcp/fuzz_http_helpers.go | 283 | Network | n/a (DoS-adjacent) | The fuzz variant loop calls `waitRateLimit` before `ResolveForIteration`. A user supplying a permanently-malformed template (e.g. `ftp://...` with N=10000 variants) burns N rate-limit slots without sending any wire traffic. Not exploitable by external attackers (control-plane only), but suboptimal. Consider validating the template once before the loop, or moving rate-limit consumption after `ResolveForIteration` succeeds. |

### Remediation

- **S-1**: In `internal/connector/upstream_proxy.go` `ParseUpstreamProxy`, replace `%s` formatting of `rawURL` with `RedactProxyURL(rawURL)` so credentials are masked in error chains that propagate to MCP responses and the flow DB.
- **S-2**: Optional. Either pre-validate the template once outside the variant loop, or reorder the per-iteration steps so `ResolveForIteration` runs before `waitRateLimit`. Low priority since the issue is operator-self-inflicted.

### Verification highlights

- CWE-93 chain verified: literal `\r\n` in template -> rejected by post-expansion guard; percent-encoded `%0d%0a` in userinfo -> decoded by `url.URL.User.Username()` but then base64-encoded into `Proxy-Authorization`, where CR/LF cannot survive the base64 alphabet.
- `targetAddr` is independent of `upstream_proxy.url_template` and retains its own CRLF guard at `dialViaHTTPProxy:119`.
- Reserved-key isolation: `mergeKVStore` drops `__`-prefix keys from hook results (with test coverage in `TestMergeKVStore_DropsReservedKeys`).
- Ctx-override priority-0 placement at `EffectiveUpstreamProxyForCtx` cannot be triggered by wire-side traffic; override is only attached from MCP handler code paths.
- `uuid.NewString()` sources `crypto/rand` (UUIDv4), so nonces are unguessable.
- `MaybeDialViaUpstreamProxy` correctly handles the present-but-nil (explicit direct dial) case and the present + non-nil case; no goroutine or conn leak surface introduced.
- Scheme allowlist (`http`, `socks5`) preserved; missing-port detection preserved.

---

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #34, opened 2026-05-22
feat(proxybuild,connector,config): per-listener upstream proxy rotation for live MITM data path (USK-959)

usk6666:
## Summary

- Brings per-iteration upstream-proxy rotation (USK-954's MCP control-plane surface) to the **live MITM data path**. Each listener can now carry a `url_template + rotation policy`; every outbound dial expands the template against a fresh `§__nonce§` so the upstream observes a rotating client IP.
- Four policies: `per_request` (per outbound TCP dial — h2 is connection-level by construction), `per_connection` (cached by inbound ConnID, released on stack Close), `per_target_host` (LRU 10000 hosts / 1h TTL), `sticky` (one URL per listener lifetime, cleared on reload).
- Polymorphic `upstream_proxy` config: accepts either a literal URL string (legacy USK-826 shape, semantics unchanged) or an object `{url|url_template, rotation:{policy}}`. Polymorphism on the same JSON key via custom `UnmarshalJSON`.
- Fail-closed at dial on malformed templates (no direct-dial bypass — privacy regression class). Stage 1 (probe-expand at `configure_tool` / `proxy_start_tool`) → MCP error. Stage 2 (each live dial) → `slog.Warn` + 502 / REFUSED_STREAM / conn-close.
- `configure_tool` is wired for dynamic tuning: re-applying a rotation drops per-conn / per-host / sticky caches; in-flight conns retain the URL they captured.

## Design decisions (Phase 1.5 — binding)

- **Resolver placement**: per-listener `RotationResolver` slot on `BuildConfig`; consulted *inside* `EffectiveUpstreamProxyForCtxErr` so all 8 dial-path call sites pick it up without per-site edits.
- **Lift**: `UpstreamProxyConfig` + `ResolveForIteration` lifted from `internal/mcp/` → `internal/connector/upstream_proxy_rotation.go`; MCP package keeps a thin `type alias` so the wire schema is unchanged.
- **per_request × h2**: rotates per OUTBOUND TCP DIAL (not per stream — h2 multiplexes on a single dialled conn). Documented in `RotationPolicy` godoc.
- **H2 pool invalidation**: already correct via `poolKeyForH2` URL hash — per_request mints fresh URLs → fresh pool keys → no reuse. No explicit eviction code.
- **`§__iteration§` on live path**: NOT exposed. KV contains only `§__nonce§`.
- **Fail-closed (3 stages)**: probe-expand at config-apply (Stage 1) + per-dial fail-closed (Stage 2). NO direct-dial fallback.
- **Config polymorphism**: string-or-object via custom `UnmarshalJSON` (sibling: `unmarshalTCPForwards`).
- **Manager API**: new `Manager.SetUpstreamProxyRotationForListener(name, *RotationConfig)`. `SetUpstreamProxyForListener(name, string)` unchanged. The two slots are mutually exclusive at write time.
- **Status surface**: `ListenerStatus` gains `UpstreamProxyTemplate` (redacted) + `UpstreamProxyRotationPolicy`. `RedactProxyURL` extended with a lexical fallback so userinfo-bearing templates with `§` chars also redact (the section sign breaks `url.Parse`).

## Files

**New (8)**: `internal/connector/upstream_proxy_rotation.go` · `internal/connector/rotation_host_cache.go` · `internal/connector/upstream_proxy_rotation_test.go` · `internal/connector/upstream_proxy_rotation_integration_test.go` · `internal/config/upstream_proxy.go` · `internal/config/upstream_proxy_test.go` · `internal/mcp/upstream_proxy_input.go` · `internal/mcp/upstream_proxy_input_test.go`

**Modified (16)**: connector stack_builder / context / connection_stack / h2_pool / connect_handler / connect_inner_dispatch / http1_forward_handler / socks5_handler / upstream_proxy · mcp configure_tool / proxy_start_tool / components / upstream_proxy_rotation · config/config.go · proxybuild/manager.go · proxybuild/tcp_forward_tls.go

## Test plan

- [x] Unit: `TestRotationResolver_*` (8 cases — per-policy semantics + nil-receiver + malformed template + fast path)
- [x] Unit: `TestBuildConfigEffectiveUpstreamProxyForCtxErr_*` (resolver vs ctx-override priority)
- [x] Unit: `TestBuildConfigReleaseConnectionState_DropsPerConn` (per-conn cleanup propagation)
- [x] Unit: `TestRotationH2PoolKeyInvalidation_*` (per_request → distinct keys; sticky → stable key)
- [x] Unit: `TestRedactProxyURL_OnTemplate` (lexical-fallback redaction for `§`-bearing templates)
- [x] Unit: config polymorphic UnmarshalJSON + Validate (`internal/config/upstream_proxy_test.go`)
- [x] Unit: MCP polymorphic input + Stage 1 probe (`internal/mcp/upstream_proxy_input_test.go`)
- [x] e2e (`e2e && !e2e_smoke`):
  - `TestLiveProxy_PerRequestRotation_DistinctNonces` — 3 requests, 3 distinct nonces
  - `TestLiveProxy_PerTargetHostRotation_SamePerHost` — same target same nonce, different targets distinct
  - `TestLiveProxy_StickyRotation_FixedForListenerLifetime` — stable across requests, fresh after reload
  - `TestLiveProxy_MalformedTemplate_FailClosed` — upstream sees zero traffic (no direct-dial bypass)
  - `TestLiveProxy_MultiListenerChain_NoRecursion` — USK-826 regression class extended to rotation
  - `TestLiveProxy_NoDelimiterTemplate_FastPath` — non-rotating template still works
  - `TestLiveProxy_PerConnectionRotation_DistinctAcrossClients`
- [x] `make build` ✓
- [x] `make test-fast` ✓
- [x] `make test-e2e-smoke` ✓
- [x] `make lint` (gofmt + vet + staticcheck + ineffassign + gocyclo) ✓

## Notes for reviewers

- Signature migration: `EffectiveUpstreamProxyForCtx` is preserved (returns nil on resolver error — caller is the legacy 8 dial-path read; in practice unreachable now). The new `EffectiveUpstreamProxyForCtxErr` is the canonical API for the migrated dial paths; all 8 live-path callers were moved over and fail closed.
- `ConnectionStack.SetBuildConfig` is the cleanup wiring point — called from `runTLSMITM` after a successful `BuildConnectionStack` so `Stack.Close()` can call `BuildConfig.ReleaseConnectionState(connID)` and drop per-conn rotation entries.
- `ContextWithDialTarget` is the new ctx key for "the target host we're about to dial". Stamped in `BuildConnectionStack`, plain-HTTP forward handler, CONNECT/SOCKS5 passthrough, and TCP-forward TLS dial. Read by `EffectiveUpstreamProxyForCtxErr` only when a rotation resolver is installed.
- `RedactProxyURL` now has a lexical-fallback branch for URLs that don't parse (template with `§` chars). Tested in `TestRedactProxyURL_OnTemplate`.
- `per_target_host` uses the full `host:port` string (not just host) as the cache key so distinct services on the same host get distinct rotation slots.

Resolves USK-959

Linear: https://linear.app/usk6666/issue/USK-959

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -621,8 +621,22 @@ type ProxyConfig struct {
 	// Use UnmarshalJSON for backward-compatible parsing.
 	TCPForwards map[string]*ForwardConfig `json:"tcp_forwards,omitempty"`
 
-	// UpstreamProxy is the upstream proxy URL for proxy chaining.
-	UpstreamProxy string `json:"upstream_proxy,omitempty"`
+	// UpstreamProxy is the upstream proxy URL for proxy chaining (legacy
+	// USK-826 string form). When the config file uses the structured
+	// object form (USK-959), this field holds the URL extracted from the
+	// object's "url" key — or remains empty when "url_template" is set
+	// instead. Use UpstreamProxyStruct for the full structured form
+	// (URL, URLTemplate, Rotation).
+	//
+	// Wire schema: the JSON key "upstream_proxy" is polymorphic — string
+	// (legacy) or object (USK-959). ProxyConfig.UnmarshalJSON dispatches.
+	UpstreamProxy string `json:"-"`
+
+	// UpstreamProxyStruct carries the structured per-listener upstream
+	// proxy config (USK-959). Set by ProxyConfig.UnmarshalJSON when the
+	// "upstream_proxy" JSON value is an object; nil when the legacy
+	// string form is used (UpstreamProxy holds the literal URL then).
+	UpstreamProxyStruct *UpstreamProxyConfig `json:"-"`
 
 	// TLSFingerprint selects the TLS ClientHello fingerprint profile for upstream connections.
 	// Valid values: "chrome" (default), "firefox", "safari", "edge", "random", "none" (standard crypto/tls).
@@ -807,27 +821,37 @@ func (c *ProxyConfig) Validate() error {
 	if err := c.GRPCSchema.Validate(); err != nil {
 		return err
 	}
+	if c.UpstreamProxyStruct != nil {
+		if err := c.UpstreamProxyStruct.Validate(); err != nil {
+			return err
+		}
+	}
 	return nil
 }
 
 // UnmarshalJSON implements json.Unmarshaler for ProxyConfig.
-// It handles backward-compatible deserialization of TCPForwards, which can be
-// either a string value (legacy: "host:port") or a structured ForwardConfig object.
+// It handles backward-compatible deserialization of:
+//   - TCPForwards, which can be either a string value (legacy: "host:port")
+//     or a structured ForwardConfig object.
+//   - UpstreamProxy, which can be either a string (legacy USK-826
+//     literal URL) or an object (USK-959 structured: url|url_template +
+//     rotation).
 func (c *ProxyConfig) UnmarshalJSON(data []byte) error {
 	// Use an alias type to avoid infinite recursion.
 	type proxyConfigAlias ProxyConfig
-	// Temporarily override TCPForwards to capture raw JSON.
-	type proxyConfigWithRawForwards struct {
+	// Temporarily override polymorphic fields to capture raw JSON.
+	type proxyConfigWithRaw struct {
 		proxyConfigAlias
-		TCPForwards json.RawMessage `json:"tcp_forwards,omitempty"`
+		TCPForwards   json.RawMessage `json:"tcp_forwards,omitempty"`
+		UpstreamProxy json.RawMessage `json:"upstream_proxy,omitempty"`
 	}
 
-	var raw proxyConfigWithRawForwards
+	var raw proxyConfigWithRaw
 	if err := json.Unmarshal(data, &raw); err != nil {
 		return err
 	}
 
-	// Copy all fields except TCPForwards.
+	// Copy all fields except the polymorphic ones.
 	*c = ProxyConfig(raw.proxyConfigAlias)
 
 	// Parse TCPForwards with backward compatibility.
@@ -839,7 +863,48 @@ func (c *ProxyConfig) UnmarshalJSON(data []byte) error {
 		c.TCPForwards = parsed
 	}
 
+	// Parse UpstreamProxy: string (legacy) or object (USK-959).
+	if len(raw.UpstreamProxy) > 0 {
+		s, obj, err := unmarshalUpstreamProxy(raw.UpstreamProxy)
+		if err != nil {
+			return fmt.Errorf("upstream_proxy: %w", err)
+		}
+		c.UpstreamProxy = s
+		c.UpstreamProxyStruct = obj
+	}
+
 	return nil
+}
+
+// unmarshalUpstreamProxy parses a JSON value that can be either a
+// string (legacy USK-826 literal URL) or an object (USK-959 structured
+// form with url|url_template + rotation). Returns the legacy string
+// form (populated for both legacy and object-with-url shapes) plus the
+// structured form (nil when the input was a plain string).
+func unmarshalUpstreamProxy(data json.RawMessage) (string, *UpstreamProxyConfig, error) {
+	trimmed := strings.TrimSpace(string(data))
+	if trimmed == "" || trimmed == "null" {
+		return "", nil, nil
+	}
+	if strings.HasPrefix(trimmed, "\"") {
+		var s string
+		if err := json.Unmarshal(data, &s); err != nil {
+			return "", nil, fmt.Errorf("must be a string or object: %w", err)
+		}
+		return s, nil, nil
+	}
+	var obj UpstreamProxyConfig
+	if err := json.Unmarshal(data, &obj); err != nil {
+		return "", nil, err
+	}
+	// Mirror the URL form into the legacy string field so existing
+	// callers (proxy_start_tool defaults merge) keep working without
+	// branching on the polymorphic shape.
+	legacy := ""
+	if obj.URL != "" {
+		legacy = obj.URL
+	}
+	return legacy, &obj, nil
 }
 
 // unmarshalTCPForwards parses a JSON value that can be either:
--- a/internal/config/upstream_proxy.go
+++ b/internal/config/upstream_proxy.go
@@ -11,6 +11,8 @@ package config
 import (
 	"encoding/json"
 	"fmt"
+	"net"
+	"net/url"
 	"strings"
 
 	"github.com/google/uuid"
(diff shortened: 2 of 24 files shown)


usk6666 (commented):
## Code Review: APPROVED

Per-listener upstream-proxy rotation lands cleanly. The `EffectiveUpstreamProxyForCtxErr` migration on all 8 dial-path call sites + parallel non-Err retention for `poolKeyForH2` is well-reasoned; `ContextWithDialTarget` stamping at `BuildConnectionStack` plus the four specialised handler entry points covers the per_target_host scoping surface. `SetBuildConfig` + `Close()` snapshot pattern under `s.mu` is correctly conditional. All lint + fast tests pass.

Four LOW / NIT items below (none blocking — merge-ready):

### Findings

| F-1 | LOW | internal/connector/upstream_proxy.go | 86-89 | CodeHealth | RedactProxyURL lexical-fallback comment says "Bounded by the FIRST '@' so we do not consume past the host segment if the password contains an '@'" — but the FIRST '@' would actually be inside the password in that scenario. Code is conservative/safe (always picks first `@` as host boundary); only the comment is misleading. |
| F-2 | LOW | internal/config/upstream_proxy.go | 303-321 | Correctness | `probeExpandTemplate` populates synthetic `__iteration: "0"` in its KV, but the live-path `RotationResolver.expandFresh` only populates `__nonce`. A template containing `§__iteration§` passes Stage 1 probe-expand and fails at Stage 2 live dial (macro stays literal → ParseUpstreamProxy fails). The binding decision says iteration is "not exposed" on live path, so the probe should be parity-aligned: drop `__iteration` from the probe KV so the template is rejected at config-apply, matching the runtime failure surface. Fail-closed is intact either way. |
| F-3 | NIT | internal/connector/upstream_proxy_rotation_integration_test.go | ~2128, 2137 | TestQuality | `_ = i` dead read in the per_connection client loop; `var _ = atomic.AddInt32` placeholder at file bottom. Minor hygiene. |
| F-4 | LOW | internal/proxybuild/manager.go | 587-612 | Correctness | `SetUpstreamProxyForListener(name, "")` clears the literal slot but does NOT clear the rotation slot for the same name (the `proxyURL == ""` branch falls through without `delete(m.upstreamProxyRotationPerListener, name)` and without `bc.SetRotationForListener(name, nil)`). `ClearUpstreamProxyForListener` does both correctly, but a direct `SetUpstreamProxyForListener(name, "")` call leaves rotation active. Either make the empty branch symmetric or document that direct setters intentionally clear only their own slot. |

### Suggestions

- F-1: Update the comment to read "the FIRST '@' is taken as the host boundary; a password containing an embedded '@' would be partially exposed but the case is rare and `url.Parse` typically handles it on the structured branch above."
- F-2: Drop the `iteration` key from `probeExpandTemplate`'s KV map so `§__iteration§` templates fail Stage 1 with a "unknown variable" macro error.
- F-3: Drop the dead `_ = i` and the unused `atomic.AddInt32` placeholder once the file's imports are verified used by the active tests.
- F-4: Mirror the asymmetric branch — `if proxyURL == "" { delete(m.upstreamProxyRotationPerListener, name); ... bc.SetRotationForListener(name, nil) }`.

STATS: CRITICAL: 0 | HIGH: 0 | MEDIUM: 0 | LOW: 3 | NIT: 1

---

usk6666 (commented):
## Security Review: APPROVED

The privacy-regression class (silent direct-dial bypass on rotation failure) is correctly guarded. All 8 dial-path call sites consume `EffectiveUpstreamProxyForCtxErr` and surface `slog.Warn` + 502 / REFUSED_STREAM / conn-close on failure. Stage 1 probe-expand at `configure_tool` / `proxy_start_tool` blocks malformed templates at the MCP boundary, and the e2e `TestLiveProxy_MalformedTemplate_FailClosed` test asserts the upstream observes zero traffic.

CWE-93 (CRLF injection) is correctly ported from the sibling resend/fuzz path: `expandAndParseTemplate` enforces `strings.ContainsAny(expanded, "\r\n")` after macro expansion, before `ParseUpstreamProxy`. CWE-200 (credential exposure) is handled by the new lexical-fallback branch in `RedactProxyURL` so `§`-bearing templates with userinfo no longer leak via status surfaces. LRU+TTL bound on `rotationHostCache` (10k entries × 1h TTL) prevents unbounded growth from `per_target_host`. `ConnectionStack.Close` releases per-connection rotation state via `BuildConfig.ReleaseConnectionState`, dropping `sync.Map` entries on client disconnect.

### Security Findings

| S-1 | LOW | internal/connector/upstream_proxy.go | 91-109 | InputValidation | CWE-200 | `RedactProxyURL` lexical fallback redacts the password half (between `:` and `@`) but leaves the username (`§__nonce§` substituted value) visible by design. For a template like `http://§__nonce§@proxy.example:8080` (nonce alone, no separate password), nothing is redacted — there is no colon between scheme and `@`. This is intentional (no password to redact) but operators should know that surfacing such templates without an additional redaction layer exposes the nonce-derived session identifier. Acceptable as-is; consider documenting that templates without a separate password should not place secret material in the username position. |
| S-2 | LOW | internal/proxybuild/manager.go | 587-612 | GoSecurity | CWE-362 | `SetUpstreamProxyForListener(name, proxyURL)` clears the literal map and BuildConfig resolver in two phases separated by `m.mu.Unlock()`: under-lock map mutation, then post-lock `bc.SetRotationForListener(name, nil)`. A concurrent reader (live dial) between the two phases observes "rotation cleared in status" but "rotation still resolving in BuildConfig". The runtime result is correct (the literal URL also routes correctly via the BuildConfig per-listener fallback), so this is not exploitable — eventual consistency holds. LOW because it complicates reasoning, not because it leaks. |

### Remediation

- S-1: Add a remark to `UpstreamProxyConfig.URLTemplate` godoc warning that the username position is left visible by status surfaces; operators must use the password position for any secret material.
- S-2: Move `bc.SetRotationForListener(name, nil)` inside the `m.mu`-held block (the `BuildConfig` has its own internal lock, so nested locking is safe — `bc.SetRotationForListener` only touches `c.rotationByListenerMu`, no reverse-lock risk).

STATS: CRITICAL: 0 | HIGH: 0 | MEDIUM: 0 | LOW: 2

---

usk6666:
Addressed all review findings in 23b4442f.

**F-1 LOW (RedactProxyURL comment vs code mismatch)** — Comment now explicitly states the FIRST `@` after the scheme delimiter is the userinfo/host boundary per RFC 3986, matching the code. Also fixed S-1 in the same change.

**S-1 LOW / CWE-200 (RedactProxyURL leaks bare-token userinfo)** — When userinfo has no `:password` colon, the WHOLE userinfo region is now masked (a bare token like a session id or API key is itself a credential). Fix applies to both the `url.Parse` branch (`u.User = url.User("xxxxx")`) and the lexical fallback. Locked in by `TestRedactProxyURL_NoPasswordMasksWholeUserinfo` covering both code paths.

**F-2 LOW (probeExpandTemplate Stage 1 / Stage 2 KV parity)** — `probeExpandTemplate` now populates ONLY `__nonce`, matching the live resolver in `upstream_proxy_rotation.go::expandFresh`. Also tightened the probe to call `url.Parse` + `net.SplitHostPort` (without importing `connector`, avoiding the documented import cycle) so an unsubstituted `§__iteration§` is rejected at config-apply time rather than at first dial. Locked in by `TestUpstreamProxyConfig_Validate_RotationProbeRejectsIterationMacro`.

**F-3 NIT (dead `_ = i` and `var _ = atomic.AddInt32`)** — Both removed. `atomic.Int64` is still legitimately used at line 52, so the `sync/atomic` import remains live.

**F-4 LOW (SetUpstreamProxyForListener empty-string asymmetry)** — Empty-string branch now also clears the rotation slot. The mental model is "this setter takes over the listener's upstream-proxy slot entirely" — both flavours of state are cleared. Existing tests (`TestManager_UpstreamProxy_RoundTrip`, `TestManager_UpstreamProxyForListener_Roundtrip`) still pass.

**S-2 LOW / CWE-362 (race between manager status and BuildConfig)** — `SetUpstreamProxyForListener` now performs the BuildConfig writes inside the same `m.mu` critical section as the status mutations. URL parsing happens BEFORE acquiring `m.mu` so a malformed URL cannot deadlock. Lock ordering is `m.mu` → `bc.upstreamProxyPerListenerMu` / `bc.rotationByListenerMu`; verified no reader of those bc muxes acquires `m.mu`, so no inversion. Logger call moved to after the unlock so a slow log sink cannot stall other manager callers.

Verification: `make build`, `make lint`, `make test`, `make test-e2e-smoke`, plus targeted `go test -tags e2e ./internal/connector/...` for the rotation suite — all green.

PR #36, opened 2026-05-22
fix(webui/mcp): harden callTool fallback and export missing hooks

usk6666:
## Summary
- `callTool()` now throws a descriptive Error (`"MCP tool '<name>' returned no parseable content"`) when the MCP server returns empty or non-text content, instead of silently casting the raw `ToolResult` to `T` and triggering downstream `JSON.parse` crashes or property-undefined surprises.
- `index.ts` re-exports `usePluginIntrospect`, `useGrpcSchema`, `useGrpcSchemaList` and their `Use*Result` types so `import { ... } from "@/lib/mcp"` resolves them correctly (they were exported from `hooks.ts` but never re-exported from the barrel).
- Added `web/src/lib/mcp/client.test.ts` covering the parse path: valid JSON-text block, missing/empty content, non-text first block, and `isError` propagation.

## Why throw instead of `Promise<T | null>`
Every existing caller of `callTool<T>` (all inside `client.ts`'s typed wrappers) treats the return value as the typed result and immediately accesses fields. Switching to `Promise<T | null>` would force every hook and React-Query caller to handle `null` everywhere. An empty/non-text content response from yorishiro-proxy's own MCP server is a contract violation — throwing is the right signal, and React Query's `useQuery` surfaces it via `error` natively.

## Test plan
- [x] `make test-ui` — 12 files / 240 tests pass (includes the 5 new `callTool` cases)
- [x] `make build` — TypeScript compiles clean; vite + go build succeed
- [x] `make lint` — gofmt / vet / staticcheck / ineffassign / gocyclo all clean
- [x] `go test ./internal/mcp/` — passes (initial `make test` flaked on a 4m sqlite timeout; rerun with 8m timeout passes — unrelated to this WebUI change)
- [x] Grep confirms no caller of `callTool` outside `client.ts` relied on the previous silent-cast behavior

Resolves USK-966
Linear: https://linear.app/usk6666/issue/USK-966

### DIFF
--- a/web/src/lib/mcp/client.test.ts
+++ b/web/src/lib/mcp/client.test.ts
@@ -1,0 +1,76 @@
+/**
+ * Tests for McpClient.callTool's content-block parsing contract (USK-966).
+ *
+ * yorishiro-proxy's MCP tools return a single JSON-text ContentBlock. If the
+ * server returns no content, an empty content array, or a non-text first
+ * block, that is a contract violation. callTool MUST throw an explicit error
+ * with the tool name — the previous behaviour silently cast the raw
+ * `ToolResult` to `T`, which propagated to callers and triggered
+ * `JSON.parse("")` crashes or property-undefined surprises downstream.
+ */
+
+import { describe, expect, it } from "vitest";
+import { McpClient } from "./client.js";
+
+/**
+ * Build an McpClient and inject a fake underlying MCP SDK client that
+ * returns the supplied `ToolResult` payload from `callTool`. Bypasses the
+ * real transport / handshake — we only exercise the result-parsing path.
+ */
+function clientReturning(result: unknown): McpClient {
+  const c = new McpClient({ url: "/mcp" });
+  // Cast to a record so we can poke at the private fields exercised by
+  // callTool's guard. `callTool` only checks `this.client` truthiness and
+  // `this._status === "connected"`; nothing else from the SDK is touched in
+  // the parsing path.
+  const internal = c as unknown as {
+    client: { callTool: (req: unknown) => Promise<unknown> } | null;
+    _status: string;
+  };
+  internal.client = {
+    callTool: async () => result,
+  };
+  internal._status = "connected";
+  return c;
+}
+
+describe("McpClient.callTool", () => {
+  it("parses a single JSON-text content block", async () => {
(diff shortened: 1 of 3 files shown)


usk6666 (commented):
## Code Review: APPROVED

`callTool` now throws a descriptive Error on empty/non-text content instead of silently casting `ToolResult` to `T`, which is the right contract signal for a MCP-server-side violation; existing typed wrappers all access fields immediately, so `Promise<T | null>` would have churned every caller. The barrel-export additions in `index.ts` align the documented surface with what `hooks.ts` actually exports (other hooks like `useConfigure` / `useQuery` are already re-exported the same way), making `import { usePluginIntrospect, useGrpcSchema, useGrpcSchemaList } from "@/lib/mcp"` resolve. The new `client.test.ts` covers the 5 relevant parsing branches (happy path + 3 contract-violation paths + isError propagation) and asserts the tool-name interpolation so regressions in the error format are caught.

### NIT (optional, no action required)

- `client.test.ts` does not cover the `isError: true && content missing` path that falls back to `"Tool call failed"`. Behaviour is unchanged by this PR — pre-existing default — so this is purely additive coverage if you want it.

---

usk6666 (commented):
## Security Review: APPROVED

WebUI TypeScript only — TLS / network upstream-dial / Go race / MCP server-side input-validation categories are N/A. Focus areas reviewed:

- **Error message info-leak**: The new throw `MCP tool '${name}' returned no parseable content` interpolates only the tool name, which is a literal under code control (`"resend_http"`, `"fuzz_grpc"`, etc., from typed wrappers in `client.ts`). No path, stack trace, session ID, credential, or attacker-controlled input flows into the message. Rendered safely via React (JSX-escaped).
- **Export surface**: `usePluginIntrospect` / `useGrpcSchema` / `useGrpcSchemaList` and their Result types are already exported from `hooks.ts` and reachable via deep imports today; the barrel re-export does not expand the public API beyond what is already reachable. No new attack surface.
- **Dependencies**: No new deps. `vitest` already in devDependencies (used by `hooks.test.ts` and `dispatch.test.ts`).
- **`isError` propagation branch unchanged**: still surfaces `result.content[0].text` from the server. Pre-existing behaviour; unrelated to this PR.

No findings.

---

PR #37, opened 2026-05-22
fix(webui/Resend): populate WS fin from loaded flow and validate raw patch offset

usk6666:
## Summary

- **WS resend**: `fin` initial state derived from the loaded flow's first send-direction frame metadata (`ws_fin`), no longer hardcoded to `true`. The `WsRequestEditor` checkbox was already controlled — the fix is on the populate path so a frame loaded with `FIN=0` round-trips faithfully instead of silently flipping to `FIN=1`.
- **RawPatchEditor**: non-numeric / empty / negative / fractional offset surfaces an inline error (`aria-invalid` + red text), the parent's "Send Raw" button is disabled while invalid via a new `onValidityChange` callback prop. Pure helper `validateRawPatchOffset` extracted for unit testing.
- Removed silent `?? 0` fallback at submit (replaced with disabled-gate + defense-in-depth `throw` inside the existing try/catch so a future regression surfaces as a toast rather than an off-by-N-byte patch).
- Added 12 unit tests for `validateRawPatchOffset` covering the bug class (`"abc"`, `""`, `"-1"`, `"1.5"`, `"1000abc"`, `"Infinity"`, `"NaN"`) and the happy path (`"0"`, integers, surrounding whitespace).

## Decisions

1. **Initial fin state** — kept `boolean` type and initialise from loaded flow (default `true` for brand-new sends, set from `flow.message_preview[*].metadata.ws_fin` on load). No tri-state — the user mental model stays binary.
2. **`?? 0` removal** — replaced with disabled-button + assertion-throw, both small. The throw is now inside the try block so unexpected gate failures produce a user-visible toast instead of an uncaught promise rejection.
3. **Offset validation** — `Number()` + `Number.isFinite` + `Number.isInteger` + non-negative check. Rejects `"abc"`, `""`, `"-1"`, `"1.5"`, and `"1000abc"` (which `parseInt` would have silently truncated to `1000`).

## Test plan

- [x] FIN=0 WS frame round-trip: `extractFlowFin()` reads `flow.message_preview[*].metadata.ws_fin === "false"` → editor populates with `fin: false` → `WsRequestEditor` shows unchecked checkbox → `buildResendWsParams` forwards `fin: false` to `resend_ws` (see `web/src/pages/Resend/typedDispatch.ts:116`).
- [x] `parseInt("abc")` rejection: `validateRawPatchOffset("abc")` returns `"offset must be a number"`. Editor renders inline error with `role="alert"`, marks input `aria-invalid="true"`, calls `onValidityChange(false)`, parent disables Send Raw with a tooltip.
- [x] `make build` passes (TypeScript compile clean, 461 modules transformed).
- [x] `pnpm test` passes (12 files, 247 tests including 12 new ones for `validateRawPatchOffset`).
- [x] `make lint` passes (no new gofmt/vet/staticcheck violations).

Resolves USK-967
Linear: https://linear.app/usk6666/issue/USK-967

### DIFF
--- a/web/src/pages/Resend/RawPatchEditor.css
+++ b/web/src/pages/Resend/RawPatchEditor.css
@@ -111,3 +111,14 @@
   color: var(--accent-danger);
   background-color: rgba(248, 81, 73, 0.1);
 }
+
+/* Inline validation error for the offset field (USK-967). */
+.raw-patch-editor-error {
+  font-size: var(--font-size-xs);
+  color: var(--accent-danger);
+  padding-left: var(--space-xs);
+}
+
+.raw-patch-editor-field[aria-invalid="true"] {
+  border-color: var(--accent-danger);
+}
--- a/web/src/pages/Resend/RawPatchEditor.test.ts
+++ b/web/src/pages/Resend/RawPatchEditor.test.ts
@@ -1,0 +1,82 @@
+/**
+ * Tests for the RawPatchEditor offset validation helper (USK-967).
+ *
+ * The editor previously coerced any non-numeric / empty / negative
+ * offset to 0 via `parseInt(val, 10)` + `?? 0`, which writes the
+ * patch at a wholly different position than the user intended —
+ * exactly the byte-level fuzz failure mode this editor exists to
+ * support. The pure `validateRawPatchOffset` helper is the gate that
+ * surfaces invalid input as a typed error string for inline display
+ * and disables the parent's submit button via onValidityChange.
+ *
+ * The component itself is JSX-heavy; tests for it would pull in
+ * jsdom + RTL. Following the pattern of grpcweb-display.test.ts, we
+ * verify the pure helper here.
+ */
+
+import { describe, expect, it } from "vitest";
+import { validateRawPatchOffset } from "./RawPatchEditor.js";
+
+describe("validateRawPatchOffset", () => {
+  describe("invalid inputs", () => {
+    it("rejects empty string", () => {
+      expect(validateRawPatchOffset("")).not.toBeNull();
+    });
+
+    it("rejects whitespace-only string", () => {
+      expect(validateRawPatchOffset("   ")).not.toBeNull();
+    });
+
+    it('rejects non-numeric "abc"', () => {
+      const err = validateRawPatchOffset("abc");
+      expect(err).not.toBeNull();
+      expect(err).toMatch(/number/);
+    });
+
+    it("rejects negative integer", () => {
+      const err = validateRawPatchOffset("-1");
+      expect(err).not.toBeNull();
+      expect(err).toMatch(/non-negative/);
+    });
+
+    it("rejects fractional value", () => {
+      const err = validateRawPatchOffset("1.5");
+      expect(err).not.toBeNull();
+      expect(err).toMatch(/integer/);
+    });
+
+    it("rejects mixed input (1000abc)", () => {
+      // Number("1000abc") returns NaN; parseInt would have truncated to
+      // 1000 silently — verify the new validator rejects it cleanly.
+      expect(validateRawPatchOffset("1000abc")).not.toBeNull();
+    });
+
+    it("rejects Infinity", () => {
+      expect(validateRawPatchOffset("Infinity")).not.toBeNull();
+    });
+
+    it("rejects NaN literal", () => {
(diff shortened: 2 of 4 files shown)


usk6666 (commented):
## Code Review: APPROVED

The PR cleanly addresses both bugs called out in USK-967: WS `fin` is now extracted from the loaded flow's first send-direction frame metadata (`ws_fin`, formatted by `record_step.go:1231` as `strconv.FormatBool`), and the raw patch offset field is gated by the new `validateRawPatchOffset` pure helper, surfacing inline errors and disabling the parent Send button via the new `onValidityChange` prop. The `?? 0` silent coercion is replaced with a disabled-button gate plus a defense-in-depth `throw` inside the existing try/catch. Tests (12 new) cover the documented bug class (`"abc"`, `""`, `"-1"`, `"1.5"`, `"1000abc"`, `"Infinity"`, `"NaN"`) and happy paths.

`pnpm test` passes (12 files, 247 tests including the 12 new ones). `tsc -b` passes. ESLint shows no new findings on the three changed `.tsx`/`.ts` files (existing lint errors in unrelated files predate this PR).

### LOW / NIT Findings

| F-1 | LOW | web/src/pages/Resend/ResendPage.tsx | 1111-1113 | CodeHealth | Comment "matching the prior behaviour" is misleading — prior code emitted `{offset: 0, data: ""}` (no-op patch entry) for find/replace-mode patches; new code skips them entirely. Functionally equivalent but the array shape differs. |
| F-2 | LOW | web/src/pages/Resend/ResendPage.tsx | 1089-1094 | Architecture | Mode detection logic is duplicated from `detectMode` in `RawPatchEditor.tsx:43-47`. Rule-of-Two — consider exporting `detectMode` to share with the submit path. |
| F-3 | LOW | web/src/pages/Resend/ResendPage.tsx | 1115-1121 | Correctness | When the user enters `find_replace_*`-mode patches only, `params.patches` is `undefined` with no UI feedback that the selections were silently ignored. This pre-dates the PR (prior code emitted no-op offset entries) but the PR description acknowledges the silent-drop; a non-blocking toast would close it. |
| F-4 | NIT | web/src/pages/Resend/RawPatchEditor.tsx | 71-85 | CodeHealth | `offsetErrors` could be wrapped in `useMemo`; for typical patch counts (<10) the cost is negligible, so this is style-only. |
| F-5 | NIT | web/src/pages/Resend/RawPatchEditor.tsx | 95-175 | Architecture | The editor now keeps `offsetInputs[index]` and `patches[index].offset` in sync across `handleModeChange`, `handleRemove`, and `handleOffsetChange`. Currently correct, but a long-term maintenance concern. |

LOW findings are not gating. Per `/review-gate` policy this PR is APPROVED.

---

usk6666 (commented):
## Security Review: APPROVED

This PR is WebUI TypeScript only — no data path, no Go changes, no TLS/CA touch, no SQL surface, no new dependencies. The change is a wire-fidelity hardening (preserves observed FIN=0 on round-trip instead of silently flipping to FIN=1; rejects non-numeric / negative / fractional offset input instead of silently coercing to byte 0), which strengthens MITM correctness rather than weakening it.

### Checked Surface

- **TLS / Certs**: no changes.
- **Network**: no changes.
- **Input Validation**: The new `validateRawPatchOffset` is itself an input-tightening pass. `Number("Infinity")` and `Number("NaN")` are correctly rejected via `Number.isFinite` / `Number.isInteger`. `Number("1e20")` parses to a finite double larger than `Number.MAX_SAFE_INTEGER` and would be accepted — backend bounds-check is the defense here (out of scope for this PR).
- **Go Security**: N/A (no Go changes).
- **MCP / API**: The defense-in-depth `throw new Error("resend_raw: invalid offset reached submit; ...")` is caught by the surrounding try/catch and surfaced as a toast — no stack trace leakage.
- **Dependencies / Licenses**: no new dependencies.

### MITM-Specific Notes

- Wire fidelity is *improved*: FIN=0 round-trips faithfully rather than silently flipping to FIN=1; garbage offset input fails closed instead of silently writing at byte 0.
- No CRLF / header-injection surface added.
- No XSS risk — error messages render via React text interpolation (auto-escaped) into a `role="alert"` div.
- `aria-invalid` and `role="alert"` are correctly bound.

### Findings

FINDINGS: None

STATS:
  CRITICAL: 0
  HIGH: 0
  MEDIUM: 0
  LOW: 0

---

PR #38, opened 2026-05-22
feat(mcp/fuzz_http,macro,flow): per-iteration pre/post macro hooks on fuzz_http (USK-960)

usk6666:
## Summary

- Add `pre_macro` / `post_macro` config to `fuzz_http` for per-iteration macro dispatch around each fuzz variant
- Re-attach dormant `hookExecutor` scaffold (orphaned since PR #688) as the dispatch engine; rename `pre_send` / `post_receive` JSON keys to match Issue body
- Reuse `__response_status` / `__response_body` reserved variables; add per-header `__response_headers__<lower(name)>__` injection (header name lowercased for stable kvStore key — wire copy unaffected)
- New `fuzz_macro_results` table (schemaV17) records pre/post outcomes keyed by `(fuzz_id, index_num, hook_name)`
- Per-iteration kvStore now owned by variant loop (lifted from `ResolveForIteration` via new `ResolveForIterationWithKV` / `SeedIterationKV` pair) — the same kvStore threads through upstream-proxy template expansion, pre_macro extracts, fuzz body §var§ expansion, and post_macro response injection
- New `Engine.RunWithKV(ctx, m, kvStore)` accepts a caller-owned KV Store so extracts land in place across multiple macro invocations within the same iteration
- `scope="job"` deferred to USK-961 — rejected at validation in this PR with a "deferred to follow-up: USK-961" diagnostic
- `on_error="continue"` with unresolved `§var§` sends literal token on wire + records unresolved-token list in `fuzz_results.error`

## Test plan

- [x] `make lint` clean
- [x] `make build` clean
- [x] `make test` clean (incl. renamed `hooks_test.go` tests after JSON-key rename)
- [x] `make test-e2e-smoke` clean
- [x] New `fuzz_http_macro_integration_test.go` (5 tests, e2e tier) covers:
  - Golden iteration: pre extracts a token → fuzz body uses §token§ → post sees `__response_status` / `__response_body` / `__response_headers__<lower(name)>__`
  - `on_error=skip`: pre fails, fuzz variant NOT sent, post NOT run, `fuzz_macro_results` records skipped status
  - `on_error=continue` with unresolved §var§: literal sent on wire + diag in fuzz_results.error
  - Unresolved §var§ in position payload: literal lands on wire, diagnostic surfaces on row
  - `scope="job"` rejection at validation with USK-961 message

Resolves USK-960
Linear: https://linear.app/usk6666/issue/USK-960
Follow-up: USK-961 (scope="job")

### DIFF
--- a/internal/connector/upstream_proxy_rotation.go
+++ b/internal/connector/upstream_proxy_rotation.go
@@ -68,6 +68,11 @@ type UpstreamProxyConfig struct {
 // Errors are returned with the user-facing prefix
 // "upstream_proxy.url_template" so MCP tool handlers can surface them
 // verbatim.
+//
+// Each call mints a throwaway kvStore seeded with §__iteration§ /
+// §__nonce§. Callers that need to share the per-iteration kvStore with
+// other consumers (e.g. fuzz_http's pre/post macro hooks per USK-960)
+// should use ResolveForIterationWithKV instead.
 func (c *UpstreamProxyConfig) ResolveForIteration(ctx context.Context, iteration int) (context.Context, error) {
 	if c == nil || c.URLTemplate == "" {
 		return ctx, nil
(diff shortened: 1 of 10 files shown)


usk6666 (commented):
## Code Review: APPROVED

USK-960 is well-scoped: dormant `hookExecutor` scaffold is re-attached as the engine for per-iteration `pre_macro` / `post_macro` on `fuzz_http`, with a shared per-iteration kvStore threading reserved keys / pre extracts / response-injection through to post. JSON-key rename (`pre_send` → `pre_macro`, `post_receive` → `post_macro`) is clean since the scaffold had no live callers. `Engine.RunWithKV` preserves `Run` semantics (`vars` still wins over `InitialVars`) and the caller-owned kvStore correctly takes precedence over hook config `Vars`. `fuzz_macro_results` schema V17 is migration-registered and uses parameterised SQL. Tests pass under `-race`. Findings are all LOW or NIT — no CRITICAL / HIGH / MEDIUM correctness issues.

### Findings

| F-1 | LOW | internal/mcp/fuzz_http_helpers.go | 444, 459 | Correctness | `row.Payloads` returned in-memory is the *expanded* `expandedPayloads` (set inside `runFuzzHTTPSingleVariant`), but `saveFuzzHTTPResult` is called with the *unexpanded* `payloads`. Pre-USK-960 they were identical (no expansion). Post-USK-960 the MCP in-memory variant row and the persisted `fuzz_results.payloads` JSON diverge when any §var§ resolves in a position payload. |
| F-2 | NIT | internal/mcp/fuzz_http_helpers.go | 329-346 | CodeHealth | Comment says "pre_macro state ('once', 'every_n', 'on_error') tracks across the whole fuzz run" but `hookExec.updateState(...)` is never called from the variant loop and `fuzzHTTPMacroConfig` forces `RunInterval: "always"` on construction. The state-tracking comment is aspirational at best; the scaffold field `preMacroExecuted` / `requestCount` / `lastStatusCode` stay at zero values forever. |
| F-3 | NIT | internal/mcp/fuzz_http_helpers.go | 461-464 | CodeHealth | `_ = preState` + the "guaranteed != fuzzHTTPPreSkipped" comment is dead since the skip branch already short-circuited via `return "", nil` at line 425. The line and the explanatory comment together signal a design constraint that the code already enforces — either drop `_ = preState` or convert the invariant into a single early-return shape. |

### Suggestions

- **F-1**: Pick one of:
  - Save `expandedPayloads` to `fuzz_results.payloads` so DB and in-memory agree (and re-running a fuzz_results row faithfully repros what hit the wire).
  - Set `row.Payloads = payloads` (unexpanded) before calling `saveFuzzHTTPResult`, then expand only on the wire path — DB and in-memory both show the operator-supplied payload, and the unresolved-token diagnostic on `row.Error` is the only signal that expansion happened.
  - Document the asymmetry in `saveFuzzHTTPResult`'s docstring as intentional (operator input vs. wire view).
- **F-2**: Either delete the "tracks across the whole fuzz run" paragraph, or wire a `hookExec.updateState(row.StatusCode, runErr != nil)` call after the variant send so the dormant state-tracking fields actually populate (preparing the ground for USK-961 `scope="job"`).
- **F-3**: Remove `_ = preState` and the explanatory comment block; the function-internal invariant is already enforced by the earlier `return "", nil` branch.

---

usk6666 (commented):
## Security Review: APPROVED

PR is a control-plane change (`internal/mcp/`, `internal/macro/`, `internal/flow/`) plus one helper addition in `internal/connector/upstream_proxy_rotation.go` (which is itself an MCP-tool-driven helper, not on the wire data path). No data-path MITM concerns surfaced. The 64 KiB `__response_body` cap (CWE-770) is well-placed, `fuzz_macro_results` SQL is fully parameterised, and the `on_error="continue"` literal-passthrough is consistent with the documented `fuzz_http` "Payload passthrough" policy (operator-driven, by-design CRLF/smuggling fuzz). The header-name lowercasing into `__response_headers__<lower(name)>__` is documented as intentional last-value-wins kvStore convenience; the wire-recorded flow keeps original casing untouched per MITM principle.

### Findings

| S-1 | LOW | internal/mcp/hooks.go | 318-337 | Network | CWE-770 | `injectResponseVars` caps `__response_body` at 64 KiB, but per-header values (`__response_headers__<lower(name)>__`) are copied verbatim from the recorded `last.Headers` map. A malicious upstream returning very long single-header values, or a high count of distinct headers, would push the full amount into kvStore. Headers are already bounded by the proxy's existing read-side limits and the recorded flow size, so in practice this is unlikely to be exploitable, but a defense-in-depth per-header-value cap and/or per-iteration kvStore size cap would make this explicit. |

### Remediation (LOW)

- **S-1**: Apply a per-value cap to header injection (e.g., reuse `maxResponseBodyKVBytes` as the per-header value ceiling, plus a hard limit on the total number of header keys injected per iteration). Document the cap in `injectResponseVars`'s docstring alongside the existing `__response_body` cap rationale.

### MITM Compatibility Triage

No findings overlap with MITM data-path principles (no header dedup suggestion, no Host=URL enforcement, no canonicalisation in wire-touching code). The `fuzz_http` "Payload passthrough" CRLF/literal-§var§-wire policy is explicitly documented as a pentest-tool feature and is the correct behaviour for this codebase.

### TLS / CA / Crypto

No changes to TLS, certificate issuance, or crypto material. Out of scope.

---

PR #39, opened 2026-05-22
fix(webui/Intercept): clear selectedId and await refetch on release/drop/forward

usk6666:
## Summary
- Release / Modify & Forward / Drop handlers now `await refetchQueue()` BEFORE clearing `selectedId`, eliminating the brief stale-detail-panel render after an action.
- New `isActionInFlight` state disables the three action buttons while a mutation + refetch is in flight, and the handlers short-circuit via early return — both layers prevent double-fire from rapid taps.
- `selectedId` is cleared only on success (inside `try`, after the `await`). The `finally` block resets `isActionInFlight` regardless of outcome.

## Test plan
- [x] `make build` passes (TypeScript compile clean, vite build clean, `go vet` clean)
- [ ] Rapid release-button taps: only one action fires, button visibly disabled until detail panel clears
- [ ] With simulated slow backend (e.g., 1s mutation delay): button remains disabled until refetch completes, then panel clears cleanly
- [ ] Release/Modify-and-Forward/Drop failure path: error toast appears, `isActionInFlight` is reset (button re-enables), `selectedId` preserved so the user can retry

## Notes for reviewer
- The early-return guard `if (!selectedId || isActionInFlight) return;` provides defense-in-depth alongside the `disabled` prop in case a button click fires during a momentary disabled-state delay.
- The dependency arrays for `useCallback` include `isActionInFlight` so the handlers see the latest value of the guard.
- No backend changes; this is purely a frontend race-condition fix.

Resolves USK-968
Linear: https://linear.app/usk6666/issue/USK-968

### DIFF
--- a/web/src/pages/Intercept/InterceptPage.tsx
+++ b/web/src/pages/Intercept/InterceptPage.tsx
@@ -46,6 +46,12 @@ export function InterceptPage() {
   // Editable raw bytes (raw mode, Base64)
   const [editRawBytes, setEditRawBytes] = useState("");
 
+  // True while a release/drop/modify_and_forward mutation is awaiting its
+  // post-mutation refetchQueue. Drives `disabled` on the action buttons so
+  // a double-tap before the server-side hold queue updates can not fire a
+  // second action against the same (now-released) intercept_id (USK-968).
+  const [isActionInFlight, setIsActionInFlight] = useState(false);
+
   const { addToast } = useToast();
   const { interceptAction, loading: executeLoading } = useInterceptAction();
 
@@ -117,8 +123,9 @@ export function InterceptPage() {
 
   // Release: forward as-is
   const handleRelease = useCallback(async () => {
-    if (!selectedId) return;
+    if (!selectedId || isActionInFlight) return;
     const phase = selectedEntry ? resolvePhase(selectedEntry) : "request";
+    setIsActionInFlight(true);
     try {
       await interceptAction({
         action: "release",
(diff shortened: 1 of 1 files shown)


usk6666 (commented):
## Code Review: APPROVED

The fix matches the Issue spec exactly across all three handlers (`handleRelease`, `handleModifyAndForward`, `handleDrop`):

- `await refetchQueue()` happens BEFORE `setSelectedId(null)`, eliminating the stale-detail-panel render window.
- `selectedId` is cleared only on success (inside `try` after the `await`); preserved on failure so the user can retry.
- `setIsActionInFlight(false)` runs in `finally` regardless of outcome — the in-flight flag is always reset.
- Buttons get `disabled={executeLoading || isActionInFlight}`, plus an early-return guard `if (... || isActionInFlight) return;` as defense-in-depth.
- `useCallback` dependency arrays correctly include `isActionInFlight` in all three handlers; no missing closures.
- `handleModifyAndForward` is restructured cleanly: the previously-duplicated `setSelectedId(null) + refetchQueue()` epilogue across raw/structured branches is consolidated into one path. Good refactor.

No issues found. Comment blocks above the in-flight state and above each `await refetchQueue()` explain the rationale and reference USK-968.

Manual verification items (rapid taps, slow backend, failure path) are correctly documented in the PR Test plan as unchecked manual checks — appropriate for a page-level component.

---

usk6666 (commented):
## Security Review: APPROVED

This PR is WebUI TypeScript only (single-file change to `web/src/pages/Intercept/InterceptPage.tsx`). It restructures three React event handlers to await a refetch before clearing local state and adds an in-flight guard to disable action buttons during the mutation+refetch window.

- No new MCP tool calls or new params sent — the same `interceptAction({ action, params })` invocations as before.
- No new network behavior, no TLS/cert touch, no Go code, no new deps.
- The fix narrows (rather than widens) the client-side attack surface: it removes the UX window in which a user could double-fire an `intercept_action` against an already-released `intercept_id`.
- Error toast text continues the pre-existing `err.message ?? String(err)` pattern; no new information-disclosure surface.

No security-relevant findings.

---

PR #40, opened 2026-05-22
fix(webui/Flows): reset pagination offset on filter change

usk6666:
## Summary

Fixes a 1-render race in `FlowsPage` where changing a debounced text filter (host, URL pattern, body contains, tag) at a high page number briefly fetched the new filter with the stale offset before the reset arrived, sometimes producing a flash of an empty page.

The non-debounced filters (protocol radio, method, status, state, scheme, blocked-by) already reset `offset` to 0 synchronously via `handleFilterChange()` in their change handlers. The four debounced text filters were the holdout: they reset via a `useRef`/`useEffect` pair that fired *after* the render in which the new debounced filter committed, so `useQuery` saw new filter + old offset once before the reset re-rendered.

Fix: the four text-input `onChange` handlers now also call `handleFilterChange()` synchronously. The user types one keystroke, `offset` becomes 0 immediately, and the actual MCP query stays debounced (300 ms) and refires with `offset=0` once. Removed the now-unused `prevTextFilterKey` ref and the corresponding `useEffect`; dropped the `useRef` import.

After this change, every filter-state setter on the page resets `offset` and `selectedIds` in the same render, so the next `useQuery` call always sees the new filter alongside `offset=0`.

## Files changed

- `web/src/pages/Flows/FlowsPage.tsx` — only file touched

## Notes for reviewer

- **URL sync**: not in play. The page uses local `useState` for filters and pagination; there is no `useSearchParams` / URL-state sync, so the fix is purely in-component state. (Confirmed with `grep useSearchParams` across `web/src/pages/Flows/`.)
- **Filter setter audit**: every filter setter on the page now resets offset:
  - Radio/select handlers (`selectProtocol`, `handleMethodChange`, `handleStatusCodeChange`, `handleStateChange`, `handleSchemeChange`, `handleBlockedByChange`) — already called `handleFilterChange`, untouched.
  - Text input handlers (`handleHostFilterChange`, `handleUrlChange`, `handleBodyContainsChange`, `handleTagFilterChange`) — **now** call `handleFilterChange`.
  - Sort and page-size handlers also reset offset (pre-existing).
- **Test gap**: the WebUI test suite deliberately avoids RTL/jsdom (see the header comment in `web/src/pages/Resend/ResendPage.dispatch.test.tsx` — tests target pure-logic helpers, not component renders). The fix is interleaved with React hooks inside the component body; extracting a testable unit would require refactoring beyond this bug-fix scope. Manual verification is in the test plan below.

## Test plan

- [x] `make lint` — passes (no new findings on `FlowsPage.tsx`; pre-existing warnings unchanged)
- [x] `make build` — succeeds (UI + Go)
- [x] `pnpm test` (webui Vitest) — 235/235 pass
- [ ] Manual: navigate Flows to page 5, then:
  - change protocol filter — expect return to page 1 with filtered set
  - type into URL pattern — expect immediate return to page 1, debounced fetch fires at offset 0
  - type into host / body / tag — same as URL pattern

Resolves USK-964
Linear: https://linear.app/usk6666/issue/USK-964

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/web/src/pages/Flows/FlowsPage.tsx
+++ b/web/src/pages/Flows/FlowsPage.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useEffect, useMemo, useRef, useState } from "react";
+import { useCallback, useEffect, useMemo, useState } from "react";
 import { useNavigate } from "react-router-dom";
 import { Badge } from "../../components/ui/Badge.js";
 import { Button } from "../../components/ui/Button.js";
@@ -228,18 +228,11 @@ export function FlowsPage() {
   const flows = data?.flows ?? [];
   const total = data?.total ?? 0;
 
-  // Reset offset and selection when debounced text filters change
-  const prevTextFilterKey = useRef("");
-  useEffect(() => {
-    const key = JSON.stringify({ debouncedUrlPattern, debouncedBodyContains, debouncedTagFilter, debouncedHostFilter });
-    if (prevTextFilterKey.current && prevTextFilterKey.current !== key) {
-      setOffset(0);
-      setSelectedIds(new Set());
-    }
-    prevTextFilterKey.current = key;
-  }, [debouncedUrlPattern, debouncedBodyContains, debouncedTagFilter, debouncedHostFilter]);
-
-  // Reset offset when filter changes (used by non-debounced filter controls)
+  // Reset offset and selection on any filter change. Called synchronously
+  // from every filter setter so the next useQuery render sees offset=0
+  // alongside the new filter — avoids the 1-render race where a debounced
+  // filter committed first and offset reset arrived one render later, so
+  // the query briefly fired with new filter + stale offset (USK-964).
   const handleFilterChange = useCallback(() => {
     setOffset(0);
     setSelectedIds(new Set());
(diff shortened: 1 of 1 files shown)


usk6666 (commented):
## Code Review: APPROVED

The fix correctly addresses USK-964 by collapsing the 1-render race in the four debounced text-input handlers. The root cause analysis in the PR body matches the diff: `useDebounce` -> `setTimeout` -> `setState` produces a state-commit ordering where the debounced filter value reaches `useQuery` one render before the post-render `useEffect` resets `offset`. Moving the reset into the `onChange` handlers via the already-existing `handleFilterChange()` ensures `offset=0` lands in the same render as the new filter state.

### What was verified

- **Filter-setter audit is complete**: all four text-input handlers (`handleHostFilterChange`, `handleUrlChange`, `handleBodyContainsChange`, `handleTagFilterChange`) now call `handleFilterChange()`. The six radio/select handlers (protocol, method, status, state, scheme, blocked-by) were already calling it pre-PR. Sort and page-size handlers reset offset inline as before. No filter setter remains that does not reset offset.
- **Dead-code removal is clean**: `useRef` import dropped from the React import line, `prevTextFilterKey` ref and its companion `useEffect` removed in one block. No dangling references; the file no longer imports or uses `useRef`.
- **`useCallback` dep arrays updated correctly**: each of the four text-input handlers now lists `[handleFilterChange]` in its deps array, matching how the radio/select handlers already declare it. `handleFilterChange` itself is stable (deps: `[]`), so handler identity is preserved across renders.
- **Comment accuracy**: the new comment on `handleFilterChange` correctly documents that synchronous invocation from every setter avoids the 1-render race, and the per-handler comments now read "debounced query, immediate offset reset" — consistent with the new behavior.

### Notes

- **Test coverage**: no new tests, which is consistent with the WebUI test-suite policy referenced in `web/src/pages/Resend/ResendPage.dispatch.test.tsx` (avoiding RTL/jsdom). The fix lives entirely inside React hooks/handlers, so extracting a testable unit would require refactoring out of bug-fix scope. The PR test plan documents manual verification steps, which is the right call for this change.
- **Minor nit (not a finding)**: every keystroke now calls `setSelectedIds(new Set())` unconditionally, allocating an empty Set even when no selection exists. In practice this is invisible (React batches with the `setHostFilter` update and the empty-Set referential identity change is cheap), so not worth fixing here.

FINDINGS: None

STATS:
  CRITICAL: 0
  HIGH: 0
  MEDIUM: 0
  LOW: 0
  NIT: 0

---

usk6666 (commented):
## Security Review: APPROVED

Pure client-side WebUI state change. Single file (`web/src/pages/Flows/FlowsPage.tsx`) modified; no changes to the data path, MCP server, transport, TLS, or any net/http code. No new dependencies added.

### What was checked

- **No new attack surface**: the fix moves the `setOffset(0)` / `setSelectedIds(new Set())` calls from a `useEffect` triggered by debounced state into the four `onChange` handlers. These are still pure React state mutations confined to the FlowsPage component.
- **No information leak via URL**: confirmed there is no `useSearchParams`/router-state sync in this page; filter and pagination state are local `useState` only, so leaking a stale offset (or a reset to 0) into the URL is not possible.
- **Filter values are not interpreted as code**: filter strings flow into the `useMemo` -> `QueryFilter` -> `useQuery("flows", ...)` chain as structured JSON params to the MCP `query` tool. They are not concatenated into shell, SQL, or HTML, and there are no `innerHTML` / `dangerouslySetInnerHTML` introductions. Backend MCP query handler is responsible for parameterization, which is unchanged here.
- **No CWE candidates triggered**: no auth/authz changes, no crypto, no input parsing/decoding additions, no resource lifetime changes. The removed `useRef` and `useEffect` do not hold any sensitive state.
- **License/dependency footprint**: no `package.json` change.

FINDINGS: None

STATS:
  CRITICAL: 0
  HIGH: 0
  MEDIUM: 0
  LOW: 0

---

PR #43, opened 2026-05-22
refactor(webui/Settings): single source of truth for intercept_queue config

usk6666:
## Summary
- Drop the duplicate Intercept Queue card from `ConnectionSettings.tsx` — InterceptRules is now the sole owner of the `intercept_queue` configure path.
- Replace the deleted card with a small pointer note so the ConnectionSettings tab still surfaces that the feature exists and where it moved.

## Why
Both tabs held independent `useState` for `timeout_ms` / `timeout_behavior`, both starting from empty defaults. Saving the (untouched) ConnectionSettings card would silently clobber any value set via InterceptRules. Single owner eliminates the race by construction.

## Test plan
- [ ] Open `Settings → Connection`: confirm no Queue Timeout input, only the pointer card.
- [ ] Open `Settings → Intercept Rules`: confirm the Intercept Queue card is unchanged and saving still works.
- [ ] Edit timeout in Intercept Rules → save → switch to Connection → switch back → no clobbering observed.
- [ ] `make build`, `make lint`, `make test` pass.

## Follow-up (not in this PR)
`ConfigResult` does not expose `intercept_queue` config, so the surviving InterceptRules card still loads with empty inputs (a pre-existing UX gap). File a follow-up Issue to add `intercept_queue: { timeout_ms, timeout_behavior }` to `ConfigResult` and prefill the form on mount.

Resolves USK-969
Linear: https://linear.app/usk6666/issue/USK-969

### DIFF
--- a/web/src/pages/Settings/ConnectionSettings.tsx
+++ b/web/src/pages/Settings/ConnectionSettings.tsx
@@ -17,7 +17,11 @@ interface ConnectionSettingsProps {
 const DEFAULT_LISTENER = "default";
 
 /**
- * ConnectionSettings — manage connection limits, timeouts, upstream proxy, and intercept queue.
+ * ConnectionSettings — manage connection limits, timeouts, and upstream proxy.
+ *
+ * Intercept queue configuration (timeout / behavior) lives on the Intercept
+ * Rules tab as the single source of truth (USK-969). A small pointer note is
+ * retained here for discoverability.
  */
 export function ConnectionSettings({ status, onRefresh }: ConnectionSettingsProps) {
   const { addToast } = useToast();
@@ -46,10 +50,6 @@ export function ConnectionSettings({ status, onRefresh }: ConnectionSettingsProp
     [status.listeners],
   );
   const [selectedListener, setSelectedListener] = useState<string>(DEFAULT_LISTENER);
-
-  // Intercept queue
-  const [queueTimeout, setQueueTimeout] = useState("");
-  const [queueBehavior, setQueueBehavior] = useState<"auto_release" | "auto_drop">("auto_release");
 
   // Sync state when status changes
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


usk6666 (commented):
## Code Review: APPROVED

Clean refactor that deletes the duplicate intercept_queue editor from `ConnectionSettings`, leaving `InterceptRules.tsx` as the single source of truth (confirmed: `handleSaveQueueSettings` at `InterceptRules.tsx:154` dispatches the same `intercept_queue` payload shape). JSDoc, imports, hooks, and JSX classes are all consistent and no dead code remains.

### Verification

- **Surviving owner exists**: `web/src/pages/Settings/InterceptRules.tsx` lines 37-39 (state), 154-181 (`handleSaveQueueSettings`), 298+ (card UI) — full queue editor lives there.
- **Tab discoverability**: pointer text says "Intercept Rules tab", which matches the `TABS` label `"Intercept Rules"` (id `"intercept"`) in `SettingsPage.tsx`. The two tabs sit in the same SettingsPage, so the cross-reference is reachable.
- **No orphaned symbols**: `useCallback` / `Input` / `Button` / `useState` / `useEffect` / `useMemo` / `useToast` all remain referenced by the Connection & Timeouts and Upstream Proxy cards.
- **Race-by-construction**: with only one editor binding `queueTimeoutMs` + `queueTimeoutBehavior` (under different names, but same payload key), there is no longer a path for a stale ConnectionSettings save to clobber InterceptRules.
- **Scope**: strictly deletion + pointer note. No drive-bys.

### NIT (non-blocking)

| F-1 | NIT | `web/src/pages/Settings/ConnectionSettings.tsx` | post-PR ~233-244 | Consistency | Pointer card retains a `settings-card-header` with only the "Intercept Queue" title and no Save button, which is visually inconsistent with the other two cards in this file (both right-align a Save). Pointer text is also passive — it states where the setting lives but does not link there. Optional polish: drop the header and render just a `<p className="settings-section-desc">` in a plain card body, or wire a clickable cross-tab link (would require lifting `setActiveTab` from `SettingsPage`). Neither is blocking — current rendering is functionally fine.

---

usk6666 (commented):
## Security Review: APPROVED

WebUI-only deletion that consolidates intercept_queue config onto the Intercept Rules tab. No validation, guard, or rate limit was lost — the surviving `handleSaveQueueSettings` in `InterceptRules.tsx` preserves the same non-negative numeric range check and Save-button disabled state. The added pointer note is a hard-coded static string with no interpolation; no `dangerouslySetInnerHTML`, no new deps, and no sensitive data exposed.

### Checks performed

- **Validation parity**: `InterceptRules.tsx:154-181` retains `isNaN(parsed) || parsed < 0` → "Timeout must be a non-negative number" warning (equivalent to the deleted code path, in fact slightly more robust since it operates on the parsed local rather than casting through `as number`).
- **XSS / unsafe rendering**: No `dangerouslySetInnerHTML`; the new pointer copy is a static literal in JSX text content.
- **Sensitive data exposure**: Note text references only the public Issue ID (USK-969) and a tab name — no paths, secrets, or internal identifiers leaked.
- **Dependencies**: Diff is confined to `web/src/pages/Settings/ConnectionSettings.tsx`; `web/package.json` untouched.
- **Removed guards**: No double-submit / race-on-save guard lived solely on the deleted handler; `disabled={configureLoading}` on the surviving Save button preserves the click-debounce pattern.

FINDINGS: None

STATS:
- CRITICAL: 0
- HIGH: 0
- MEDIUM: 0
- LOW: 0

---

PR #46, opened 2026-05-22
fix(webui/Resend): enforce json_path/regex mutex in BodyPatchEditor

usk6666:
## Summary
- Add a per-row `json_path` / `regex` mode `<select>` in `BodyPatchEditor`, mirroring the `RawPatchEditor` (USK-967) pattern.
- On mode switch, clear the other-mode fields to prevent silent contamination of the wire shape.
- Disable the Resend Send button while any body patch is empty or has both modes populated, via a new `onValidityChange` prop wired to `bodyPatchesValid` state in `ResendPage`.
- Export pure helpers `detectBodyPatchMode` and `isBodyPatchValid`; cover both with table-driven vitest cases in `BodyPatchEditor.test.ts` (node env, no DOM).

## Issue framing note
The Issue body describes the backend as "silently ignoring" one of the two modes when both fields are populated. The actual behavior is that the backend `validateBodyPatch` (`internal/mcp/body_patch.go`) returns a hard error. The defect this PR addresses is the UX gap that allowed the conflicting input + the needless server round-trip; the wire/data correctness was never at risk. The `BodyPatch` TS type and the Go struct are unchanged.

## Test plan
- [x] `cd web && npm run lint && npm run build && npm test` all pass (332 vitest cases, including the new `BodyPatchEditor.test.ts`)
- [x] `make build` passes (UI rebuild + `go vet` + `go build`)
- [ ] In the running app, BodyPatchEditor renders a mode `<select>` on every row; switching mode clears the other-mode fields.
- [ ] Adding an empty patch disables the Resend Send button; populating either mode re-enables it.
- [ ] Mode switching does not lose currently-active-mode field values.

Resolves USK-973
Linear: https://linear.app/usk6666/issue/USK-973/fixwebuiresendbodypatch-enforce-mutex-between-json-path-and-regex

### DIFF
--- a/web/src/pages/Resend/BodyPatchEditor.css
+++ b/web/src/pages/Resend/BodyPatchEditor.css
@@ -27,6 +27,34 @@
   border-radius: var(--radius-md);
 }
 
+.patch-editor-mode-row {
+  display: flex;
+  gap: var(--space-sm);
+  align-items: center;
+}
+
+.patch-editor-mode-select {
+  height: 28px;
+  padding: 0 var(--space-sm);
+  font-size: var(--font-size-xs);
+  color: var(--text-primary);
+  background-color: var(--bg-secondary);
+  border: 1px solid var(--border-muted);
+  border-radius: var(--radius-sm);
+  cursor: pointer;
+  appearance: none;
+  background-image: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' width='12' height='12' viewBox='0 0 12 12'%3E%3Cpath fill='%238b949e' d='M3 5l3 3 3-3'/%3E%3C/svg%3E");
+  background-repeat: no-repeat;
+  background-position: right 6px center;
+  padding-right: 20px;
+}
+
+.patch-editor-mode-select:focus {
+  outline: none;
+  border-color: var(--accent-primary);
+  box-shadow: 0 0 0 2px rgba(88, 166, 255, 0.2);
+}
+
 .patch-editor-row {
   display: flex;
   gap: var(--space-sm);
@@ -63,6 +91,7 @@
   width: 28px;
   height: 28px;
   flex-shrink: 0;
+  margin-left: auto;
   font-family: var(--font-mono);
   font-size: var(--font-size-sm);
   color: var(--text-muted);
@@ -82,4 +111,15 @@
 .patch-editor-spacer {
   width: 28px;
   flex-shrink: 0;
+}
+
+/* Inline validation error for empty or conflicting patches (USK-973). */
+.patch-editor-error {
+  font-size: var(--font-size-xs);
+  color: var(--accent-danger);
+  padding-left: var(--space-xs);
+}
+
+.patch-editor-field[aria-invalid="true"] {
+  border-color: var(--accent-danger);
 }
--- a/web/src/pages/Resend/BodyPatchEditor.test.ts
+++ b/web/src/pages/Resend/BodyPatchEditor.test.ts
@@ -1,0 +1,119 @@
+/**
+ * Tests for the BodyPatchEditor pure helpers (USK-973).
+ *
+ * The editor previously surfaced both `json_path + value` and
+ * `regex + replace` rows simultaneously, allowing the user to populate
+ * both modes on the same patch. The backend `validateBodyPatch`
+ * (`internal/mcp/body_patch.go`) rejects that shape with a hard error,
+ * so the UX gap caused a needless server round-trip for input that
+ * could never succeed. `detectBodyPatchMode` selects the rendering
+ * mode for an existing patch (defensive against legacy data) and
+ * `isBodyPatchValid` mirrors the backend's exactly-one-of contract so
+ * the parent's submit button can be disabled inline.
+ *
+ * The component itself is JSX-heavy; tests for it would pull in jsdom
+ * + RTL. Following the pattern of RawPatchEditor.test.ts, we verify
+ * the pure helpers here.
+ */
+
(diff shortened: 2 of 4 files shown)


usk6666 (commented):
## Code Review

**VERDICT**: APPROVED

**SUMMARY**: PR #46 enforces the backend `validateBodyPatch` exactly-one-of contract at the editor boundary by replacing the dual-row layout with a per-row mode `<select>` + inline error + parent submit-gate. The implementation paint-by-numbers mirrors `RawPatchEditor` (USK-967): same `onValidityChange` prop shape, same `mode-row` + per-mode rows structure, same pure-helper export pattern for table-driven vitest cases. Backend Go (`internal/mcp/body_patch.go`) and the `BodyPatch` TS type are unchanged — this is a pure UX gate.

**FINDINGS**:

- **LOW** [`web/src/pages/Resend/BodyPatchEditor.tsx:69-71`] The `useEffect` notification fires unconditionally on mount with `isValid = true` for the default empty `patches` array, producing one no-op `setBodyPatchesValid(true)` call per parent mount. Same pattern as the sibling `RawPatchEditor` and intentional; not actionable.
- **NIT** [`web/src/pages/Resend/BodyPatchEditor.css`] File ends without a trailing newline (diff shows `\ No newline at end of file`). Minor.
- **NIT** [`web/src/pages/Resend/BodyPatchEditor.tsx:33-36`] `detectBodyPatchMode` resolves `{ replace: "bar" }` (no `regex` field) as `regex` mode. The UI then renders empty regex inputs alongside the `replace` value plus the validation error. Defensive behaviour for legacy data is appropriate; the editor's own write path never produces this shape.

**STATS**:
- CRITICAL: 0
- HIGH: 0
- MEDIUM: 0
- LOW: 1
- NIT: 2

### Strengths

1. **Paint-by-numbers consistency with RawPatchEditor (USK-967)** — same prop surface, same effect shape, same eslint-disable lint exception for the pure-helper exports. Reduces review surface and future maintenance burden.
2. **Pure-helper extraction** — `detectBodyPatchMode` + `isBodyPatchValid` are exported with `react-refresh/only-export-components` disable comments and exercised by 18 vitest cases without pulling in jsdom/RTL.
3. **Strong typing** — single justified `as BodyPatchMode` cast on the `<select>` onChange; no `any`. `useCallback` / `useEffect` deps all correct.
4. **Accessibility** — `aria-invalid` on the relevant field, `role="alert"` + per-patch `aria-label` on the error message. CSS adds an `[aria-invalid="true"]` red border rule.
5. **Test coverage** — Happy + both error modes + whitespace-only edge cases + null-value-is-valid + empty-replace-is-valid (strip semantic). The "blank mode-switch shape" test pins the exact transition produced by `handleModeChange`.
6. **Wire shape is preserved** — `mode` is a UI-only render selector; the `BodyPatch` TS interface and the Go struct are untouched, so this PR cannot regress data correctness.
7. **Backend reality check is documented in the PR body** — Issue described "silently ignoring" but the backend actually hard-rejects; the PR is correctly scoped to closing the UX gap before the round-trip.

### Verified

- `BodyPatch` type in `web/src/lib/mcp/types.ts:916` unchanged.
- Backend `validateBodyPatch` in `internal/mcp/body_patch.go:30-58` unchanged — exactly-one-of contract matches the new client helper.
- Flow-load reset in `ResendPage.tsx:796` clears `bodyPatches` to `[]`, confirming `detectBodyPatchMode`'s legacy-defensive path is genuinely defense-only.
- Submit-gate wired correctly at `ResendPage.tsx:1551` (`disabled={httpSending || !bodyPatchesValid}`) with tooltip explanation.

Reviewed against `web/src/pages/Resend/RawPatchEditor.tsx` + `RawPatchEditor.test.ts` for sibling consistency. Patterns match.

usk6666 (commented):
## Security Review

**VERDICT**: APPROVED

**SUMMARY**: WebUI-only change. No new attack surface, no new dependencies. The regex field travels to the backend as an opaque string and is compiled by Go's RE2 engine (no catastrophic backtracking). The PR is a pre-submit UX gate over an existing backend validation contract — it cannot weaken any control.

**FINDINGS**: None.

### Checks performed

1. **XSS / DOM injection** — No `dangerouslySetInnerHTML`, no `innerHTML`, no `eval`, no `new Function` in any changed file. All user input flows into controlled `<input value>` / `<select value>` bindings rendered through React's text-escaping path.
2. **ReDoS** — The `regex` field is stored as a string and forwarded to the MCP backend; the editor never compiles it client-side. Backend uses `regexp.MustCompile` (RE2, no backtracking).
3. **State leaks** — No `console.*`, no `localStorage`, no `sessionStorage`, no `document.cookie`. Patch state lives in React `useState` and is dispatched only via `body_patches` on Send.
4. **MCP wire shape** — Confirmed: `BodyPatchMode` is render-only; the submitted JSON contains only the existing `BodyPatch` fields (`json_path`, `regex`, `replace`, `value`, `encoding`). No client-only fields cross the wire.
5. **Dependencies** — No `web/package.json` or `web/package-lock.json` changes; license-allow-list unaffected.

### Notes

- The PR replaces a state where `validateBodyPatch` was the sole guard against `{json_path, regex}` co-occurrence with a state where the client also enforces it. This is defense-in-depth; the backend remains authoritative.
- Backend (`internal/mcp/body_patch.go`) is unchanged, so no rule/safety/intercept path is affected.
- WebUI is local-only over MCP transport; this PR does not alter the transport binding.

Reviewed `BodyPatchEditor.tsx`, `BodyPatchEditor.test.ts`, `BodyPatchEditor.css`, and the `ResendPage.tsx` integration sections.

PR #49, opened 2026-05-22
perf(webui): pause polling on hidden tab and adapt to intercept queue state

usk6666:
## Summary
- Add a `pauseWhenHidden?: boolean` option to the `useQuery` hook (default `false`, backwards-compatible). When enabled, the polling timer pauses while `document.visibilityState === "hidden"` and fires one immediate refetch on transition back to `"visible"`.
- `MacrosPage` opts in to `pauseWhenHidden: true` so the 5s `macros` poll halts in background tabs.
- `InterceptPage` opts in to `pauseWhenHidden: true` AND switches its `pollInterval` from a fixed 1s to `queue.items.length > 0 ? 1s : 5s`, dropping idle-queue load to 12 RPM.
- Page Visibility API only (no new dependencies). SSR-safe (`typeof document !== "undefined"` guards). No behavior change for other `useQuery` consumers.

## Test plan
- [ ] `cd web && pnpm run lint && pnpm run build && pnpm test` all pass (including the three new cases in `hooks.test.ts`)
- [ ] `make build` and `make test` pass
- [ ] Manual: open Macros, switch tabs, watch Network panel — `query macros` calls cease. Switch back — one immediate fetch fires, then 5s cadence resumes.
- [ ] Manual: open Intercept with 0 items — Network panel shows 5s cadence. Add a held item via a request — cadence switches to 1s. Drop / forward the last item — back to 5s.

Resolves USK-974
Linear: https://linear.app/usk6666/issue/USK-974/perfwebui-adaptive-polling-for-macros-list-and-intercept-queue

### DIFF
--- a/web/src/lib/mcp/hooks.test.ts
+++ b/web/src/lib/mcp/hooks.test.ts
@@ -31,7 +31,7 @@
  * invariant — keeping the webui test surface dependency-free.
  */
 
-import { beforeEach, describe, expect, it, vi } from "vitest";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import type { McpClient } from "./client.js";
 
 // ---------------------------------------------------------------------------
@@ -229,6 +229,7 @@ const {
   useSecurity,
   useConfigure,
   useProxyControl,
+  useQuery,
 } = await import("./hooks.js");
 
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


usk6666 (commented):
## Code review (review-gate background, inline)

**Verdict: APPROVED (with 1 MEDIUM and 1 LOW)**

The PR cleanly adds an opt-in `pauseWhenHidden` to `useQuery` and wires it into `MacrosPage` + `InterceptPage` (the latter with an additional empty/non-empty queue cadence). Backwards compatibility is preserved: every other `useQuery` consumer (Dashboard, ProxyControl, Flows, Fuzz, FuzzResults) is untouched and `pauseWhenHidden` defaults to `false`. The polling-effect gating in `hooks.ts` is correct under real React (mount-visible → tab-hide → tab-visible → exactly one immediate refetch, then re-armed interval, ref cleared) and survives strict-mode double-invoke (the ref is set idempotently on the hidden branch; the visible branch consumes-and-clears the ref before arming). The new `wasPausedHiddenRef` is intentionally outside the `pauseWhenHidden` gating so the flag is never set when the option is off — confirmed by reading every branch of the polling effect.

`InterceptPage`'s `queueHasItems` boolean (rather than tracking raw length) is the right call: the `useQuery` polling effect re-arms on `pollInterval` change, so storing the raw count would tear down + rearm the timer on every poll that mutates the queue at fixed length. The mirror effect uses functional `setState` with the equality short-circuit `prev === next ? prev : next`, so identical writes do not produce a render. `INTERCEPT_POLL_ACTIVE_MS` / `INTERCEPT_POLL_IDLE_MS` are extracted as file-local consts with the comment justifying the 5s idle cadence (consistent with the Issue body).

`usePageVisible` is correctly file-private (no `export`), SSR-safe (`typeof document !== "undefined"` on both the `useState` initializer and the effect body), seeds once on mount via `handler()` to recover from a render that occurred while hidden, and removes the listener with the same function reference passed to `addEventListener`. No new npm deps.

`make test` (full Go suite) was not run inline (background review-gate context); the webui side runs clean. `cd web && pnpm test` → 317/317 pass.

---

### MEDIUM — `hooks.test.ts`: the transition-refetch test does not exercise the transition-refetch code path

`describe(\"useQuery — pauseWhenHidden + Page Visibility\") it(\"fires one immediate refetch on hidden → visible and re-arms the interval\")` (`hooks.test.ts:783-825`) **passes for the wrong reason**.

The in-house React-hook driver in this file (`driveHook` at `hooks.test.ts:259-271`) creates a fresh `Renderer` (with a fresh `slots: []`) **per `driveHook` invocation**:

```ts
function driveHook<T>(fn: () => T, count: number, between?: (i: number) => void): T[] {
  const renderer = createRenderer();   // <-- fresh slots, fresh refs, fresh state
  ...
}
```

Hook identity (including `wasPausedHiddenRef`) is preserved only across renders **inside a single `driveHook` call** (the `for (let i = 0; i < count; i++)` loop). The test makes **two separate `driveHook(..., 1)` calls** — these are two separate component mounts, not a re-render of the same instance. In the second call:

1. `useRef(false)` allocates a fresh `{current: false}` (slots are empty).
2. The polling-effect branch `pauseWhenHidden && wasPausedHiddenRef.current` evaluates to `false`, so the transition-refetch `fetchData()` is **not** invoked.
3. The querySpy call count still increases because the initial-fetch `useEffect` at `hooks.ts:358-361` fires on the new mount, and the assertion `callsAfterTransition > callsAfterInitial` passes.

Concretely: if you delete the entire transition-refetch block from `hooks.ts`,

```ts
if (pauseWhenHidden && wasPausedHiddenRef.current) {
  wasPausedHiddenRef.current = false;
  fetchData();
}
```

the test would still pass with the same call-count delta (initial-fetch effect contributes the increment). The regression guard for the immediate-refetch-on-transition behavior is therefore not actually in place.

**Fix options (caller's choice)**:

- **Recommended** — extend the driver to expose the renderer so `driveHook` can take an existing renderer (or expose `createRenderer` + the active-renderer setter), then drive both renders against the same renderer with `doc.fireVisibilityChange()` between them. This is a small change to the driver and gives a genuine same-instance re-render.
- **Or** — register an explicit setter into the React closure (e.g., via a `forceRender` ref) so the visibility-change handler triggers an in-renderer re-render. This more accurately models React's `setState` → re-render flow.
- **Or** — replace this single test case with a deeper assertion that the new mount with `pageVisible=true` arms the interval (which it does), and add an `expect(querySpy).toHaveBeenCalledTimes(1)` after initial / explicit comment that the transition-refetch path is covered by the manual-test plan only. This is the weakest option and I'd flag the loss of coverage.

The other two test cases (`does not arm the polling interval while the page is hidden` and `arms the polling interval normally when pauseWhenHidden is omitted (regression guard)`) **are** correct — they each use a single `driveHook` call and assert behaviour that doesn't depend on cross-render state preservation across separate `driveHook` calls.

---

### LOW — Initial fetch is not gated by `pauseWhenHidden`; consider documenting the WHY in the option's docstring

The docstring on `pauseWhenHidden` (`hooks.ts:187-193`) already says \"The initial fetch and dependency-driven refetches are not gated by visibility — only the periodic timer.\" That's accurate and intentional (the user opening a tab in background loads the page once). Could be slightly clearer about WHY (one-shot work is cheap; the cost is periodic polling). Optional polish; the current wording is fine.

---

### Other notes (non-findings)

- The polling effect dep list `[enabled, status, fetchData, pollInterval, pauseWhenHidden, pageVisible]` is exhaustive and correct.
- `pauseWhenHidden = options.pauseWhenHidden === true` explicit-coerce is good (treats `undefined` as `false`, matches docstring).
- No `dangerouslySetInnerHTML`, no `eval`, no new global mutation, no localStorage.
- Pre-existing lint errors in unrelated files (`Settings/*.tsx`, `Resend/TcpResponseViewer.tsx`) — not introduced by this PR.

usk6666 (commented):
## Security review (review-gate background, inline)

**Verdict: APPROVED (no findings)**

This PR is a WebUI-only polling-hook optimization with a tiny threat surface: no I/O, no auth, no credential flow, no new endpoints, no new dependencies. The Page Visibility API is read via `document.visibilityState` and `document.addEventListener(\"visibilitychange\", ...)` — both DOM-standard, browser-vended, no third-party. Reviewed against the WebUI-adapted criteria from the gate prompt:

1. **XSS / DOM** — N/A. No `dangerouslySetInnerHTML`, no `eval`, no innerHTML, no untrusted-string-to-DOM sink. The only DOM read is `document.visibilityState` (an enum); the only DOM write is `addEventListener` / `removeEventListener` with a fixed string event name and an internal function reference.
2. **Visibility-API listener lifecycle** — `usePageVisible` (`hooks.ts:219-240`) captures `handler` in the effect body, registers with `document.addEventListener(\"visibilitychange\", handler)`, and tears down with `document.removeEventListener(\"visibilitychange\", handler)` using the **same function reference**. On every effect re-run (only on mount/unmount here since deps are `[]`), the cleanup runs first, so listener leakage across remounts is impossible. The handler closure captures the `setVisible` state setter, which React-runtime guarantees to be stable, so no stale-closure / second-listener-on-same-element risk.
3. **Timer leaks** — The polling `useEffect` (`hooks.ts:375-395`) returns `clearInterval(timer)` whenever it arms a timer. When `pauseWhenHidden && !pageVisible`, the effect returns `undefined` (no timer was created, so no cleanup needed). When the effect re-runs because `pollInterval` / `pageVisible` / `fetchData` changes, React invokes the previous cleanup first — verified by reading the dep list `[enabled, status, fetchData, pollInterval, pauseWhenHidden, pageVisible]`. No stranded timers across remounts.
4. **State leaks / refs** — `usePageVisible` returns only a `boolean`; no internal refs escape. `wasPausedHiddenRef` is local to the `useQuery` body, never exported, never returned. The `InterceptPage` `queueHasItems` state is a plain boolean.
5. **Information disclosure** — No change to polling endpoints, no new request fields, no logging additions. Background-tab polling reduction actually *decreases* the rate at which the MCP server logs the query tool calls — incidental privacy/log-bloat improvement, not a regression.
6. **Dependencies** — `web/package.json` is unchanged in the PR diff; only `.ts` / `.tsx` files modified.
7. **CWE-833 / SSR globals** — `typeof document !== \"undefined\"` guards are present on both the `useState` initializer and the effect body. If `hooks.ts` is ever imported in a Node SSR context (or a vitest node test without a `document` stub), the helper short-circuits cleanly and returns `visible=true`.
8. **DoS / resource exhaustion** — On the contrary, this PR *reduces* idle resource usage in two ways: the background-tab pause stops polling entirely while hidden, and the Intercept queue idle cadence drops from 1 RPS to 0.2 RPS. No new unbounded loops, no new buffers.

### MITM compatibility triage

WebUI-only change; no data-path code (`internal/envelope/`, `internal/layer/`, `internal/connector/`, `internal/pipeline/`, `internal/pluginv2/`, `internal/proxybuild/`, `internal/rules/`, `internal/safety/`, `internal/flow/`, `internal/session/`) touched. No header / casing / dedup logic introduced. No `net/http` dependencies in data path. MITM-compat criteria N/A.

PR #50, opened 2026-05-22
feat(mcp/query): expose per-listener upstream_proxy_template + rotation_policy

usk6666:
## Summary
- Extend `queryListenerStatusEntry` (and the `internal/mcp/components.go` `ListenerStatus` mirror + converter) with `upstream_proxy_template` + `upstream_proxy_rotation_policy` so `query("status").listeners[*]` carries the rotation config alongside the literal URL.
- Apply `connector.RedactProxyURL` to the template at the publish boundary; policy passes through as an enum string.
- WebUI: add the two new fields (plus the previously-missing `upstream_proxy`) to `ListenerStatusEntry` in types.ts; prefill the rotation form in `ConnectionSettings.tsx` per-listener so a page reload restores the rotation editor's mode and values; remove the now-obsolete NOTE block from `UpstreamProxyEditor.tsx`.

## Test plan
- [x] `make lint` passes
- [x] `make build` passes (Go + WebUI)
- [x] `make test` passes (`internal/mcp` runs at 289s under parallel load; on a quiet runner the Makefile-side 4m timeout is unchanged from main)
- [x] New backend test in `internal/mcp/multi_listener_test.go` configures rotation on one listener, queries status, asserts redacted template + policy on that entry and omitempty on a sibling literal-URL listener.

Resolves USK-976
Linear: https://linear.app/usk6666/issue/USK-976

### DIFF
--- a/internal/mcp/components.go
+++ b/internal/mcp/components.go
@@ -282,12 +282,20 @@ type proxyManager interface {
 // UpstreamProxy (USK-826) carries the per-listener upstream-proxy URL.
 // query_tool redacts the URL before publishing it on the status resource;
 // the raw form is preserved here for in-process inspection.
+//
+// UpstreamProxyTemplate / UpstreamProxyRotationPolicy (USK-959 / USK-976)
+// carry the rotating-template form. Both fields are empty when the listener
+// is not using rotation. query_tool redacts UpstreamProxyTemplate via
+// connector.RedactProxyURL before publishing; the raw form is preserved
+// here for in-process inspection (matches the UpstreamProxy convention).
 type ListenerStatus struct {
-	Name              string `json:"name"`
-	ListenAddr        string `json:"listen_addr"`
-	ActiveConnections int    `json:"active_connections"`
-	UptimeSeconds     int64  `json:"uptime_seconds"`
-	UpstreamProxy     string `json:"upstream_proxy,omitempty"`
+	Name                        string `json:"name"`
+	ListenAddr                  string `json:"listen_addr"`
+	ActiveConnections           int    `json:"active_connections"`
+	UptimeSeconds               int64  `json:"uptime_seconds"`
+	UpstreamProxy               string `json:"upstream_proxy,omitempty"`
+	UpstreamProxyTemplate       string `json:"upstream_proxy_template,omitempty"`
+	UpstreamProxyRotationPolicy string `json:"upstream_proxy_rotation_policy,omitempty"`
 }
 
 // managerIsNil reports whether m is either a nil interface or wraps a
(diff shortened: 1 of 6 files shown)


usk6666 (commented):
## Code Review: APPROVED

Mechanical control-plane plumbing of `upstream_proxy_template` + `upstream_proxy_rotation_policy` through the `query("status")` surface, paralleled by a WebUI prefill path that restores the rotation editor state on reload. Field shapes match the documented USK-976 design (plain `string` with `omitempty` on the status surface, mirroring sibling `UpstreamProxy`), redaction is applied at the correct publish boundary, and the new backend test pins down both the positive (`secret` redacted, template macro preserved) and `omitempty` correctness paths for sibling listeners.

### Notes (no action required)

- `deriveUpstreamProxyValue` (ConnectionSettings.tsx) requires both `upstream_proxy_template` AND `upstream_proxy_rotation_policy` to be truthy before entering rotation mode. This is a sane belt-and-braces guard — partial state would be a backend bug, but graceful fallback to the literal-URL form is the right call.
- The redacted `xxxxx` placeholder flowing back into the rotation editor's `url_template` is a known UX (operator must re-enter credentials when editing). The `deriveUpstreamProxyValue` docstring documents this explicitly.

### MITM Triage

This PR is entirely in `internal/mcp/` + `web/src/` (control plane + WebUI). No data-path code touched, so no MITM-fidelity concerns apply.

---

usk6666 (commented):
## Security Review: APPROVED

Per-listener rotation template and policy are surfaced through the MCP `query("status")` resource. The template is redacted at the publish boundary via `connector.RedactProxyURL` (CWE-200 mitigation), with the raw form retained only in the manager for in-process inspection — matching the existing `UpstreamProxy` convention. The policy is an enum string with no credential content and passes through verbatim.

### Defense-in-depth observations

- The new test uses a template embedding `secret` as the password component and asserts that the **raw JSON** of the status response does NOT contain `secret` anywhere. This catches future regressions where redaction is bypassed or a new field surfaces unredacted.
- `RedactProxyURL("")` returns `""` (verified at `internal/connector/upstream_proxy.go:80-82`), so unconditional redaction on the non-rotation path is a no-op. Combined with `omitempty`, non-rotation listeners do not carry the rotation fields at all on the wire.
- The lexical-fallback path inside `RedactProxyURL` handles `§__nonce§`-style USK-959 templates that break `url.Parse`, ensuring credentials are scrubbed even when the URL is structurally malformed.

### Surface review

- No new attacker-controlled inputs are introduced.
- No new dependencies.
- No TLS / cert / network behavior changes.
- The WebUI receives only the redacted form; the rotation editor's `url_template` will visibly carry `xxxxx` as the password placeholder, which is the intended UX (operator re-enters creds on save).

---

PR #51, opened 2026-05-22
fix(webui/Settings): confirm dialog before TcpForwards Add triggers proxy_start

usk6666:
## Summary
- Add a confirm dialog to `TcpForwards.tsx`'s `handleAdd` path before it calls the `proxy_start` MCP tool. The dialog describes the actual side effect — `proxy_start` resets in-session settings on the targeted listener (intercept rules, transform rules, capture scope, TLS fingerprint, connection limits) and may interrupt in-flight connections on that listener. Cancel is a true no-op (returns before `start`); Apply proceeds with the existing call path.
- Reuses the existing `useDialog` infrastructure (`web/src/components/ui/Dialog.tsx`) — same pattern as `TlsFingerprint.tsx` / `GrpcSchemas.tsx`. No new dependencies, no new files.
- Tighten the disclaimer paragraph on the same page to reflect the real semantics.

## Issue framing note
The Issue body describes the side effect as "restart drops other listeners' state". Per Phase 1.5 design review, the actual semantics are **listener-scoped** (not cross-listener), and on already-running listeners the call returns `ErrAlreadyRunning` without any state change. The dialog text in this PR reflects the real semantics rather than the inaccurate framing.

## Out of scope (follow-up candidates)
- Per-listener restart MCP tool / `configure`-tool extension for `tcp_forwards` (would let the UI narrow the impact without a full `proxy_start`).
- Remove-forward UI button (does not exist today; only Add is supported in the UI).
- `ErrAlreadyRunning` UX improvement (requires backend change so add-while-running is meaningful).

## Test plan
- [x] `pnpm run build` passes
- [x] `pnpm test` passes (314/314)
- [ ] Manual: in Settings → TcpForwards, fill in a forward → click Add. Dialog appears with the described text. Cancel — no MCP call (Network panel). Apply — the forward is applied as today.

Resolves USK-975
Linear: https://linear.app/usk6666/issue/USK-975/fixwebuisettingstcpforwards-warn-on-proxy-restart-impact-before

### DIFF
--- a/web/src/pages/Settings/TcpForwards.tsx
+++ b/web/src/pages/Settings/TcpForwards.tsx
@@ -1,5 +1,5 @@
 import { useCallback, useState } from "react";
-import { Badge, Button, Input, useToast } from "../../components/ui/index.js";
+import { Badge, Button, Input, useDialog, useToast } from "../../components/ui/index.js";
 import { useProxyControl } from "../../lib/mcp/hooks.js";
 import type { ConfigResult, ForwardConfig, ForwardProtocol } from "../../lib/mcp/types.js";
 
@@ -21,6 +21,7 @@ const PROTOCOL_OPTIONS = ["auto", "raw", "http", "http2", "grpc", "websocket", "
  */
 export function TcpForwards({ config, onRefresh }: TcpForwardsProps) {
   const { addToast } = useToast();
+  const { showDialog } = useDialog();
   const { start, loading } = useProxyControl();
 
   // Form state for adding a new mapping
@@ -83,6 +84,18 @@ export function TcpForwards({ config, onRefresh }: TcpForwardsProps) {
     // Build new forwards map: existing + new entry
     const newForwards: Record<string, ForwardConfig> = { ...forwards, [port]: newEntry };
 
+    const confirmed = await showDialog({
+      title: "Apply TCP Forward",
+      message:
+        "Applying TCP forward changes calls proxy_start, which resets all " +
+        "in-session settings on the listener (intercept rules, transform rules, " +
+        "capture scope, TLS fingerprint, connection limits) and may interrupt " +
+        "in-flight connections on that listener. Continue?",
+      variant: "confirm",
+      confirmLabel: "Apply",
+    });
+    if (!confirmed) return;
+
     try {
       await start({
         tcp_forwards: newForwards,
@@ -105,7 +118,7 @@ export function TcpForwards({ config, onRefresh }: TcpForwardsProps) {
         message: `Failed to add TCP forward: ${err instanceof Error ? err.message : String(err)}`,
       });
     }
-  }, [localPort, upstreamHost, upstreamPort, protocol, tlsEnabled, upstreamTlsEnabled, forwards, start, addToast, onRefresh]);
+  }, [localPort, upstreamHost, upstreamPort, protocol, tlsEnabled, upstreamTlsEnabled, forwards, start, addToast, onRefresh, showDialog]);
 
   const handleKeyDown = useCallback(
     (e: React.KeyboardEvent) => {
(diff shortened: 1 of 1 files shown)


usk6666 (commented):
## Code Review

**VERDICT: APPROVED**

### Summary
UI-only diff adding a confirm dialog before `handleAdd` invokes `proxy_start` via `useProxyControl().start`. Reuses the existing `useDialog` infrastructure with the same shape as `TlsFingerprint.tsx`.

### Findings

#### Correctness
- `showDialog` is invoked AFTER all input validation and AFTER the `newForwards` snapshot is captured (line 84) and BEFORE `await start(...)` (line 99). `if (!confirmed) return;` correctly short-circuits the MCP call. Cancel is a true no-op as documented in the PR body.
- `useCallback` dependency array correctly extended with `showDialog`.

#### Sibling consistency
- Matches the `TlsFingerprint.tsx` confirm-flow pattern (`variant: "confirm"`, `confirmLabel: "Apply"`, `await showDialog(...); if (!confirmed) return;`).
- Deviation note (not a finding): `TlsFingerprint` gates the dialog on `status.running`, this one does not. That is the correct semantic here — TCP forwards are listener start-time config, so the confirm always applies regardless of the current proxy state.

#### TS/React conventions
- Clean import addition (`useDialog` added to the same import line). No unused imports. Types preserved.

#### Architecture / scope
- Reuses existing UI infrastructure. No new file, no new component, no new dependency.
- Disclaimer paragraph copy updated on the same page to reflect the actual side effect (listener-scoped reset of in-session settings). Consistent with the dialog text.

#### Test quality
- No new tests. Consistent with sibling confirm-flows (`TlsFingerprint`, `GrpcSchemas`) which also rely on `Dialog.test.tsx` covering the infrastructure. Acceptable.

#### Code health
- Dialog text accurately describes the real side effect (in-session settings reset on the listener, potential interruption of in-flight connections on that listener). The PR body explicitly documents the divergence from the Issue body's inaccurate "cross-listener" framing, and the dialog text matches the corrected semantics. No emojis, no decorative comments.

### Severity
No findings of any severity.

usk6666 (commented):
## Security Review

**VERDICT: APPROVED**

### Scope
WebUI-only change adding a confirm dialog before an existing MCP `proxy_start` call in `TcpForwards.tsx`. WebUI is not in the data path; threat surface for this diff is essentially zero.

### Findings

#### XSS / DOM injection
- Dialog `title` and `message` are plain string literals. No user-controlled content interpolated, no `dangerouslySetInnerHTML`. N/A.

#### CSRF / clickjacking
- The diff adds an extra user-confirmation gate before a destructive (in the sense of: resets in-session settings on the listener) MCP call. This strengthens, not weakens, the browser-trust boundary. No regression.

#### TOCTOU / race
- `newForwards` is computed at the top of `handleAdd` (line 84) BEFORE `await showDialog(...)` (line 87). The config that gets applied is the snapshot captured when the dialog was shown — even if the user mutates the form inputs in another tab/route during the dialog, the snapshot is invariant. The MCP call sends what the user implicitly reviewed.
- `useDialog` itself carries the USK-747 F-1 race rescue (documented in `Dialog.tsx` `applyShowDialog` / `markDialogSettled`); a racing second `showDialog` cannot pre-empt the user's actual click result for the first dialog.

#### State leakage / sensitive data
- No sensitive data exposed in the dialog (no secrets, no tokens, no internal IDs). N/A.

#### Dependencies / supply chain
- No new npm dependencies; no new transitive deps. N/A.

#### Authorisation / MCP boundary
- The dialog is purely client-side. The MCP server still authorises `proxy_start` on its own merits; adding a client-side confirm does not affect server-side authorisation.

### Severity
No findings of any severity.

PR #58, opened 2026-05-23
feat(mcp/fuzz_raw): per-iteration + per-job macro hooks with on_status reject

usk6666:
## Summary

USK-986 — sub-Issue 4/4 of USK-980. fuzz_raw is the **+1 adaptor** in
the N-1 uniform + 1 adaptor pattern. Raw is a byte stream with no L7
status concept, so this Issue mirrors the fuzz_http macro hook surface
(USK-960 / USK-961 / USK-981) but:

- **Validation-rejects** `run_interval="on_status"` with verbatim
  `fuzz_raw: post_macro run_interval="on_status" not supported (raw has no L7 status)` (and the matching pre_macro variant). Raw has no L7 status; use `on_match` against `__response_body` instead.
- Injects raw-specific `__response_body` (64 KiB cap) + `__response_chunks` + `__response_truncated`. Does **not** inject `__response_status` or `__response_headers*` (MITM principle: do not invent hypothetical surface).
- Restructures `runFuzzRawVariants` into a `fuzzRawVariantLoop` struct mirroring `fuzzHTTPVariantLoop`. No behaviour change to the existing fuzz_raw surface; the macro hooks layer on top.
- Forces `PassResponse=false` on the raw iteration executor so the HTTP-shaped `injectResponseVars` does not fire and overwrite the raw-specific keys.
- Adds the cross-protocol `__response_*` key matrix to `docs/rfc/envelope.md` §11 (HTTP and raw rows definitive; WS / gRPC rows marked TBD pending USK-984 / USK-985 acceptance).

9 integration tests including the adaptor-specific reject test all pass.

## Test plan

- [ ] `make lint` PASS
- [ ] `make build` PASS
- [ ] `make test` PASS — all 9 new tests + existing fuzz_raw suite
- [ ] `make test-e2e-smoke` PASS

Resolves USK-986
Linear: https://linear.app/usk6666/issue/USK-986

### DIFF
--- a/docs/rfc/envelope.md
+++ b/docs/rfc/envelope.md
@@ -1421,6 +1421,27 @@ This RFC is **accepted** as of 2026-04-12. Implementation proceeds on N1.
 **Post-N9 deferred design decisions:**
 - [ ] `WireLevelTap` interface unification — **deferred 2026-05-15 (USK-900)**. The five frame-level record callback sibling Options (`http2.WithFrameRecordCallback`, `http1.WithChunkRecordCallback`, `grpc.WithLPMFrameRecordCallback`, `httpaggregator.WithH2FrameRecordCallback`, `grpc.WithH2DataFrameRecordCallback`) already share their session-side closure builder (`session/h2_frame_record.go` `wireLevelRecordCallback()`), so the main boilerplate cost is already absorbed. Layer-side Option contracts remain per-Layer ad-hoc by design: `http1.WithChunkRecordCallback` is constrained to `func([]byte)` by the parser-level `ChunkRecordSetter` hook (parser owns chunk-boundary detection, not the Layer), so a naive `WireLevelTap` seam would degenerate into "four uniform + one adaptor". Re-evaluate when a 6th sibling appears (e.g. a WebSocket per-frame record producer) or when the http1 parser hook is revisited for an unrelated reason — at that point the seam will be either fully uniform or clearly fragmented, and the decision becomes unambiguous.
 
+### 11.1 Macro hook `__response_*` key matrix per protocol
+
+The `pre_macro` / `post_macro` hooks on the typed fuzz tools (`fuzz_http` / `fuzz_ws` / `fuzz_grpc` / `fuzz_raw`) inject a per-protocol set of reserved keys into the per-iteration KV Store so `post_macro` templates can reference the response. Each protocol's surface differs because the wire reality differs — raw has no L7 status, WS has framed message events instead of headers, gRPC has trailers in addition to headers, and HTTP has the canonical status / headers / body triplet. This is consistent with the MITM principle "do not invent hypothetical surface" (CLAUDE.md §MITM Implementation Principles #6).
+
+| Protocol | `__response_status` | `__response_body` | `__response_headers__<lower(name)>__` | `__response_chunks` | `__response_truncated` |
+| -------- | ------------------- | ----------------- | -------------------------------------- | ------------------- | ----------------------- |
+| http     | YES (3-digit status, base-10 decimal string) | YES (64 KiB cap) | YES (≤ 256 entries × 8 KiB) | n/a | n/a |
+| ws       | TBD pending USK-984 acceptance | TBD pending USK-984 | n/a (WS has no headers post-handshake) | TBD pending USK-984 | TBD pending USK-984 |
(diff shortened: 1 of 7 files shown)


usk6666:
## Review summary

**Security**: APPROVED. CWE-770 caps fire at capture (16 MiB receive cap in `runFuzzRawSingleExchange`) AND at injection (64 KiB kvStore cap in `injectRawResponseVars`). Verbatim error string contains no user-controlled bytes. Reserved-key shadowing protected by `InjectVarsRespectingReserved`. No shell/SQL injection surface.

**Code review**: CHANGES_REQUESTED on one MEDIUM finding (correctness — documented contract divergence). Adaptor reject is correctly implemented and verbatim-matches the Issue spec. CI is green across build/lint/test/test-e2e-smoke.

## Verified positives

- Reject error strings VERBATIM match the Issue spec at `internal/mcp/fuzz_raw_helpers.go:109-114`:
  - `fuzz_raw: post_macro run_interval="on_status" not supported (raw has no L7 status)` — exact match.
  - Mirrored pre-side string with `pre_macro` substituted (per implementer's report).
- Test 8 (`TestFuzzRawMacro_RunIntervalOnStatus_Rejected`) asserts the verbatim post substring AND that the call is rejected before any dial (no upstream connection, no fuzz job persistence). ✓
- `PassResponse=false` override on the raw hookExec correctly prevents the HTTP-shaped `injectResponseVars` from clobbering raw's `__response_*` keys. Confirmed by tracing `executePostMacro:311` (`if h.PassResponse { injectResponseVars(...) }`).
- `shouldRunPostMacro` still receives `respBytes` as `responseBody` argument independent of `PassResponse`, so `run_interval="on_match"` works correctly. Test 9 confirms positive + negative paths.
- `§11.1` matrix in `docs/rfc/envelope.md` is placed inside §11 (Acceptance Record), HTTP and raw rows definitive, WS/gRPC rows marked `TBD pending USK-984/985` with the right Issue IDs.
- 16 MiB receive cap fires at capture time (`fuzz_raw_helpers.go:582`), not at injection — answers the "receive-loop body buffering" question favorably.
- Help doc and JSON schema both document the adaptor reject prominently.
- 9 new tests pass per CI (`test-e2e-smoke` green).

## Findings

### M1 (MEDIUM, correctness): pre_macro extracts are not visible to post_macro

**Location**: `internal/mcp/fuzz_raw_helpers.go:577` and `:668`.

**Issue**: `runOne` creates a per-iteration `kvStore` at line 577 with the intent (per the inline comment) that "pre_macro extracts and the response __response_* keys (post path) land here". But `runPreMacro` (called at line 582 without the kvStore parameter) builds its OWN local `kvStore` at line 668 via a second `PrepareIteration` call and passes that to `executePreMacro`. The runPreMacro-local map is then discarded — pre's extracts never land in the runOne-local kvStore that `runPostMacro` consumes at line 615.

This means **any §var§ token in the post_macro's macro definition that references a pre_macro extract will fail to resolve**, even though:
- The schema description for `pre_macro` says: *"The hook macro shares the KV Store with the variant so extracts flow into §var§ template expansion."*
- The help doc says fuzz_raw follows the same shape as the fuzz_http siblings (where pre→post extract propagation works correctly via a single threaded kvStore — see `fuzz_http_helpers.go:579,593` which pass the same kvStore through `runOne→runPreMacro→runPostMacro`).
- The runOne comment on line 575-577 explicitly promises this.

None of the 9 new tests exercise the pre→post extract chain. Test 1 only counts pre call counts; tests 3/4/5 test on_error policies on a pre that always fails; tests 6/7 are scope=job (different code path).

**Suggested fix** (mechanical, ~3 lines): mirror the fuzz_http sibling by threading the runOne-local `kvStore` into `runPreMacro` as a parameter, and remove the inner `PrepareIteration` call in `runPreMacro`. Add a 10th test that records a pre extract and asserts a post_macro §var§ expansion sees it (e.g., pre extracts a token from a JSON response, post embeds that token in a URL query string, assert the post-server saw the expanded value).

### L1 (LOW, informational): `PassResponse=false` divergence justification

This is the divergence the PR description flags. The buildFuzzRawVariantLoop comment explains it well: HTTP-shape `injectResponseVars` would overwrite the raw-shape keys. No harmonization is needed; the divergence is load-bearing. Filing a follow-up Issue is **not required** in my view — the comment + the help doc + the §11.1 matrix together explain the choice.

## Verdict

- code_review: CHANGES_REQUESTED (M1)
- security_review: APPROVED
- final_verdict: CHANGES_REQUESTED (1 round to fix M1)

PR #59, opened 2026-05-23
feat(mcp/fuzz_grpc): per-iteration + per-job macro hooks

usk6666:
## Summary
- Refactored `fuzz_grpc` variant loop into a `fuzzGRPCVariantLoop` struct mirroring `fuzzHTTPVariantLoop` (no behaviour change to outputs).
- Extended `runFuzzGRPCSingleVariant` to expose the concatenated DATA payload (capped at 64 KiB at capture time) so `post_macro` receives a bounded `__response_body` without loading streaming responses into memory (CWE-770).
- Wired `pre_macro` / `post_macro` MCP tool inputs (`*MacroConfig`) to USK-983's shared seam: `BuildIterationHookExecutor` / `BuildJobHookExecutor` / `PrepareIteration`.
- Added `injectGRPCResponseVars` injecting gRPC-specific `__response_status` (0-16 domain), `__response_status_message`, `__response_body` (64 KiB cap), `__response_message_count`, `__response_total_bytes`. The HTTP-shaped injection inside `executePostMacro` becomes a no-op because nil body / nil headers are passed; the `statusCode` parameter passed feeds the `on_status` RunInterval gate with gRPC codes (0-16).
- `pre_macro` fires before the dial; `post_macro` fires after the end trailer is observed. `scope="job"` + mix-scope work end-to-end via the shared seam.
- Schema (`schema_fuzz_grpc.json`) + help-doc (`help_fuzz_grpc.md`) updated; help-doc explicitly documents the gRPC status domain (0-16 per `google.golang.org/grpc/codes`) and includes a canonical code→name table to prevent operators supplying HTTP codes to `run_interval=on_status`.
- 8 e2e integration tests in `fuzz_grpc_macro_integration_test.go` (`//go:build e2e && !e2e_smoke`):
  1. `TestFuzzGRPCMacro_PreMacroFiresBeforeDial`
  2. `TestFuzzGRPCMacro_PostMacroFiresAfterEndTrailer`
  3. `TestFuzzGRPCMacro_OnErrorSkip_SubsequentVariantsRun`
  4. `TestFuzzGRPCMacro_OnErrorAbort_StopsLoop`
  5. `TestFuzzGRPCMacro_OnErrorContinue_LiteralTokensInWire`
  6. `TestFuzzGRPCMacro_ScopeJobPre_FiresOnce`
  7. `TestFuzzGRPCMacro_ScopeJobPost_FiresOnce`
  8. `TestFuzzGRPCMacro_ResponseStatusInjected_GRPCDomain` (OK = 0 vs UNAVAILABLE = 14)

## Test plan
- [x] `make lint` PASS
- [x] `make build` PASS
- [x] `make test` PASS (unit tests)
- [x] `make test-e2e-smoke` PASS (merge gate)
- [x] All 8 new tests PASS under `-tags 'e2e'` (nightly tier)

Resolves USK-985
Linear: https://linear.app/usk6666/issue/USK-985

### DIFF
--- a/internal/mcp/fuzz_grpc.go
+++ b/internal/mcp/fuzz_grpc.go
@@ -96,6 +96,25 @@ type fuzzGRPCInput struct {
 
 	Positions   []fuzzGRPCPosition `json:"positions" jsonschema:"REQUIRED ordered position list; each describes a typed path into the GRPCStart/GRPCData shape and the payloads to substitute"`
 	StopOnNonOK bool               `json:"stop_on_non_ok,omitempty" jsonschema:"when true, abort the remaining variants once any variant returns a non-OK gRPC status (or terminates without a trailer)"`
+
+	// PreMacro / PostMacro: pre/post macro hooks (USK-985 — gRPC sibling
+	// of USK-960/USK-961/USK-981 on fuzz_http). scope="iteration" (default)
+	// shares a per-variant KV Store with the variant's pipeline so reserved
+	// keys §__iteration§ / §__nonce§, pre extracts, and (for post) reserved
+	// __response_* keys are visible to template expansion. scope="job"
+	// runs the hook exactly once outside the variant loop against a
+	// job-scoped KV Store that is merged into each iteration's store
+	// before per-iteration reserved keys are seeded — so pre=job extracts
+	// flow into every variant's request templating. mix-scope (pre and
+	// post with independent scope) is supported.
+	//
+	// gRPC-specific status domain: __response_status carries the gRPC
+	// status code (0 = OK, 1..16 = canonical gRPC error codes per
+	// google.golang.org/grpc/codes), NOT an HTTP status. RunInterval
+	// on_status matches against this domain — see help_fuzz_grpc.md for
+	// the canonical code-to-name mapping.
+	PreMacro  *MacroConfig `json:"pre_macro,omitempty" jsonschema:"pre macro hook executed before a variant's dial. scope=iteration (default; fires once per variant) | job (fires once before the variant loop; extracts shared with every variant). on_error: skip|abort|continue (default skip). For scope=job, skip returns a successful job result with completed_variants=0 and stopped_reason set"`
+	PostMacro *MacroConfig `json:"post_macro,omitempty" jsonschema:"post macro hook executed after a variant's end-trailer is received. scope=iteration (default; fires once per variant against gRPC-specific __response_* reserved keys: __response_status (gRPC 0-16 domain), __response_status_message, __response_body (capped 64KiB), __response_message_count, __response_total_bytes) | job (fires once after the variant loop completes; no __response_* keys, since there is no single response). on_error: skip|abort|continue (default skip)"`
 }
 
 // fuzzGRPCPosition describes one fuzz position. Path is a typed
@@ -218,7 +237,7 @@ func (s *Server) handleFuzzGRPC(ctx context.Context, _ *gomcp.CallToolRequest, i
 	// row existing. Mirrors USK-827 / fuzz_http precedent.
 	s.saveFuzzGRPCJob(context.Background(), fuzzID, &input, plan)
 
-	rows, completed, stopReason, runErr := s.runFuzzGRPCVariants(ctx, plan, timeout, input.StopOnNonOK, input.Tag, fuzzID)
+	rows, completed, stopReason, runErr := s.runFuzzGRPCVariants(ctx, plan, timeout, input.StopOnNonOK, input.Tag, fuzzID, input.PreMacro, input.PostMacro)
 
 	// Use a fresh background ctx so the closing UPDATE always lands,
 	// even on caller-side ctx cancel. The store-write is best-effort:
--- a/internal/mcp/fuzz_grpc_helpers.go
+++ b/internal/mcp/fuzz_grpc_helpers.go
@@ -67,6 +67,15 @@ const maxFuzzGRPCPositions = 32
 // service-name / payload fuzz cases and matches fuzz_http.
 const maxFuzzGRPCPayloadSize = 1 << 20
 
+// maxFuzzGRPCResponseBodyCapture caps the per-variant concatenated DATA
+// payload retained in-memory for the post_macro __response_body kvStore
+// key (USK-985). 64 KiB matches maxResponseBodyKVBytes (the HTTP fuzz
+// cap on __response_body) — the gRPC LPM Layer does not bound message
+// size on its own, so capping at capture time avoids loading multi-MiB
+// streaming responses into memory just to discard them at template
+// expansion (CWE-770).
+const maxFuzzGRPCResponseBodyCapture = 64 * 1024
+
 // validFuzzGRPCRoots lists the GRPCStart scalar field paths that
 // fuzz_grpc accepts. metadata[N].* and messages[N].payload paths are
 // matched separately via regex; this set covers the scalar fields.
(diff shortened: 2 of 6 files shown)


usk6666:
## Code Review (USK-985 — review-gate inline review)

### MEDIUM (correctness)

#### M1. `__response_body` reserved key for gRPC post_macro is silently clobbered to empty string

`runPostMacro` (`internal/mcp/fuzz_grpc_helpers.go:708-723`) correctly injects the captured gRPC body via `injectGRPCResponseVars`:

```go
injectGRPCResponseVars(kvStore, row.Status, row.StatusMessage, body, row.ResponseMessageCount, row.ResponseTotalBytes)
// ↑ writes kvStore["__response_body"] = string(body)
hookErr := l.hookExec.executePostMacro(ctx, int(row.Status), nil, nil, kvStore)
// ↑ executePostMacro → injectResponseVars(kvStore, statusCode, nil, nil)
//                       → kvStore["__response_body"] = string(nil) = ""
```

`hooks.go:383-390` unconditionally overwrites `macroResponseBodyKey` regardless of whether `responseBody` is nil:

```go
body := responseBody                // nil
if len(body) > maxResponseBodyKVBytes { ... }
kvStore[macroResponseBodyKey] = string(body)  // == ""
```

So after `executePostMacro` returns, `kvStore["__response_body"]` is always empty for gRPC. The help-doc (`help_fuzz_grpc.md:1546`) promises this key holds the concatenated DATA payload up to 64 KiB.

**Fix**: pass the captured `body` to `executePostMacro` instead of `nil`. The HTTP-shape injection will then write the same bytes that `injectGRPCResponseVars` already wrote (idempotent). Bonus: this fixes M2 too.

No test exercises `__response_body` value reaching the macro — the suite covers `__response_status`, `__response_status_message`, `__response_total_bytes`, `__response_message_count`, but not body content. Add a test for this.

#### M2. `run_interval="on_match"` for gRPC post_macro cannot fire

`runPostMacro` passes `nil` body to `executePostMacro`, which delegates to `shouldRunPostMacro(h, statusCode, nil)` (`hooks.go:461-490`). The `on_match` branch evaluates `h.compiledPattern.Match(nil)` — never matches gRPC content even when the body contains the pattern.

The help-doc (`help_fuzz_grpc.md`) and schema (`schema_fuzz_grpc.json:1677, 1722`) both document `match_pattern` as `"Matched against the concatenated DATA payload bytes (capped at 64 KiB)"`. The implementation cannot deliver on this contract.

**Fix**: same as M1 — pass the captured `body` to `executePostMacro`. Then `shouldRunPostMacro` sees the real bytes and the regex match works.

Add an integration test (e.g., `TestFuzzGRPCMacro_OnMatch_PostFiresOnBodyMatch`) to lock in the behavior.

### LOW

#### L1. Test gap: no assertion on `__response_body` value

Tests #2 (`PostMacroFiresAfterEndTrailer`) and #8 (`ResponseStatusInjected_GRPCDomain`) thread the post macro and inspect `__response_status` / `__response_message_count` / `__response_status_message` / `__response_total_bytes` via OverrideURL templating, but never include `__response_body` in the OverrideURL. Closing M1+M2 should add a test that asserts the body content reaches the macro (and that `on_match` fires on a body regex match).

#### L2. Cross-PR PassResponse divergence (per orchestrate prompt)

The orchestrate prompt flagged that USK-984/985 set `PassResponse=true` while USK-986 sets it `false`. The gRPC code chose `true` (via shared `BuildIterationHookExecutor`) and then nil-out body+headers at the call site as a workaround for the HTTP-shape injection. This works once M1+M2 are fixed (because then we pass the real body, not nil). Suggest a follow-up Issue to harmonize `PassResponse` semantics across the four fuzz tools — either always-true with protocol-specific projections, or always-false with explicit per-protocol injectors. Not blocking this PR.

### APPROVED with required fixes for M1 and M2

The structural work (variant loop refactor, scope=job seam wiring, gRPC status-domain documentation, 64 KiB body capture cap with bounded buffer pre-allocation, deterministic concat order) is solid. The eight new e2e tests cover scope/on-error/job-pre/job-post/status-domain paths well. The two bugs above only affect two specific contract points — `__response_body` content and `on_match` matching — and the fix is a one-line change at one call site.

---

## Security Review

### Scope: MCP control plane

- **MCP input validation**: `validateFuzzGRPCInput` calls the shared `ValidateMacroConfig`. OK.
- **Integer overflow in `N` field**: shared validator handles this. OK.
- **Race conditions**: variant loop is single-goroutine per fuzz job; rate-limit / store writes follow established patterns. OK.
- **Resource exhaustion (`__response_body` 64 KiB cap)**: `captureGRPCResponseBody` pre-allocates a 64 KiB buffer with `make([]byte, 0, cap)` and breaks out of the loop when `remaining <= 0`. Per-LPM payloads are independently capped by `grpc.Channel.maxMessageSize`. The receive slice `recvData` is bounded by stream lifetime — same as pre-PR behavior, no new exhaustion risk introduced. OK.
- **gRPC status code parsing**: `row.Status` comes from the gRPC end-trailer parse path, not from operator input. uint32 cast to int for shouldRunPostMacro is safe (status is 0-16 nominally; 32-bit range exceeds int on all supported platforms). OK.
- **Macro variable injection / template expansion safety**: reserved-key prefix dropping is handled by the shared seam. OK.

### Specific PR concerns from the orchestrate prompt

- **`runFuzzGRPCSingleVariant` signature change + body-capture path memory leak**: NO LEAK. Buffer is bounded at allocation time; iteration breaks at the cap. The full `recvData` lives in memory regardless of capture (per-LPM cap applies upstream). The cap-at-capture vs cap-at-injection distinction in the PR description is accurate for the buffer copy; whole-stream memory pressure is already bounded by the grpc Layer's `maxMessageSize`.
- **DATA concat determinism**: VERIFIED — `recvData` is in receive (wire) order from `receiveResendGRPCResponses`; `captureGRPCResponseBody` iterates without reordering. Deterministic per-stream.
- **`fuzzGRPCInput` did NOT gain UpstreamProxy**: VERIFIED — only PreMacro/PostMacro added.
- **gRPC status domain documentation**: VERIFIED — `help_fuzz_grpc.md` explicitly documents the 0-16 domain with a canonical code→name table (lines 1550-1556) and includes a worked example (lines 1607-1627) showing UNAVAILABLE=14, contrasted with HTTP 503.

### No security findings.

### Verdict: SECURITY APPROVED

usk6666:
## Review-gate fix applied (commit `960b7215`)

Both MEDIUM findings (M1: `__response_body` clobber, M2: `on_match` body-regex broken) are addressed by a single change at `runPostMacro` — pass the captured `body` to `executePostMacro` instead of `nil`. The HTTP-shape `injectResponseVars` then writes the same bytes that `injectGRPCResponseVars` already wrote (idempotent on `__response_status` and `__response_body`), and `shouldRunPostMacro` sees the real bytes for `on_match`.

Added two regression tests:

1. `TestFuzzGRPCMacro_ResponseBodyInjected` — templates `§__response_body§` into `override_body`, asserts the gRPC echo response substring lands at the post server (and is NOT the regression-signature empty body).
2. `TestFuzzGRPCMacro_OnMatchPostFiresOnBodyMatch` — drives two variants; only the one whose echo response matches the regex triggers a post call.

All 10 fuzz_grpc macro e2e tests pass (8 original + 2 new). `make build` / `make lint` / `make test` / `make test-e2e-smoke` all green.

L2 (cross-PR `PassResponse` divergence) remains as a follow-up suggestion — not blocking this PR.

### Final verdict: APPROVED

PR #60, opened 2026-05-23
feat(mcp/fuzz_http): wire updateState so on_error pre-macro fires after real errors

usk6666:
## Summary
- Wire `hookExecutor.updateState(statusCode, runErr != nil)` at the fuzz_http variant loop's normal-path post-iteration bump (replaces `BumpHookIterationCount` there). Skip + template-expand-error call sites keep the counter-only bump.
- Change `shouldRunPreMacro`'s `on_error` iter-0 from `return true` (vacuous fire) to `return false` — fire only after a real error.
- Add `TestFuzzHTTPMacro_RunIntervalOnError_FiresAfterError` and `TestFuzzHTTPMacro_RunIntervalOnError_DoesNotFireWithoutError`.
- Update `help_fuzz_http.md` `on_error` semantics. Update `TestShouldRunPreMacro_OnError` unit test to match the new iter-0 = no-fire contract.

## Design notes
Per Phase 1.5 design review:
- Shared seam left untouched (Option C) — fuzz_ws/grpc/raw have different status semantics; forcing a single shared signature would hit the "hardcoded-flag anti-pattern". Follow-up Issues will wire each protocol's on_error with protocol-appropriate "error" definitions.
- `shouldRunPostMacro` has no `on_error` case (only `always` / `on_status` / `on_match`), so no symmetry fix was needed there.

## Test plan
- [x] `make test` passes (all existing fuzz_http macro tests still pass — D-Q6 changes only the on_error branch)
- [x] New `TestFuzzHTTPMacro_RunIntervalOnError_FiresAfterError` shows pre fires exactly once (iter 2, after iter 1's 500)
- [x] New `TestFuzzHTTPMacro_RunIntervalOnError_DoesNotFireWithoutError` shows pre fires zero times when all responses are 200

Resolves USK-982
Linear: https://linear.app/usk6666/issue/USK-982

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #62, opened 2026-05-23
feat(mcp/fuzz_ws): wire updateState so on_error pre-macro fires after transport errors

usk6666:
## Summary

- USK-987: ports the USK-982 / PR #60 pattern to fuzz_ws. The variant loop's wire-completion-point now calls `hookExecutor.updateState(0, runErr != nil)` instead of `BumpHookIterationCount`, so `pre_macro.run_interval="on_error"` actually fires after WebSocket dial / upgrade / receive failures.
- statusCode is hardcoded to `0` (not `int(row.CloseCode)`) per design Q1=(a) / Q3: WS close frame codes (RFC 6455 §7.4) are L7 graceful-shutdown signals, not error signals. Passing the close code would spuriously trigger `on_error` on Close=1000 normal closure (1000 >= 400).
- Pre-wire abort branches (prep.skipped, template-expansion error) keep the counter-only `BumpHookIterationCount` since they never reached the wire and the previous wire-completed iteration's `(lastStatusCode, lastError)` should carry over.
- Close-code-driven `on_error` (Issue-body options b/c/d) explicitly deferred per YAGNI / MITM Principle #6 — file a follow-up when a concrete operator scenario emerges.

Parent: USK-982 (PR #60, commit `8bd9b0e4`) — same one-line behavior change at the WebSocket variant loop's wire-completion point. Sibling USK-989 (raw) lands the same change in `fuzz_raw_helpers.go` in parallel.

## Test plan

- [x] `go test -race -tags e2e -run TestFuzzWSMacro_RunIntervalOnError ./internal/mcp/` — both new tests pass
  - `TestFuzzWSMacro_RunIntervalOnError_FiresAfterError`: 3 variants × immediately-closed listener → preCalls == 2 (iter 0 doesn't fire because requestCount==0; iter 1+2 fire after the prior transport errors)
  - `TestFuzzWSMacro_RunIntervalOnError_DoesNotFireWithoutError`: 3 variants × echo server → preCalls == 0 (close-frame codes do NOT trigger on_error)
- [x] `go test -race -tags e2e -run TestFuzzWSMacro ./internal/mcp/` — all 10 macro tests pass, including the 8 pre-existing ones (no regressions)
- [x] `go test -race -tags e2e -run TestFuzzWS ./internal/mcp/` — full fuzz_ws e2e suite green
- [x] `make build` + `make lint` + `make test` (fast tier) — pass

Refs: USK-987


usk6666 (commented):
## Code Review: APPROVED

PR #62 cleanly ports the USK-982 / PR #60 `updateState` wiring pattern to fuzz_ws with the principled `statusCode=0` adjustment for WS semantics. The diff is a one-line behavior change at the variant loop's wire-completion point plus two well-targeted tests and a documentation bullet — exactly matching the Issue scope.

Verified:
- `updateState(0, runErr != nil)` at the wire-completion path correctly bumps `requestCount` AND records `lastError` so `shouldRunPreMacro`'s `on_error` gate (`hooks.go:506`) sees the right state on the next iteration.
- `statusCode=0` hardcoding is the right call: passing `int(row.CloseCode)` would make a normal closure (Close=1000) trip the `on_error` gate via `lastStatusCode >= 400`. The Q1=(a) / Q3 design decision honored.
- Pre-wire abort branches (`prep.skipped`, template-expansion error) correctly keep counter-only `BumpHookIterationCount` per the shared helper's documented contract — `lastStatusCode`/`lastError` carry over from the previous wire-completed iteration.
- Tests cover both directions: `FiresAfterError` proves transport errors fire pre on iter 1+2 (not 0), `DoesNotFireWithoutError` proves Close=1000 does NOT trigger on_error against an echo server. Sanity guards present.
- Comment blocks preserve cross-references to USK-981 / USK-982 / USK-987 / USK-989 — design context survives in-tree per `feedback_sub_issue_design_settled_at_parent.md`.
- gofmt-clean, no new exported API, mirrors USK-982 shape exactly.

No findings.

---

usk6666 (commented):
## Security Review: APPROVED

PR #62 is a control-plane (`internal/mcp/`) wiring change with no data-path impact and no new attacker-controlled input surface.

Threat-model fit:
- **TLS/Cert**: N/A — no TLS code touched.
- **Network**: N/A — no listener/dial code touched. Test uses loopback ephemeral-port listener with proper cleanup.
- **InputValidation**: `updateState` receives a hardcoded `0` and a boolean derived from `runErr` (transport observation). No operator-supplied data flows in.
- **GoSecurity**: `updateState` runs serially in the single-goroutine `runVariantLoop` — no race surface introduced. No goroutine leak (test's accept-and-close goroutine exits on listener close).
- **MCP/API**: No new MCP tool parameters. Hook-state poisoning vector considered: WS upstream close codes are deliberately ignored (`statusCode=0` hardcoded) — no spoof surface. `runErr` is driven by proxy-internal transport observation, not attacker-controllable directly.
- **Dependencies**: No new deps.

The `statusCode=0` decision actively **prevents** a CWE-840 (Business Logic) class of bug where Close=1000 normal-closure would be mis-classified as an error. Net security-positive change.

No findings.

---

PR #69, opened 2026-05-25
feat(connector): add parseClientHelloALPN extractor for sniff-first MITM

usk6666:
## Summary
- Add `parseClientHelloALPN(buf []byte) ([]string, error)` to `internal/connector/tls_clienthello.go`, mirroring the fail-soft semantics of `parseClientHelloSNI`.
- Refactor: extract shared private helper `parseClientHelloExtensions(buf) ([]byte, error)` that combines `unwrapClientHelloBody` + `skipToExtensions`. Both SNI and ALPN entry points now sit on top of it.
- Add `findALPNInExtensions(exts) ([]string, error)` to decode RFC 7301 §3.1 `ProtocolNameList`.
- Add constant `tlsExtensionALPN = 0x0010`.
- Add table-driven tests covering all 6 cases from the Issue body: `[h2, http/1.1]`, `[http/1.1]`, `[h2]`, ext absent, truncation, cross-extension SNI+ALPN independence.

No callers; behavioral no-op. Prelude for USK-997 wire-up.

## Test plan
- [x] `make lint` clean
- [x] `make test` passes (race-enabled)
- [x] New tests for `parseClientHelloALPN` cover the 6 cases enumerated in the Issue body
- [x] `parseClientHelloSNI` tests still pass after the refactor (no regression in shared helper extraction)

Resolves USK-996
Linear: https://linear.app/usk6666/issue/USK-996

### DIFF
--- a/internal/connector/tls_clienthello.go
+++ b/internal/connector/tls_clienthello.go
@@ -33,6 +33,10 @@ const tlsExtensionServerName = 0x0000
 // server_name extension (RFC 6066 §3).
 const tlsServerNameTypeHostname = 0x00
 
+// tlsExtensionALPN is the ExtensionType value for the Application-Layer
+// Protocol Negotiation (ALPN) extension (RFC 7301 §3.1, IANA-assigned 16).
+const tlsExtensionALPN = 0x0010
+
 // parseClientHelloSNI extracts the SNI host_name value from a TLS
 // ClientHello carried at the start of buf. Returns an empty string with
 // nil error when the ClientHello does not include the SNI extension or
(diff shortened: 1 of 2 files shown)


usk6666 (commented):
## Code Review: APPROVED

Clean prelude PR. The `parseClientHelloExtensions` refactor pulls the framing prelude into one place and both SNI + ALPN entry points sit on top of it without duplication. `parseClientHelloALPN` mirrors the `parseClientHelloSNI` fail-soft contract exactly. The new `findALPNInExtensions` / `parseALPNExtension` decode RFC 7301 ProtocolNameList correctly with length-prefix bounds checks at every step. Tests cover all six cases enumerated in the Issue body (three offer shapes, no-ext, empty list, truncated, SNI+ALPN independent, non-handshake). godoc comments are thorough and explain the fail-soft rationale.

FINDINGS: None

STATS:
  CRITICAL: 0
  HIGH: 0
  MEDIUM: 0
  LOW: 0
  NIT: 0

---

usk6666 (commented):
## Security Review: APPROVED

Pure byte parser on attacker-controlled ClientHello bytes. Every length-prefix dereference (`uint16` extDataLen, `uint16` listLen, `uint8` nameLen) is preceded by an explicit `len(...) < N` bounds check; no integer overflow risk since the widest field is uint16 which fits in `int` trivially. Both loops make guaranteed forward progress (consume at least one header byte per iteration even when the payload length is zero), so no infinite loop on malformed input. No panic surface — all slice operations are bounded by previously-checked lengths and no nil derefs are reachable.

Fail-soft semantics correctly distinguish the four security-relevant cases:
- ALPN extension absent → `(nil, nil)`.
- Empty ProtocolNameList (zero entries) → `(nil, nil)` — explicitly intended per RFC 7301 §3.1.
- Truncation at any length-prefix boundary → `errClientHelloIncomplete` (not silently `(nil, nil)`).
- Non-ClientHello / non-handshake → `errNotClientHello` propagated from the shared helper.

This means malformed ALPN cannot masquerade as "no ALPN" to the USK-997 wire-up logic — the error path is preserved end-to-end. The RFC-7301-violating `nameLen == 0` skip is documented as a deliberate MITM passthrough-faithful choice and does not change the error surface.

No new dependencies, no CA-key or TLS-cert handling, no network I/O, no concurrency, no `unsafe`.

FINDINGS: None

STATS:
  CRITICAL: 0
  HIGH: 0
  MEDIUM: 0
  LOW: 0

---

PR #71, opened 2026-05-25
fix(layer/http1,connector,proxybuild): detect stale h1 upstream conn and redial before send (USK-998 Phase 1)

usk6666:
## Summary

Brings HTTP/1.x to functional parity with HTTP/2 for stale-upstream-connection recovery, fixing a live browser-parity bug observed against `try.nextcloud.com` / `demo1.nextcloud.com` after browser idle.

- `http1.Layer.HealthCheck()` — non-blocking 1ms-deadline 1-byte peek; treats timeouts as alive and EOF/RST/idle-byte as stale.
- `http1.Layer.EnvelopeContextTemplate()` — value-copy accessor mirroring `http2.Layer`.
- `connector.RedialUpstreamH1` — mirrors `h2_redial.go`; offers ALPN `http/1.1` only; constructs the fresh Layer with the exact `buildStackFromRoute` option list; reuses `redialConnIDLabel` for wire-log continuity.
- `proxybuild.h1Chain` — thinner than h2's `redialChain` (no pool, no GOAWAY observer, no pre-warm).
- `runHTTP1Exchange` dial closure rewired to `chain.EnsureFresh(ctx)`.
- 13 unit tests + 1 e2e integration test (`//go:build e2e && !e2e_smoke`, exhaustive tier).

## Live bug reproduced

Chrome → `https://try.nextcloud.com/access` → idle ~4 min → re-navigate yields `write: broken pipe` in 14ms because `spawnLoopReceive` (`internal/layer/http1/layer.go:516`) parks on `pendingNotify` with `pendingQ` empty and never observes the upstream FIN after `Keep-Alive: timeout=5`. The next `sendRequestOpaque` zero-copy `Write` (`internal/layer/http1/channel.go:1034-1036`) hits EPIPE on the dead conn, recorded as `state=error` with no retry.

## Scope: Phase 1 only

Per the Issue body's rollout plan, Phase 1 (案 C: proactive healthcheck + redial) absorbs ~95% of the symptom with localized changes and minimal blast radius on the wire path. Phase 2 (案 A: `StaleUpstreamError` + replay-safe retry for the healthcheck-miss race) and Phase 3 (metrics) are explicitly out of scope and will be filed as separate follow-up Issues.

## MITM principle adherence

- Wire fidelity: HealthCheck may drop one idle byte from a misbehaving upstream — documented in its godoc. The byte cannot be attributed to any HTTP transaction (RFC 9112 §9.6 forbids data on idle keep-alive conns), and the conn is being terminated either way.
- ALPN: `http/1.1`-only offer on redial avoids inadvertent Layer-type mismatch via h2 upgrade.
- No new `net/http` imports in data-path packages.
- No header dedup / Host=URL enforcement / casing normalization introduced.

## Test plan

- [x] `make lint` passes (gofmt + vet + staticcheck + ineffassign + gocyclo)
- [x] `make test` passes (48 packages, includes the 13 new unit tests under `-race`)
- [ ] `make test-e2e` passes (includes `TestHTTP1Forward_StaleConnRecovery_AfterIdleClose`) — verified to compile; run nightly
- [ ] Manual: confirm h2 redial machinery untouched (no regression in USK-991/992/993 territory)

Resolves USK-998 (Phase 1)
Linear: https://linear.app/usk6666/issue/USK-998

### DIFF
--- a/internal/connector/h1_redial.go
+++ b/internal/connector/h1_redial.go
@@ -1,0 +1,102 @@
+package connector
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/usk6666/yorishiro-proxy/internal/envelope"
+	"github.com/usk6666/yorishiro-proxy/internal/layer/http1"
+)
+
+// RedialUpstreamH1 establishes a fresh upstream HTTP/1.x *Layer for target,
+// reusing the EnvelopeContext template carried by stale (so the new Layer
+// stamps envelopes with the same ConnID / TargetHost / TLS reality the
+// caller already recorded). The TLS knobs (per-host overlay, fingerprint,
+// upstream proxy) are resolved through cfg the same way the original
+// CONNECT path did — runtime-mutable overrides therefore reach the redial.
+//
+// generation identifies which step of the redial chain this dial is for
+// (USK-998 — mirroring the USK-991 h2 redial scheme): generation==1 is the
+// first redial (label suffix `/upstream-redial`); generation>=2 appends
+// `-N`. Callers pass len(chain)+1 (where chain is the slice of prior fresh
+// Layers, including the original).
+//
+// USK-998: invoked by proxybuild's h1 chain inside the per-exchange dial
+// closure when [http1.Layer.HealthCheck] reports the current Layer is
+// stale (peer FIN / RST during idle keep-alive, or RFC 9112 violation
+// during idle). HTTP/1 has no connection pool — the per-CONNECT
+// lifetime owns its single upstream Layer (and any redialed successors).
+//
+// ALPN: offers only "http/1.1". By definition the prior negotiation
+// produced http/1.1 (we were holding a *http1.Layer); re-negotiating to
+// h2 would invalidate the Layer type already wired into the client side
+// of the stack.
+//
+// Returned errors are upstream dial / TLS errors verbatim so the caller
+// can wrap them via fmt.Errorf("dial: %w", ...) the same way the regular
+// dial closure does, keeping ClassifyError taxonomy intact.
+//
+// Wire fidelity: the fresh Layer is constructed identically to the
+// original buildStackFromRoute "http1" upstream path (same options, same
+// EnvelopeContext template), so envelopes flowing through it round-trip
+// with the same wire shape.
+func RedialUpstreamH1(
+	ctx context.Context,
+	target string,
+	stale *http1.Layer,
+	cfg *BuildConfig,
+	generation int,
+) (*http1.Layer, error) {
+	if cfg == nil {
+		return nil, fmt.Errorf("connector: RedialUpstreamH1: cfg is nil")
+	}
+	if stale == nil {
+		return nil, fmt.Errorf("connector: RedialUpstreamH1: stale is nil")
+	}
+
+	host := extractHost(target)
+	if host == "" {
+		return nil, fmt.Errorf("connector: RedialUpstreamH1: invalid target %q", target)
+	}
+
+	hostTLS, err := resolvePerHostTLS(target, cfg)
+	if err != nil {
+		return nil, err
+	}
+
+	// Offer only http/1.1: the prior negotiation produced http/1.1 and the
+	// stack's client side is wired for an *http1.Layer. Re-negotiating to
+	// h2 would force a Layer type swap upstream-side that the per-CONNECT
+	// stack is not prepared for.
+	upstreamConn, _, err := dialUpstreamWithALPN(ctx, target, host,
+		[]string{ALPNProtocolHTTP11},
+		hostTLS.insecureSkip, hostTLS.clientCert, hostTLS.rootCAs, cfg,
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	// Reuse the original Layer's EnvelopeContext template (mirroring h2:
+	// the older TLS snapshot is kept on the template; Pipeline / RecordStep
+	// already snapshotted the original handshake on prior envelopes, so
+	// keeping the template coherent for the CONNECT's lifetime is more
+	// useful than re-recording TLS here).
+	envCtx := stale.EnvelopeContextTemplate()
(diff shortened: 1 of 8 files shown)


usk6666 (commented):
## Code Review: APPROVED

The PR implements Phase 1 (案 C — proactive healthcheck + redial) of USK-998 cleanly. The change brings HTTP/1.x to functional parity with HTTP/2's stale-upstream recovery, with appropriate scope discipline: no Phase 2 (StaleUpstreamError + replay retry) or Phase 3 (metrics) leakage. The h1Chain is correctly thinner than h2's redialChain (no pool / no GOAWAY observer / no pre-warm), reflecting HTTP/1's lack of multiplexing. HealthCheck's pendingMu+1ms-deadline design is sound and avoids racing the parser goroutine in the documented (pendingQ empty) invariant. Wire-fidelity trade-off (one idle byte dropped on RFC 9112 §9.6 violation) is explicitly documented in godoc.

Build/test verified locally: `go build ./...`, `go test -race ./internal/{layer/http1,connector,proxybuild}/` and the e2e integration test `TestHTTP1StaleConnRecovery_AfterIdleClose` all pass under `-tags e2e`. gofmt / vet / staticcheck / ineffassign all clean.

### Findings (LOW only)

| F-1 | LOW | internal/connector/h1_redial_test.go | 152-163 | TestQuality | `TestRedialUpstreamH1_InvalidTarget` does not actually exercise `RedialUpstreamH1`'s invalid-target branch; it tests `extractHost("")` directly and assigns `_ = cfg` for an unused variable. Either restructure the test to construct a real stale Layer + invoke RedialUpstreamH1 with an empty target, or rename it to `TestExtractHost_Empty_ReturnsEmpty` so the test name matches what it exercises. |
| F-2 | LOW | internal/proxybuild/h1_chain_stale_recovery_integration_test.go | 284 | CodeHealth | `var _ = sync.Mutex{}` is a dead-import workaround for the unused `sync` import. Remove both the line and the `sync` import (only the test's other dependencies are actually used). |
| F-3 | LOW | internal/proxybuild/h1_chain_stale_recovery_integration_test.go | 884 | TestQuality | `dialInitialUpstream` passes `generation=0` to `RedialUpstreamH1`, which `redialConnIDLabel` maps to the `/upstream-redial` suffix (generation<=1 branch). So the *initial* Layer constructed by the test ends up tagged as a redial in its ConnID. Test assertions still pass (they check the EnvelopeContextTemplate's ConnID, not the streamID-suffix), but conceptually the initial Layer should not carry the redial label. Consider a dedicated bootstrap path that calls `dialUpstreamWithALPN` directly (or pass generation=-1 / suppress suffix for the initial). |
| F-4 | LOW | internal/layer/http1/healthcheck_test.go | 209-211, 345-347 | TestQuality | `TestHealthCheck_PeerClosed_ReturnsErr` and `TestHealthCheck_IdleByteFromUpstream_TreatedAsStale` use `time.Sleep(20ms)` to wait for OS state to propagate through loopback. This is flake-prone on slow / loaded CI. Replace with a polling assertion that retries the check (or the FIN-aware Read) up to e.g. 500ms or until the expected condition is observed. |

### Suggestions

- F-1: Easiest fix is to rename to `TestExtractHost_EmptyReturnsEmpty` and move it to a location that doesn't pretend to test the helper.
- F-2: Drop `"sync"` from the import block and the `var _ = sync.Mutex{}` line.
- F-3: Add an `if generation <= 0 { return baseConnID }` short-circuit in `redialConnIDLabel`, or have the test bootstrap call a non-redial helper.
- F-4: Wrap the assertion in an `until ... do sleep 10ms` style loop with a bounded deadline (already used elsewhere in the codebase, e.g. integration tests waiting for accept-loop state).

### STATS

| | Count |
|-|-|
| CRITICAL | 0 |
| HIGH | 0 |
| MEDIUM | 0 |
| LOW | 4 |
| NIT | 0 |

---

usk6666 (commented):
## Security Review: APPROVED

No security findings.

### Audit notes

- TLS upstream-dial path inherits the MITM-specific exception in `dialUpstreamWithALPN` — no MinVersion / cipher restriction is appropriate for a pentest proxy. The redial reuses the same path, no new TLS knobs introduced.
- ALPN advertisement narrowed to `["http/1.1"]` on redial. Intentional — prevents an inadvertent h2 upgrade that would invalidate the `*http1.Layer` Layer-type wired into the client side of the stack. This is a *narrowing* of negotiation, not a downgrade attack surface (the prior negotiation already produced http/1.1 by definition; the stack would have been on `*http2.Layer` otherwise).
- `HealthCheck`'s 1-byte idle peek may drop a surplus byte during idle when pendingQ is empty. Per RFC 9112 §9.6 this byte cannot be attributed to any HTTP transaction (idle keep-alive forbids data between exchanges). The trade-off is explicitly documented in the function godoc and accepted by the design review per Issue body — the connection is being terminated either way.
- `pendingMu` is held during the 1ms `conn.Read` syscall. No deadlock: `spawnLoopReceive` parks on `pendingNotify` (not `pendingMu`) when pendingQ is empty, and the chain's `c.mu` serializes EnsureFresh so the Close-stale path's internal `pendingMu` acquisition cannot race with a concurrent HealthCheck on the same Layer.
- `h1Chain.closeAll` iterates and Closes every redial step; `http1.Layer.Close` is `sync.Once`-guarded so the Close in `EnsureFresh` followed by `closeAll` is idempotent.
- No new `net/http` imports in data-path code. No header dedup / Host=URL enforcement / casing normalization introduced. MITM Principle #1 maintained.
- Test cert uses `crypto/rand` + ECDSA P-256 — correct crypto primitives.

### MITM threat-model fit

The change reduces the time window where the proxy holds a half-dead upstream conn — that is a defensive correctness improvement, not a normalization or wire-mutation change. It does not block any pentesting scenario.

### STATS

| | Count |
|-|-|
| CRITICAL | 0 |
| HIGH | 0 |
| MEDIUM | 0 |
| LOW | 0 |

---

PR #72, opened 2026-05-25
fix(layer/http1,proxybuild): replay-safe retry on EPIPE for stale h1 upstream (USK-999)

usk6666:
## Summary
- Add `*http1.StaleUpstreamError` + `isStaleConnErr` + `isReplaySafe` in `internal/layer/http1/channel.go`; translate EPIPE / ECONNRESET / io.EOF / `net.ErrClosed` from all three request-Send paths (zero-copy / patched-header / synthetic), header and body Writes.
- Add `h1Chain.Redial(ctx)` in `internal/proxybuild/h1_chain.go` — force-redial path complementary to Phase 1's `EnsureFresh` (HealthCheck short-circuit bypassed because the failed Write already proved the peer is gone).
- Add `retryingUpstreamChannel` in `internal/proxybuild/h1_retry.go`; wrap the dispatched h1 upstream channel in `runHTTP1Exchange.dial`. On `*StaleUpstreamError{ReplaySafe:true}` the wrapper consumes its one-shot retry budget, force-redials the chain, opens a fresh OpenExchange on the new Layer, re-applies the dispatch closure (so a gRPC-Web exchange stays a `grpcweb.Wrap` on the retry leg), atomically swaps the inner channel, and replays the same Envelope through Send exactly once.
- Closes the ~5% race window left by USK-998 Phase 1 (server FIN between HealthCheck and the proxy's first wire Write).

Browser-parity north star: when a browser would silently retry through `Connection: keep-alive` churn, the proxy should too — without surfacing as a hard `state=error`. (Mirrors the h2 USK-991/992/993 precedent on the h1 wire.)

## Test plan
- [x] `make lint` clean
- [x] `make build` succeeds
- [x] `make test` passes (fast tier)
- [x] `make test-e2e-smoke` passes (merge gate)
- [x] Unit: `TestIsReplaySafe_TruthTable` covers all method × body × stage combinations
- [x] Unit: `TestIsStaleConnErr_TypedSentinels` covers EPIPE / ECONNRESET / io.EOF / `net.ErrClosed` bare and wrapped through `*net.OpError` → `*os.SyscallError` chains
- [x] Unit: `TestStaleUpstreamError_UnwrapChain` verifies `errors.As` / `errors.Is`
- [x] Unit: `TestStaleUpstreamError_PluginStateNotDoubleFired` locks the no-double-fire invariant (failed-Send Channel close fires zero `ReleaseTransaction` calls — `emittedFlowIDs` populated only on parse)
- [x] Integration (`//go:build e2e && !e2e_smoke`, internal/proxybuild/h1_replay_retry_integration_test.go):
    - `TestHTTP1_StaleConn_GET_RetryPath` (GET race FIN → EPIPE → replay → success)
    - `TestHTTP1_StaleConn_POST_NoRetry` (POST → not replayed → error verbatim)
    - `TestHTTP1_StaleConn_PUT_WithBuffer_Retry` (PUT with BodyBuffer → replay succeeds)
    - `TestHTTP1_NoEnvelopeLossOnRetry` (response observed exactly once on the fresh leg)
    - `TestHTTP1_StaleConn_RetryBudgetOneShot` (budget consumed atomically)
    - `TestHTTP1_StaleConn_NonReplaySafe_ReturnsErrorVerbatim`
    - `TestHTTP1_FlakyConn_NonStaleErrorNotRetried`

Resolves USK-999
Linear: https://linear.app/usk6666/issue/USK-999

### DIFF
--- a/internal/layer/http1/channel.go
+++ b/internal/layer/http1/channel.go
@@ -6,11 +6,13 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"net"
 	"net/url"
 	"strconv"
 	"strings"
 	"sync"
 	"sync/atomic"
+	"syscall"
 	"time"
 
 	"github.com/google/uuid"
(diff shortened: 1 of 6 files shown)


usk6666 (commented):
## Code Review

**VERDICT: APPROVED**

**SUMMARY**: USK-998 Phase 2 retry wrapper is correctly scoped, well-documented, and the unit + integration tests pin both the replay-safety truth table and the one-shot budget invariant. The MITM-faithful approach is preserved (no header normalization, no Host=URL enforcement, no protocol downgrade). All unit + e2e_smoke + USK-999 integration tests pass under -race.

### FINDINGS

- **ID: F-1**
  - **Severity: LOW**
  - **File: internal/proxybuild/h1_retry.go**
  - **Line: 144**
  - **Category: Observability**
  - **Description**: When `chain.Redial` fails, the returned error wraps `derr` with `%w` but only includes the original `*StaleUpstreamError` via `%v`. Callers using `errors.As(returned, &*StaleUpstreamError{})` to read `Stage`/`ReplaySafe` on the original failure will no longer reach the original stale signal — only the redial error chain is reachable. The session records `state=error` with the dial error reason, losing the upstream-FIN diagnostic.
  - **Suggestion**: Either wrap both with `%w` via `errors.Join(derr, err)` (Go 1.20+) so both chains remain reachable, or document in the comment that the original stale error is intentionally informational-only in the redial-failure path.

- **ID: F-2**
  - **Severity: LOW**
  - **File: internal/proxybuild/h1_retry.go**
  - **Line: 128-135**
  - **Category: Code Health**
  - **Description**: The retry budget check uses both `r.retried.Load()` and `r.retried.CompareAndSwap(false, true)`. The CAS alone provides the one-shot guarantee; the prior Load is redundant. Not incorrect (it's a tiny early-exit optimization), but slightly non-idiomatic for atomic-budget patterns.
  - **Suggestion**: Drop the `Load` check and rely on CAS, or keep it but add a comment that the Load is a fast-path that the CAS validates.

- **ID: F-3**
  - **Severity: LOW**
  - **File: internal/proxybuild/h1_replay_retry_integration_test.go**
  - **Line: ~340-365 (TestHTTP1_StaleConn_RetryBudgetOneShot)**
  - **Category: Test Quality**
  - **Description**: The test name promises that "a wrapper that exhausts its 1-retry budget refuses a second retry," but the assertion only checks `wrapper.retried.Load() == true` after the first successful retry. It does not actually attempt a SECOND retry-triggering Send to confirm the wrapper rejects it. The one-shot contract is weakly verified vs. the test name.
  - **Suggestion**: After the first retry succeeds, trigger another EPIPE on the new inner (e.g. wrap chain.current's conn with a second flakyConn) and confirm the wrapper returns the StaleUpstreamError verbatim without firing a second `chain.Redial`. Alternatively, rename the test to match what it actually proves ("RetryBudgetFlippedAfterSuccess").

- **ID: F-4**
  - **Severity: NIT**
  - **File: internal/layer/http1/replay_safe_test.go**
  - **Line: 200-208**
  - **Category: Test Quality**
  - **Description**: `errorContains` reimplements `strings.Contains`. The comment claims independence from `strings` imports, but every other h1 test file in the package already imports `strings`.
  - **Suggestion**: Use `strings.Contains` directly.

- **ID: F-5**
  - **Severity: NIT**
  - **File: internal/proxybuild/h1_replay_retry_integration_test.go**
  - **Line: ~480-498 (replayStubChannel.Closed)**
  - **Category: Test Quality**
  - **Description**: `replayStubChannel.Closed()` uses the same `sync.Once` as `Close()` to lazily initialise `s.closed`. If a caller invokes `Closed()` before `Close()`, the Once consumes its slot creating an UNCLOSED channel, and a subsequent `Close()` becomes a no-op for the channel — leaving `Closed()` parking forever. This violates the layer.Channel contract that requires `Closed()` to fire when the channel terminates. Not exercised in the current test (the test only calls Send), but a latent fixture bug.
  - **Suggestion**: Initialise `s.closed` in a constructor or via a separate `sync.Once` from the close-signal Once.

- **ID: F-6**
  - **Severity: NIT**
  - **File: internal/proxybuild/h1_replay_retry_integration_test.go**
  - **Line: ~470 (replayStubChannel.Send)**
  - **Category: Test Quality**
  - **Description**: `Send` holds `s.mu` to read `s.sendErr` which is set at construction and never mutated. The lock is unnecessary.
  - **Suggestion**: Drop the mutex or document that it's defensive for future mutators.

### STATS
- CRITICAL: 0
- HIGH: 0
- MEDIUM: 0
- LOW: 3
- NIT: 3

LOW/NIT only → APPROVED.

usk6666 (commented):
## Security Review

**VERDICT: APPROVED**

**SUMMARY**: No CRITICAL/HIGH findings. The retry path is bounded by a one-shot atomic budget per per-exchange wrapper. No new attack surface, no DoS amplification beyond 2x, no new dependencies, no `unsafe`/`math/rand`/CRLF-injectable inputs. TLS and dial parameters are reused from the original CONNECT-time `BuildConfig` (per-host overlay, fingerprint, upstream proxy) — no relaxation of the floor.

### FINDINGS

(none)

### Notes on MITM threat model

- The retry wrapper does not modify any wire bytes — it replays the exact same `*envelope.Envelope` through a fresh upstream Channel. Header order, casing, duplicates, whitespace, and method are preserved byte-identical.
- `*StaleUpstreamError.Error()` includes the underlying syscall reason via `Unwrap`. This is debug-only via `slog.Debug` and contains no attacker-controlled or sensitive bytes; surface is limited to operator-visible debug logs.
- DoS: a malicious upstream forcing FIN-after-HealthCheck triggers at most one extra TLS handshake per client request (amp factor ≤ 2x). The CONNECT lifetime owns its h1Chain, so a long-lived attacker conn cannot fan out across other client sessions.
- `chain.Redial` is serialised via `chain.mu`; the per-exchange one-shot budget plus h1's wire-serial spawn loop (spawnLoopSend waits for `termDone`) means two Redials cannot race against the same Layer install.
- Self-signed cert + `InsecureSkipVerify: true` in the integration test (`generateSelfSignedCert`) is test-only and gated by `//go:build e2e && !e2e_smoke`.

### STATS
- CRITICAL: 0
- HIGH: 0
- MEDIUM: 0
- LOW: 0

LOW only (or none) → APPROVED.

PR #88, opened 2026-07-16
feat(tls): capture client ClientHello JA3/JA4 (USK-1015)

usk6666:
## Summary

Implements the dead `EnvelopeContext.TLS.ClientFingerprint` field (USK-1015). The MITM ClientHello peek already buffers the full hello but discarded everything past SNI+ALPN; this computes the client's **JA3** (MD5) and **JA4** (TLS client) fingerprints from the same buffer and surfaces them to plugins, the `query` MCP tool, and flow recording.

Wire fidelity is preserved: fingerprints are pure observation overlay — `env.Raw` and the peeked bytes are never mutated, and JA3 preserves wire order (JA4's internal cipher/extension sort is derived data only). Fail-soft on malformed/truncated input (empty result, never a panic).

### What changed
- **`internal/tlsfingerprint`** (new): stdlib-only (`crypto/md5`, `crypto/sha256`), self-contained JA3/JA4 computation. No new dependency; the FoxIO JA4 library is deliberately not imported (FoxIO-1.1 licence).
- **connector**: `ClientHelloPeek` carries `ClientJA3/ClientJA4`; both the sniff-first peek and the passthrough SNI peek compute them from the buffered hello. `performClientMITM` stamps them onto the client-facing `TLSSnapshot` for **every** MITM build branch (sniff / cache-hit / cache-miss / pool-hit / fallback), decoupled from the `peeked.ALPN` gate so ALPN-less / HTTP/1.1-only clients still get fingerprints.
- **envelope**: `TLSSnapshot` gains `ClientJA3/ClientJA4`; `ClientFingerprint` is set to `ClientJA4` for back-compat. `TLSHandshakeMessage` gains the fields for the passthrough audit flow.
- **pluginv2**: `ctx.tls` and `tls.on_handshake` dicts gain `ja3` / `ja4` keys (existing `client_fingerprint` key now populated).
- **flow**: `ConnectionInfo` + `StreamUpdate` + **schemaV18** columns `tls_client_ja3` / `tls_client_ja4`, wired through sqlite insert/scan/update, export, and import. `RecordStep` projects the client fingerprint from the Send-side snapshot without clobbering the upstream TLS columns.
- **mcp**: `query` `conn_info` surfaces `tls_client_ja3` / `tls_client_ja4`.

### Out of scope (deferred)
- Mirroring/replaying the client fingerprint to the upstream dial (capture only; noted "unimplemented" at the seam).
- ClientHellos larger than the 4 KiB peek cap are not fingerprinted — matches the existing SNI/ALPN silent fallback.
- Raw `raw_passthrough` (TLS-terminate + bytechunk) mode is not wired (its bytechunk layer does not stamp EnvelopeContext, so it has no flow surface today).

## Test plan
- `internal/tlsfingerprint`: table-driven golden vectors (JA3 MD5 + JA4 sections computed independently via Python), GREASE exclusion, no-SNI / no-ALPN, empty-cipher JA4_b sentinel, malformed/truncated no-panic, input-immutability, and a **real utls Firefox ClientHello** realism check.
- connector: sniff-first + passthrough peek fingerprint tests.
- flow: `ConnInfo` column round-trip (Save→Get), `UpdateStream` no-clobber, export/import round-trip.
- pluginv2: `ja3`/`ja4`/`client_fingerprint` dict keys.
- pipeline: Send-side projection sets only the fingerprint columns.
- `make lint` / `make build` / `make test` all green (49 packages, `-race`).

Resolves USK-1015

Linear: https://linear.app/usk6666/issue/USK-1015

### DIFF
--- a/internal/connector/connect_handler.go
+++ b/internal/connector/connect_handler.go
@@ -331,8 +331,7 @@ func runTLSMITM(
 	// USK-997: sniff-first peek. Test-only DisableClientHelloPeek bypasses.
 	var peeked ClientHelloPeek
 	if buildCfg == nil || !buildCfg.DisableClientHelloPeek {
-		sni, alpn := peekClientHelloSNIAndALPN(pc)
-		peeked = ClientHelloPeek{SNI: sni, ALPN: alpn}
+		peeked = peekClientHelloSNIAndALPN(pc)
 	}
 
 	stack, clientSnap, upstreamSnap, err := BuildConnectionStack(ctx, pc, target, buildCfg, peeked)
--- a/internal/connector/stack_builder.go
+++ b/internal/connector/stack_builder.go
@@ -640,6 +640,17 @@ type ClientHelloPeek struct {
 	// "forward this list verbatim to upstream and advertise upstream's
 	// pick back to the client".
 	ALPN []string
+
+	// ClientJA3 / ClientJA4 are the client's ClientHello fingerprints
+	// computed from the same peek buffer (USK-1015). Empty when the peek
+	// failed, the ClientHello was malformed, or it exceeded the peek cap.
+	// Unlike ALPN these are pure observation overlay — they never steer
+	// the build path; performClientMITM stamps them onto the client-facing
+	// TLSSnapshot so plugins / flow recording can surface them. Capture
+	// only: mirroring the client fingerprint to the upstream dial is
+	// unimplemented (deferred).
+	ClientJA3 string
+	ClientJA4 string
 }
 
 // BuildConnectionStack constructs a ConnectionStack for the given CONNECT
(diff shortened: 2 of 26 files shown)


usk6666 (commented):
## Code Review: APPROVED ✅

Clean, well-scoped implementation of the dead ClientFingerprint field. The new stdlib-only `internal/tlsfingerprint` package is self-contained and correct; JA3/JA4 derivation is spec-accurate (validated against independently-computed Python golden vectors), GREASE is stripped per RFC 8701, and wire order is preserved for JA3 while the JA4-internal sort operates only on derived hex copies. Schema V18 is a correct additive ALTER TABLE, and the RecordStep client-fingerprint projection uses the empty-field-skip no-clobber contract so it never overwrites the upstream TLS columns. Test coverage is strong: golden vectors, GREASE exclusion, no-SNI/no-ALPN, empty-cipher JA4_b sentinel, malformed/truncated no-panic, input immutability, and a real utls Firefox ClientHello realism check.

No findings.

---

usk6666 (commented):

The top security concern — parsing attacker-controlled ClientHello bytes — is handled correctly. Every length-prefixed field is bounds-checked through a single `cursor` abstraction plus explicit framing checks in `clientHelloBody`; there are no out-of-bounds reads, the extension walk always makes forward progress (no unbounded loop), allocations are bounded by the 4 KiB peek cap, and all decode paths are fail-soft (empty result, never a panic). Verified by the `TestComputeMalformed` table (nil / short header / non-handshake / truncated body / truncated mid-extensions / wrong handshake type / zero record length) and the input-immutability assertion — `env.Raw` and the peeked bytes are never mutated.

Schema migration V18 is an additive `ALTER TABLE ADD COLUMN ... DEFAULT ''` (no table recreate, FK-safe, backward-compatible). New columns store only derived fingerprints (MD5/SHA256-truncated) — no raw ClientHello or credential material. All SQL is parameterized.

No findings.

---

PR #91, opened 2026-07-19
test(e2e): fingerprint coherence harness (JA3/JA4 + H2 = Firefox)

usk6666:
## Summary
M47 capstone e2e verification: a fingerprint-coherence harness proving that when a browser `tls_fingerprint` profile is selected, the fingerprint the **upstream observes coming from the proxy** is coherent on both axes — TLS ClientHello (JA4 + cipher set) AND HTTP/2 send-shape (SETTINGS / WINDOW_UPDATE / pseudo-header order).

- **New capturing upstream** (`internal/mcptest/upstream_capture.go`, `UpstreamProto:"capture"`): peek-then-replays the raw proxy→upstream ClientHello off the `net.Conn`, hand-parses the proxy's H2 frames via `internal/layer/http2/frame` + hpack, answers with a minimal `:status 200` end-stream response. Captured artifacts exposed via `Harness.UpstreamCapture`. Single-writer teardown (`sync.Once`), per-conn deadlines, `Close()` joins all goroutines.
- **Test** (`internal/mcptest/fingerprint_coherence_integration_test.go`, `//go:build e2e && !e2e_smoke`): drives a real h2 request through the proxy over a CONNECT tunnel; **fails loudly** if the upstream leg's ALPN != h2. Per profile: JA4 asserted **exact** vs a reference computed dynamically from the same uTLS profile the proxy uses; cipher set asserted **structurally** (GREASE-stripped, order-independent — JA3 exact-match avoided due to uTLS extension shuffle); H2 goldens asserted inline (firefox FF120 vs chrome baseline). Recording verified via MCP `query` (Protocol/Scheme/State=complete, non-empty `raw_request` = env.Raw preserved).

## Test plan
- [x] Runs and passes under `make test-e2e` / `-tags e2e` (3× deterministic)
- [x] Excluded from fast + smoke tiers (`-tags 'e2e e2e_smoke'` → no tests to run)
- [x] `make lint`, `make build`, `make test`, `make test-e2e-smoke` green
- [x] `firefox` and `chrome` sub-cases pass; `none` sub-case `t.Skip`-ped pending USK-1021 (the real bug it caught), ready to un-skip once fixed

Resolves USK-1011
Linear: https://linear.app/usk6666/issue/USK-1011

### DIFF
--- a/internal/mcptest/fingerprint_coherence_integration_test.go
+++ b/internal/mcptest/fingerprint_coherence_integration_test.go
@@ -1,0 +1,564 @@
+//go:build e2e && !e2e_smoke
+
+// Package mcptest_test holds USK-1011's capstone verification for the M47
+// Firefox Fingerprint Fidelity milestone: when tls_fingerprint=firefox is
+// selected, the fingerprint the UPSTREAM observes coming FROM THE PROXY must
+// be Firefox-coherent on BOTH axes —
+//
+//   - TLS: the proxy→upstream ClientHello's JA4 equals a uTLS
+//     HelloFirefox_Auto reference (exact), and its cipher set matches
+//     (structural, order-independent — the "JA3 structural" check per the
+//     design review's U1: uTLS may shuffle extensions per handshake, so an
+//     exact JA3 wire-order hash would be flaky, whereas JA4's sorted form is
+//     shuffle- and GREASE-immune and therefore deterministic).
+//   - HTTP/2: the proxy's upstream SETTINGS (wire order + values), the
+//     stream-0 WINDOW_UPDATE increment, and the request pseudo-header order
+//     match the Firefox FF120 goldens (USK-1007).
+//
+// The chrome sub-test asserts the Chrome-shape baseline (both axes) and the
+// none sub-test asserts the standard-TLS / not-a-browser baseline, so an
+// accidental profile bleed (e.g. "none" silently dialing Firefox) is caught
+// as a regression rather than passing silently.
+//
+// Wiring: this drives a real HTTP/2 data-plane request through the proxy via
+// a CONNECT tunnel + client-offered ALPN h2, so the MITM'd upstream leg
+// negotiates h2 and the capture upstream (mcptest UpstreamProto="capture")
+// records the proxy's real on-wire ClientHello + H2 frames. It runs in the
+// exhaustive (non-smoke) e2e tier.
+package mcptest_test
+
+import (
+	"bufio"
+	"crypto/tls"
+	"encoding/base64"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net"
+	gohttp "net/http"
+	"sort"
+	"strings"
+	"testing"
+	"time"
+
+	utls "github.com/refraction-networking/utls"
+	"golang.org/x/net/http2"
+
+	"github.com/usk6666/yorishiro-proxy/internal/layer/http2/frame"
+	"github.com/usk6666/yorishiro-proxy/internal/mcptest"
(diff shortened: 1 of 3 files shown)


usk6666 (commented):
## Code Review: APPROVED

USK-1011 fingerprint-coherence e2e harness. Verified inline against CLAUDE.md Concurrency Checklist, e2e Subsystem Verification Checklist, and MITM-diagnostic test philosophy.

**Concurrency (capture upstream)** — sound:
- Every goroutine has explicit termination: `acceptLoop` exits on `ln.Close()`; per-conn `handleConn` bounded by a 10s `SetDeadline` + per-frame `SetReadDeadline`, so a missing frame cannot wedge teardown.
- Single-writer channel close: `done` closed exactly once via `doneOnce`; `Close()` guarded by `closeOnce`.
- `Close()` centralizes teardown (close listener → force-close tracked conns → `wg.Wait()` joins all goroutines). `trackConn` refuses registration after close, so no accept-vs-close leak. WaitGroup `Add` is always called while the accept goroutine's own +1 is outstanding (counter never hits zero mid-loop) — the safe pattern.
- `Snapshot()` deep-copies all slice fields; captured fields guarded by `mu`; `done` closed under `mu` after fields are set, giving WaitH2→Snapshot a clean happens-before. Runs green under `-race`.

**Correctness** — peek-then-replay reads exactly one TLS record (fits every profile's hello) and hands a `prefixConn` to `tls.Server`, so the raw ClientHello is retained byte-for-byte for `tlsfingerprint.Compute` while the rest of the handshake reads live. JA4-exact vs a dynamically-built uTLS reference (same HelloID + SNI + forwarded ALPN list) with cipher-set structural check is the correct shuffle/GREASE-immune split (design U1). H2 goldens (FF120 SETTINGS 65536/0/131072/16384, WINDOW_UPDATE 12517377, pseudo :method :path :authority :scheme) match USK-1007. frame/hpack API usage verified. Fresh HPACK decoder is correct for the single first-request HEADERS block.

**Test integrity** — asserts recording OUTCOMES (JA4 equality, exact H2 shape, Stream Protocol/Scheme/State, env.Raw via non-empty base64 `raw_request`), not "bytes flowed". `none` sub-case is `t.Skip`-ped referencing USK-1021 (a real bug it caught), assertions un-weakened. Build tags `e2e && !e2e_smoke` correct; ephemeral ports throughout.

Verified locally: `make lint` + `make build` green; test passes 3x under `-tags e2e` (firefox PASS, chrome PASS, none SKIP→USK-1021); excluded from the smoke tier ("no tests to run") while the untagged harness support still compiles.

No findings.

---

usk6666 (commented):
## Security Review: APPROVED

Scope is light — this is test-harness code (`internal/mcptest`), not the data path. Focus was the capture upstream's parse of proxy-sent bytes.

- **Bounded parsing of proxy frames**: `readFirstTLSRecord` rejects implausible record lengths (`<=0` or `>1<<16`); the H2 frame reader enforces `MaxFrameSize`; the HPACK decoder is capped at a 65536 table. Per-read deadlines (10s) prevent a stalled/partial input from hanging the harness. No panic surface on malformed frames — errors return and tear the conn down.
- **No CA-key / sensitive-material handling**; the capture upstream mints its own self-signed leaf. All listeners are loopback + ephemeral, test-controlled (no attacker-reachable surface).
- `InsecureSkipVerify` is scoped to the test data-plane client accepting the proxy's MITM leaf and is properly `//nolint:gosec` annotated.
- No new external dependency (utls, x/net/http2 already in go.mod; no go.mod/go.sum change).

No findings.

---

PR #93, opened 2026-07-27
fix(tls): resolve "none" fingerprint sentinel at the uTLS dial seam

usk6666:
## Summary
- `config.ResolveTLSFingerprint` is now **identity-preserving**: it still owns the `firefox` default and the proxyCfg → cfg → default precedence, but returns the `"none"` opt-out sentinel verbatim instead of flattening it to `""`.
- New `config.UTLSProfileFor(fingerprint) string` owns the `"none"` escape hatch (case-insensitive, trimmed). Only that sentinel maps to `""`; everything else passes through, so `tlslayer`'s fail-hard-on-typo property survives (a `"firefx"` typo must never silently degrade to a Go-native ClientHello).
- New `connector.BuildConfig.EffectiveUTLSProfile()` derives the **dial** value from `EffectiveTLSFingerprint()`. The two `DialRawOpts.UTLSProfile` sites (`stack_builder.go`) and the two cache-key sites (`h2_pool.go`, `alpn_routing.go`) switch to it.
- The four `WithH2Fingerprint` sites intentionally keep reading `EffectiveTLSFingerprint()` — `resolveH2Fingerprint` is already sentinel-aware and the resulting wire is byte-identical either way.
- `InitTLSTransport` (resend axis) becomes `UTLSProfileFor(ResolveTLSFingerprint(...))`, preserving its current end behavior.

**Root cause**: `-tls-fingerprint none` resolved correctly at boot but was re-broken by `proxy_start`. `applyProxyDefaultTLSStrings` copied the raw `"none"` into the input, `applyTLSFingerprint` installed it as the dynamic live-dial override, and `tlslayer` hard-errored on every MITM upstream dial with `unsupported uTLS profile "none"` (the client leg then EOF'd).

**Why not resolve `"none"` → `""` before `manager.SetTLSFingerprint`** (the originally-filed fix direction): `BuildConfig.SetTLSFingerprint("")` *clears* the override, so an explicit runtime `none` on top of a boot-time `firefox` would have silently fallen back to spoofing Firefox. Putting the resolution at the dial seam instead makes that failure mode structurally unreachable, keeps the two-state override channel consistent with its two sibling knobs (`upstreamProxyDynamic`, `maxConcurrentStreamsDynamic`), and changes **zero lines** in `internal/mcp/` production code and `internal/layer/`.

**Invariant after this change**: `EffectiveTLSFingerprint()` = the identity the proxy claims (reporting, H2 shape); `EffectiveUTLSProfile()` = the uTLS parrot to dial, with `""` meaning standard crypto/tls.

## Behavior change (intentional)
`NewLiveBuildConfig` now **stores** `"none"` rather than flattening it to `""`. A server booted with `-tls-fingerprint none` therefore reports `tls_fingerprint: "none"` from `query config` / `query status` instead of `""`.

This closes a pre-existing reporting asymmetry: boot-`none` reported `""` while a runtime `configure{tls_fingerprint:"none"}` reported `"none"`. `""` is already pinned as "no fingerprint configured" (`query_tool_test.go`), and the `"none"` round-trip is pinned by two existing fast tests, so `"none"` is the correct spelling for an explicit opt-out. Reporting `""` for an explicit opt-out would be the USK-809 lying-reporter class.

## Test plan
- [x] `internal/config/config_tls_test.go` — new `TestUTLSProfileFor` (9 rows: `none`/`NONE`/`  none  `/`  NoNe  ` → `""`; `firefox`/`chrome`/`""`/`firefx`/`nonexistent` passthrough). `TestResolveTLSFingerprint`'s three `none` rows updated to the identity-preserving expectation.
- [x] `internal/connector/stack_builder_test.go` — new `TestDialUpstreamWithALPN_NoneFingerprintDialsStandardTLS` performs a **real TLS handshake** through `dialUpstreamWithALPN` against a local server, so a regression at the `DialRawOpts.UTLSProfile` call sites (not just the accessor) fails the fast tier. Covers Case A (boot `none`), Case B (runtime `none` over boot `firefox`), a spelling variant, and a `firefox` positive control.
- [x] `internal/connector/stack_builder_test.go` — new `TestBuildConfig_EffectiveUTLSProfile` (11 rows) asserting both the dial value **and** that the claimed identity is untouched; plus `TestALPNCacheKeyFromConfig_NoneResolvesToDialIdentity` and `TestPoolKeyForH2_NoneResolvesToDialIdentity` for cache-key coherence.
- [x] `TestBuildConfig_TLSFingerprint_DynamicOverride` left **unchanged** — it proves the two-state clear semantics survives.
- [x] `internal/mcp/configure_tls_fingerprint_test.go` — Case A and Case B through the **real MCP surface** with a bound `connector.BuildConfig`: `proxy_start` inheriting `proxyDefaults{TLSFingerprint:"none"}`, `proxy_start{tls_fingerprint:"none"}` over boot-`firefox`, and `configure{tls_fingerprint:"none"}` over boot-`firefox` (with a switch-back-to-firefox positive control). Existing `"none"` round-trip assertions untouched and green.
- [x] `internal/mcp/query_config_resource_test.go` — existing `{"none","none"}` and clear-semantics cases green **unchanged** (regression guard against a fix that flattens the reported value).
- [x] `internal/mcpserver/init_test.go` — `TestNewLiveBuildConfig_TLSFingerprintDefault`'s `none` row updated to `wantStored:"none"` / `wantEffect:"none"` and a new `wantUTLS:""` column added across all rows.
- [x] `internal/layer/tlslayer/tlslayer_test.go` — `TestClient_UnsupportedUTLSProfile` table-ised with a `"none"` row documenting that this layer must **keep** erroring, so the resolution boundary cannot migrate down into `internal/layer/`.
- [x] `internal/mcptest/fingerprint_coherence_integration_test.go` — the `none` sub-case is un-skipped with all assertions unweakened (JA4 ≠ firefox-ref, JA4 ≠ chrome-ref, zero GREASE ciphers, default H2 shape) and the `//go:build e2e && !e2e_smoke` tag preserved. Cosmetic: `chromeH2Shape` → `defaultH2Shape` (it is the proxy's historical send-shape, not a real Chrome 120 capture).
- [x] Negative control verified: temporarily reverting `EffectiveUTLSProfile` to return the raw identity fails `TestDialUpstreamWithALPN_NoneFingerprintDialsStandardTLS`, `TestBuildConfig_EffectiveUTLSProfile`, and `TestALPNCacheKeyFromConfig_NoneResolvesToDialIdentity`.
- [x] `make lint`, `make build`, `make test` (fast tier) all pass; `go vet` clean under `-tags e2e` and `-tags 'e2e e2e_smoke'`.

## Not executed
The un-skipped `none` e2e sub-case was **not run** — the full tier needs network and a live capture upstream. It compiles clean under `-tags e2e` and `-tags 'e2e e2e_smoke'`; the fast-tier tests above lock Case A and Case B deterministically. It will first execute in the nightly full-tier run.

Resolves USK-1021
Linear: https://linear.app/usk6666/issue/USK-1021

https://claude.ai/code/session_013th9oiLKhMcQCNf3eQmsYR

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -339,19 +339,23 @@ func ResolveDBPath(value string) (string, error) {
 // Chrome parity remains available via explicit configuration.
 const DefaultTLSFingerprint = "firefox"
 
-// ResolveTLSFingerprint returns the effective upstream TLS ClientHello
-// fingerprint profile name, applying the precedence:
+// ResolveTLSFingerprint returns the fingerprint *identity* the proxy claims
+// for upstream connections, applying the precedence:
 //
 //  1. proxyCfg.TLSFingerprint (per-listener config / `-tls-fingerprint` flag)
 //  2. cfg.TLSFingerprint (top-level config; pass a nil cfg to skip this tier
 //     and preserve the live-dial path's proxyCfg-only contract)
 //  3. DefaultTLSFingerprint ("firefox")
 //
-// The opt-out sentinel "none" (case-insensitive) resolves to "" so the caller
-// dials with the standard crypto/tls stack (no uTLS spoofing). Any other value
-// passes through unchanged for downstream profile validation. This is the
-// single source of truth for the firefox default and the "none" escape hatch;
-// call it from every boot-time fingerprint resolution site (USK-1013).
+// The returned value is identity-preserving: the opt-out sentinel "none" is
+// returned verbatim rather than flattened to "", so the reporting surfaces
+// (query config / configure results) and the HTTP/2 send-shape selection can
+// distinguish "operator explicitly opted out" from "nothing configured".
+// Mapping the sentinel to the uTLS parrot name is the separate job of
+// UTLSProfileFor, which must be applied at the dial seam.
+//
+// This is the single source of truth for the firefox default (USK-1013);
+// call it from every boot-time fingerprint resolution site.
 func ResolveTLSFingerprint(cfg *Config, proxyCfg *ProxyConfig) string {
 	fingerprint := ""
 	if proxyCfg != nil && proxyCfg.TLSFingerprint != "" {
@@ -362,6 +366,22 @@ func ResolveTLSFingerprint(cfg *Config, proxyCfg *ProxyConfig) string {
 	if fingerprint == "" {
 		fingerprint = DefaultTLSFingerprint
 	}
+	return fingerprint
+}
+
+// UTLSProfileFor maps a fingerprint identity to the uTLS parrot profile name
+// the dial path should instantiate. The opt-out sentinel "none"
+// (case-insensitive, surrounding whitespace trimmed) maps to "", which every
+// dial seam reads as "use the standard crypto/tls stack, no uTLS spoofing".
+// Every other value — including "" itself — passes through unchanged so the
+// downstream profile lookup keeps its fail-hard-on-typo property: a
+// misspelled "firefx" must error loudly rather than silently degrade to a
+// Go-native ClientHello (USK-1021).
+//
+// This is the single source of truth for the "none" escape hatch. Apply it at
+// the uTLS-profile consumption seam only; the identity carried through config,
+// runtime overrides, and reporting stays unresolved.
+func UTLSProfileFor(fingerprint string) string {
 	if strings.EqualFold(strings.TrimSpace(fingerprint), "none") {
 		return ""
 	}
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -349,10 +349,12 @@ const DefaultTLSFingerprint = "firefox"
 //
 // The returned value is identity-preserving: the opt-out sentinel "none" is
 // returned verbatim rather than flattened to "", so the reporting surfaces
-// (query config / configure results) and the HTTP/2 send-shape selection can
-// distinguish "operator explicitly opted out" from "nothing configured".
-// Mapping the sentinel to the uTLS parrot name is the separate job of
-// UTLSProfileFor, which must be applied at the dial seam.
+// (query config / configure results) can distinguish "operator explicitly
+// opted out" from "nothing configured". The HTTP/2 send-shape selection is
+// only sentinel-tolerant — it accepts the raw identity but maps both "none"
+// and "" to the default shape. Mapping the sentinel to the uTLS parrot name
+// is the separate job of UTLSProfileFor, which must be applied at the dial
+// seam.
 //
 // This is the single source of truth for the firefox default (USK-1013);
 // call it from every boot-time fingerprint resolution site.
--- a/internal/config/config_tls_test.go
+++ b/internal/config/config_tls_test.go
@@ -55,8 +55,12 @@ func TestLoadFile_TLSFingerprint(t *testing.T) {
 }
 
 // TestResolveTLSFingerprint covers the boot-time fingerprint resolution
-// precedence (proxyCfg > cfg > firefox default) and the "none" opt-out
-// sentinel that maps to "" (standard crypto/tls) — USK-1013.
+// precedence (proxyCfg > cfg > firefox default) — USK-1013.
+//
+// USK-1021: the resolver is identity-preserving. The "none" opt-out sentinel
+// is returned verbatim so reporting can distinguish "explicitly opted out"
+// from "nothing configured"; mapping it to the uTLS parrot name ("") is
+// UTLSProfileFor's job and happens at the dial seam.
 func TestResolveTLSFingerprint(t *testing.T) {
 	tests := []struct {
 		name     string
@@ -95,22 +99,22 @@ func TestResolveTLSFingerprint(t *testing.T) {
 			want:     "firefox",
 		},
 		{
-			name:     "explicit none maps to empty (standard TLS)",
+			name:     "explicit none is preserved as an identity",
 			cfg:      &Config{},
 			proxyCfg: &ProxyConfig{TLSFingerprint: "none"},
-			want:     "",
+			want:     "none",
 		},
 		{
-			name:     "none is case-insensitive and trimmed",
+			name:     "none spelling is preserved verbatim (no fold/trim)",
 			cfg:      &Config{},
 			proxyCfg: &ProxyConfig{TLSFingerprint: "  NONE  "},
-			want:     "",
+			want:     "  NONE  ",
 		},
 		{
-			name:     "top-level none also opts out",
+			name:     "top-level none is preserved as an identity",
 			cfg:      &Config{TLSFingerprint: "none"},
 			proxyCfg: &ProxyConfig{},
-			want:     "",
+			want:     "none",
 		},
 		{
 			name:     "explicit firefox passes through",
@@ -123,6 +127,36 @@ func TestResolveTLSFingerprint(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			if got := ResolveTLSFingerprint(tt.cfg, tt.proxyCfg); got != tt.want {
 				t.Errorf("ResolveTLSFingerprint() = %q, want %q", got, tt.want)
+			}
+		})
+	}
+}
+
+// TestUTLSProfileFor covers the "none" opt-out sentinel mapping applied at
+// the uTLS dial seam (USK-1021). Only "none" (case-insensitive, trimmed)
+// collapses to ""; every other value — including a typo — passes through
+// unchanged so the tlslayer profile lookup keeps its fail-hard property
+// rather than silently degrading to a Go-native ClientHello.
+func TestUTLSProfileFor(t *testing.T) {
+	tests := []struct {
+		name        string
+		fingerprint string
+		want        string
+	}{
+		{"none opts out", "none", ""},
+		{"none is case-insensitive", "NONE", ""},
+		{"none is trimmed", "  none  ", ""},
+		{"none is trimmed and folded", "  NoNe  ", ""},
+		{"firefox passes through", "firefox", "firefox"},
+		{"chrome passes through", "chrome", "chrome"},
+		{"empty passes through", "", ""},
+		{"unknown profile passes through for downstream validation", "firefx", "firefx"},
+		{"substring of none is not the sentinel", "nonexistent", "nonexistent"},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := UTLSProfileFor(tt.fingerprint); got != tt.want {
+				t.Errorf("UTLSProfileFor(%q) = %q, want %q", tt.fingerprint, got, tt.want)
 			}
 		})
 	}
--- a/internal/connector/alpn_routing.go
+++ b/internal/connector/alpn_routing.go
@@ -217,14 +217,17 @@ func canonicalRedialALPNOffer(clientALPN string) []string {
 
 // ALPNCacheKeyFromConfig constructs an ALPNCacheKey for the given target
 // using the TLS configuration from BuildConfig. The fingerprint
-// component reads via EffectiveTLSFingerprint so a runtime
+// component reads via EffectiveUTLSProfile so a runtime
 // proxy_start / configure fingerprint switch produces a distinct cache
 // key — ALPN learned under the previous fingerprint is not reused for
-// dials that should now use a different uTLS profile (USK-809).
+// dials that should now use a different uTLS profile (USK-809). The
+// resolved (dial-identity) form is used so a boot-time "none" and a
+// runtime "none" share one keyspace, matching the handshakes they
+// actually key (USK-1021).
 func ALPNCacheKeyFromConfig(target string, cfg *BuildConfig) ALPNCacheKey {
 	key := ALPNCacheKey{
 		HostPort:    target,
-		Fingerprint: cfg.EffectiveTLSFingerprint(),
+		Fingerprint: cfg.EffectiveUTLSProfile(),
 	}
 	if cfg.ClientCert != nil {
 		key.ClientCertHash = hashCert(cfg.ClientCert)
--- a/internal/connector/h2_pool.go
+++ b/internal/connector/h2_pool.go
@@ -71,13 +71,16 @@ func poolKeyForH2(ctx context.Context, target string, cfg *BuildConfig, hostTLS 
 
 	writeField(&buf, "ca", caHash)
 
-	// Use EffectiveTLSFingerprint so a runtime proxy_start / configure
+	// Use EffectiveUTLSProfile so a runtime proxy_start / configure
 	// fingerprint switch produces a distinct pool key — pooled h2 Layers
 	// established under the previous fingerprint must not be reused for
-	// dials that should now use a different uTLS profile (USK-809).
+	// dials that should now use a different uTLS profile (USK-809). The
+	// resolved (dial-identity) form is used so a boot-time "none" and a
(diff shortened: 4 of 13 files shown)


usk6666 (commented):
## Code Review: APPROVED ✅

The identity/dial split is correctly implemented and the conversion at the dial seam is **complete**: I traced every path that can reach `tlslayer.Client` and confirmed all of them now carry a resolved profile. `DialRawOpts.UTLSProfile` is assigned in exactly two places (`stack_builder.go:1612` `dialUpstreamWithALPN`, `stack_builder.go:2010` `buildRawPassthroughStack`) and both were converted; every other `DialUpstreamRaw` caller (`tls_passthrough.go` — forces `TLSConfig=nil`, `connect_inner_dispatch.go` `dialPlainUpstream`, `http1_forward_handler.go`, `proxybuild/tcp_forward_tls.go` — documented USK-916 `UTLSProfile=""` deferral, `mcp/fuzz_raw_helpers.go`) leaves the field zero. `h1_redial.go` / `h2_redial.go` / `stack_builder_target_override.go` all funnel through `dialUpstreamWithALPN`, so they inherit the fix. `ALPNCacheKeyFromConfig` and `poolKeyForH2` are the only two fingerprint-keyed caches, and both were converted — correctly, since boot-`none` / runtime-`none` / unset all mint byte-identical ClientHellos and therefore belong in one keyspace. No direct `bc.TLSFingerprint` field read survives outside the accessor. `EffectiveUTLSProfile` is nil-safe via `EffectiveTLSFingerprint`'s guard, and lock-free (atomic pointer load) so it adds no contention to the dial hot path.

I also verified the two settled design decisions that depend on external code: `resolveH2Fingerprint` (`internal/layer/http2/fingerprint.go:64`) folds/trims and special-cases only `"firefox"`, so `"none"` and `""` both yield `H2FingerprintDefault` — keeping the four `WithH2Fingerprint` sites on `EffectiveTLSFingerprint()` is genuinely byte-identical on the wire. And `InitTLSTransport`'s `UTLSProfileFor(ResolveTLSFingerprint(...))` preserves its prior end behavior exactly.

Findings are all documentation accuracy and test-coverage polish — no correctness, concurrency, or architecture issues.

### Findings

| F-1 | MEDIUM | internal/connector/stack_builder.go | 52-58, 549-556 | GoConventions | `EffectiveTLSFingerprint`'s godoc still declares itself the dial accessor, contradicting the new `EffectiveUTLSProfile` godoc 20 lines below |
| F-2 | LOW | internal/connector/stack_builder_test.go | 1130-1140 | TestQuality | The second `DialRawOpts.UTLSProfile` site (`buildRawPassthroughStack`) has no `none` regression coverage, though the test godoc claims to cover "the call sites" (plural) |
| F-3 | LOW | internal/connector/stack_builder_test.go | 1149, 1188-1198 | TestQuality | The "firefox positive control" row does not verify uTLS was actually used — only that the handshake did not error |
| F-4 | LOW | internal/proxybuild/manager.go | 945-952 | GoConventions | `Manager.TLSFingerprint()` godoc says "effective uTLS browser fingerprint profile"; it can now return `"none"`, which is not a uTLS profile |
| F-5 | NIT | internal/config/config.go | 350-355 | GoConventions | Godoc overclaims that the H2 send-shape selection "can distinguish" opted-out from unconfigured; `resolveH2Fingerprint` maps both to the default shape |
| F-6 | NIT | internal/connector/stack_builder_test.go | 1092 | GoConventions | `nilReceive` reads better as `nilReceiver` |

### Suggestions

**F-1** — This is the exact trap that reintroduces USK-1021. `EffectiveTLSFingerprint`'s godoc currently says it "returns the uTLS browser fingerprint profile the live data path should use for the next upstream dial", that it "Returns the empty string when neither is configured (the dial path then falls back to standard TLS)", and that it is "the canonical accessor for live dial-path code (USK-809)". All three clauses are now false, and the `BuildConfig.TLSFingerprint` field doc at line 55 repeats the instruction ("Live data-path readers MUST call `EffectiveTLSFingerprint()`"). A future contributor adding a dial site reads the *older* accessor's doc first and picks the wrong one. Retarget both to the identity role and point the dial instruction at `EffectiveUTLSProfile` — e.g. "returns the fingerprint *identity* the proxy claims … Live **dial-path** readers MUST call `EffectiveUTLSProfile()` instead; this accessor is for reporting and H2 send-shape selection (USK-1021)."

**F-2** — `TestDialUpstreamWithALPN_NoneFingerprintDialsStandardTLS` pins only `stack_builder.go:1612`. Reverting `stack_builder.go:2010` alone would leave the bug live for any deployment combining `tls_fingerprint: none` with `raw_passthrough_hosts`, and no fast-tier test would fail. `TestBuildConnectionStack_RawPassthrough` (same file, line 21) already builds the whole MITM + upstream-dial harness and leaves `TLSFingerprint` unset — adding a `TLSFingerprint: "none"` variant (or table-ising it over `{"", "none", "firefox"}`) closes the gap cheaply. Also soften the test godoc from "the `DialRawOpts.UTLSProfile` call sites" to name the one it actually covers.

**F-3** — A hypothetical regression where `EffectiveUTLSProfile` returned `""` unconditionally would pass the firefox row, since a standard-TLS handshake against the test server also succeeds. To make it a real control, capture the ClientHello server-side via `tls.Config.GetConfigForClient` (`hello.CipherSuites` is populated with the raw parsed cipher IDs) and assert the firefox row's cipher list differs from the none row's. Comparing the two observed lists is more robust than asserting GREASE presence, since Firefox GREASE behaviour is version-dependent — note the e2e helper `fpCipherSet` strips GREASE for both profiles rather than relying on it.

**F-4** — Reword to "the claimed fingerprint identity from the bound BuildConfig … may be the `none` opt-out sentinel; this is a reporting accessor, not a dial value". All four consumers (`configure_tool.go:929`, `query_tool.go:2263` / `2461`, via `currentTLSFingerprint`) are reporting, so the behaviour is right — only the wording is stale.

**F-5** — Suggested wording: "…so the reporting surfaces (query config / configure results) can distinguish 'operator explicitly opted out' from 'nothing configured'. The HTTP/2 send-shape selection is sentinel-tolerant — `resolveH2Fingerprint` maps both to the default shape — so it is unaffected either way." The same overclaim appears at `stack_builder.go:576-578`.

### Verified, not flagged

- `internal/layer/tlslayer` correctly still fail-hards on `"none"`; the added table row documents that boundary. `Client()` branches on `opts.UTLSProfile != ""` *before* trimming, so a whitespace-only fingerprint still errors loudly — pre-existing and consistent with the fail-hard-on-typo guarantee.
- `SetTLSFingerprint("")` clear-semantics untouched; `TestBuildConfig_TLSFingerprint_DynamicOverride` left green as the guard.
- `applyTLSFingerprint` lowercases and allow-lists before `manager.SetTLSFingerprint`, so the trim/fold inside `UTLSProfileFor` is pure defence-in-depth for direct `BuildConfig` construction — correct to have, not dead code.
- New MCP tests bind a real `*connector.BuildConfig` through `newTestProxybuildManagerWithBuildConfig`, whose `t.Cleanup` runs `StopAll` — no listener leak from the `proxy_start` calls. `t.Helper()` present on both new helpers.
- e2e un-skip keeps all four `none` assertions (JA4 ≠ firefox-ref, JA4 ≠ chrome-ref, zero GREASE ciphers, default H2 shape) and the `e2e && !e2e_smoke` tag. `chromeH2Shape` → `defaultH2Shape` rename is complete with no stragglers.
- `gofmt -l` clean across all 11 changed files.

---

usk6666 (commented):

The identity/dial split is correctly and completely applied: every path that reaches `tlslayer` (both `DialRawOpts.UTLSProfile` sites, `InitTLSTransport`) now carries the resolved profile, and every path that must carry the claimed identity (4× `WithH2Fingerprint`, `Manager.TLSFingerprint()` reporting) still does. No fingerprint lie, no fail-open, no new race, no cross-fingerprint cache reuse. Three LOW notes below, two of which are pre-existing.

### Verification performed

**1. Fingerprint-identity confusion — clean.** Exhaustive grep of `EffectiveTLSFingerprint` / `EffectiveUTLSProfile` / `UTLSProfileFor` / `UTLSProfile` over all non-test code yields exactly two partitions with no leakage:

- dial value (`EffectiveUTLSProfile`): `stack_builder.go:1612`, `stack_builder.go:2010`, `h2_pool.go:83`, `alpn_routing.go:230`, `mcpserver/init.go:753`
- identity (`EffectiveTLSFingerprint`): `stack_builder.go:1777`, `connect_inner_dispatch.go:237`, `h2_redial.go:111`, `stack_builder_target_override.go:460`, `proxybuild/manager.go:961`

The four identity sites are all `WithH2Fingerprint`, and `resolveH2Fingerprint` (`layer/http2/fingerprint.go:64`) trims + lowercases and only special-cases `"firefox"`, so `"none"` and `""` both yield `H2FingerprintDefault` — byte-identical wire either way, as the PR claims. The other two `DialRawOpts` constructors in the tree (`proxybuild/tcp_forward_tls.go:465`, `mcp/fuzz_raw_helpers.go:996`) never set `UTLSProfile` at all, so they are unaffected.

**2. Cache-key poisoning — no live collision.** `UTLSProfileFor` is the identity map except on the `"none"` fold-set, so the only identities that now collapse onto one key are `{"", "none", "NONE", "  none  ", …}`. All members produce the same dial (`clientStandard`) *and* the same H2 send-shape (`H2FingerprintDefault`), so the merge is sound. No other pair collides. In production the bare-`""` identity is unreachable anyway — `NewLiveBuildConfig` always resolves to at least `firefox`, and `SetTLSFingerprint("")` falls back to that static value. Note this change actually *restores* the documented contract in `alpn_cache.go:35-37` (`Fingerprint == ""` means the standard crypto/tls library was used), which pre-PR runtime-`none` violated by storing `"none"`.

**3. Sentinel robustness — no Unicode surprise.** I enumerated the full `unicode.SimpleFold` orbit over all of U+0000–U+10FFFF: no non-ASCII rune folds into `n`, `o`, or `e`. `strings.EqualFold(x, "none")` therefore matches exactly the 16 ASCII case permutations — no homoglyph or Kelvin-sign-class bypass. `TrimSpace` additionally absorbs Unicode whitespace (NBSP, NEL), which only widens the opt-out set in the non-spoofing direction. The value is operator-controlled, and both operator entry points that matter (`ApplyTLSFingerprintFlag` in `mcpserver/server.go:469`, `applyTLSFingerprint` in `mcp/proxy_start_tool.go:1163`) lowercase and whitelist-validate against the 6 accepted names.

**4. Fail-open vs fail-closed — preserved.** `internal/layer/tlslayer/client.go:83-86` is untouched and still hard-errors on any name outside the parrot table; `"none"` reaching it errors too, and the new `TestClient_UnsupportedUTLSProfile/none` row pins the resolution boundary so it cannot migrate down into `internal/layer/`. `UTLSProfileFor` maps *only* the exact sentinel to `""` — `"firefx"` passes through and still fails hard on the live dial path.

**5. Nil-safety and concurrency — sound.** `EffectiveUTLSProfile` performs exactly one `atomic.Pointer.Load` (via `EffectiveTLSFingerprint`, which nil-checks its receiver) and then applies a pure function; no new load, no torn read, nil-receiver safe (covered by the `nil receiver` table row). The read count and read points on the dial hot path are unchanged from pre-PR.

Verified `TestUTLSProfileFor`, `TestResolveTLSFingerprint`, `TestBuildConfig_EffectiveUTLSProfile`, `TestDialUpstreamWithALPN_NoneFingerprintDialsStandardTLS`, `TestALPNCacheKeyFromConfig_NoneResolvesToDialIdentity`, `TestPoolKeyForH2_NoneResolvesToDialIdentity` and `TestClient_UnsupportedUTLSProfile` pass under `-race`. No new dependencies.

### LOW findings (non-blocking)

| S-1 | LOW | internal/connector/h2_pool.go, internal/connector/alpn_routing.go | 74-85, 218-233 | TLS/Cert | CWE-1023 | Cache keys now use the many-to-one dial value; the "no two identities with different H2 shapes may collapse" invariant is load-bearing but unstated |
| S-2 | LOW | internal/connector/stack_builder_target_override.go | 755 | TLS/Cert | CWE-1059 | Exported `DialUpstreamWithALPN` contract still names `EffectiveTLSFingerprint` as governing uTLS selection |
| S-3 | LOW | internal/mcpserver/init.go | 753-762 | TLS/Cert | CWE-636 | Pre-existing: unvalidated config-file `tls_fingerprint` typo fails open to standard TLS on the resend axis |

**S-1 — cache key collapses distinct identities (LOW, no live exploit).**
`poolKeyForH2` and `ALPNCacheKeyFromConfig` moved from the identity to `EffectiveUTLSProfile()`, which is many-to-one. Today the only collapsed set is the `none`-variants plus `""`, and every member yields both the same TLS handshake *and* the same `resolveH2Fingerprint` result, so a pooled h2 Layer or cached ALPN entry can never be reused across genuinely different wire shapes. The change is correct as written. What is new is an unstated invariant: **`config.UTLSProfileFor` must never merge two identities that `layer/http2.resolveH2Fingerprint` distinguishes**, because the pooled h2 connection's SETTINGS / WINDOW_UPDATE / pseudo-header shape is fixed at creation from the *identity* while the pool key is now derived from the *dial value*. Per CLAUDE.md MITM Principle #6 this is not a live vulnerability — it requires architecture not yet present.

*Re-open trigger*: a second sentinel or alias is added to `UTLSProfileFor`, **or** `resolveH2Fingerprint` starts discriminating any profile beyond `"firefox"`. Suggested (optional) remediation: add one sentence to the `EffectiveUTLSProfile` godoc stating the invariant, so the next person to touch either function sees the coupling.

**S-2 — stale contract on the exported forward-dial seam (LOW, doc-only).**
`stack_builder_target_override.go:755` documents `DialUpstreamWithALPN` (the USK-912 export consumed by the TCP-forward path) as: *"cfg.EffectiveTLSFingerprint and cfg.EffectiveUpstreamProxyForCtx continue to govern uTLS profile selection…"*. Post-PR the governing accessor is `EffectiveUTLSProfile`. The seam itself is correct (it delegates to `dialUpstreamWithALPN`, which was fixed), so nothing is broken today — but this is the contract a downstream forward-path implementer reads, and following it literally at a *new* dial site would reintroduce the exact USK-1021 failure. Suggested remediation: s/`EffectiveTLSFingerprint`/`EffectiveUTLSProfile`/ in that comment.

**S-3 — pre-existing fail-open on the resend axis (LOW, not introduced here).**
Flagging for the record only: this PR is behaviour-preserving on `InitTLSTransport` (`UTLSProfileFor(ResolveTLSFingerprint(...))` is exactly the old composition), but it touched the line, so it is worth stating that the asymmetry survives. `ProxyConfig.Validate()` (`config/config.go:871`) does **not** validate `tls_fingerprint`, and `ApplyTLSFingerprintFlag` only whitelist-validates when the CLI flag is non-empty — so a config-file `"tls_fingerprint": "firefx"` is never rejected. It then diverges by axis: the live MITM path fails closed and loudly (`tlslayer` hard-errors every dial), while `InitTLSTransport:756-761` logs a `Warn` and silently returns `StandardTransport`, so the typed-resend tools emit a Go-native ClientHello while `query config` reports `"firefx"`. Operator-triggered only (not reachable from wire traffic), and the live-path hard failure makes the misconfiguration hard to miss in practice, hence LOW. If tightened later, the natural fix is validating `TLSFingerprint` in `ProxyConfig.Validate()` against the same 6-name whitelist the CLI flag and the MCP tool already enforce — out of scope for this PR.

---

PR #98, opened 2026-09-13
fix(mcp): document the §var§ template syntax in the macro tool schema

usk6666:
## Summary

M50 Layer 1 (P0). An MCP user reported that a Coding Agent could not work out how to embed variables in a macro, tried several syntaxes, and abandoned Macro entirely. Root cause: no in-band path existed for an agent to reach the correct answer — the `macro` tool Description contained not a single `§`, and `macroStepInput` / `extractionInput` / `guardInput` carried **zero** `jsonschema` tags, so `override_url` / `override_headers` / `override_body` were exposed as "string, no description".

- **Tool Description** (233 B → 909 B, under the ~1 KB de-facto ceiling): names the `§name§` syntax with the U+00A7 hint, lists the four expansion sites, gives a copyable `override_headers {"Cookie": "PHPSESSID=§session_cookie§"}` example, mentions the encoder pipe chain, and states that `{{name}}` / `${name}` / `%name%` are NOT expanded (plus the post-substitution warning that fires on url / header values / body).
- **Every field** of `macroStepInput` / `extractionInput` / `guardInput` now has a `jsonschema` tag, in the `internal/mcp/fuzz_http.go` prose style. `action` / `params` / `name` / `description` / `steps` / `macro_timeout_ms` were filled in too, so the whole macro input surface is documented.
- **Four, not three, expansion sites.** `override_method` goes through `ExpandTemplate` (`internal/macro/engine.go:381`). The Issue body and `help_macro.md:24` both claimed otherwise; both are corrected here so USK-1036 does not propagate the error across its ~21-site bulk replacement.
- **`initial_vars` / `vars`** now explain how a step references their keys as `§key§`.
- **`help_macro.md`**: fixed the `override_method` line, added `override_method` to the substitution example block, noted that only header *values* (not names) expand and that `extract` / `when` never expand, and documented the encoder chain.
- **`schema_macro.json`**: fixed type drift against the generated schema — `override_body` / `extract` / `when` are nullable because the Go fields are `*string` / slice / `*guardInput`. No `§` documentation added here; USK-1036 owns that file's content and its three-way sync plan.

Control plane only (`internal/mcp/` + two markdown/JSON resources). No data-path file touched.

### Landmine avoided

A `jsonschema` tag must not begin with a `WORD=` token — `jsonschema-go` rejects it (`infer.go:399` `disallowedPrefixRegexp`) and `gomcp.AddTool` **panics at server construction**. This Issue's own acceptance example, `PHPSESSID=§session_cookie§`, is exactly such a token, so it is kept mid-sentence everywhere. Any test that constructs a Server catches a violation.

## Test plan

New `internal/mcp/macro_schema_test.go` (untagged unit test, fast tier — 8 tests / 21 subtests):

- [x] `TestMacroToolSchema_OverrideFieldsDocumentTemplateSyntax` — `tools/list` → `properties.params.properties.steps.items.properties.<field>.description` contains `macro.DelimOpen` for all **four** `override_*` fields (tag-stripping guard)
- [x] `TestMacroToolSchema_VarsDocumentTemplateSyntax` — same for `initial_vars` / `vars`
- [x] `TestMacroToolSchema_EveryFieldDocumented` — every property of root / `params` / `steps[]` / `steps[].extract[]` / `steps[].when` has a non-empty description; a new Go field without a tag fails here
- [x] `TestMacroToolSchema_StepDefaultsMatchEngine` — documented defaults/limits match `macro.DefaultRetryCount` (3), `MaxRetryCount` (10), `DefaultRetryDelayMs` (1000), `DefaultStepTimeoutMs` (60000), `DefaultMacroTimeoutMs` (300000), and the `OnError` constants
- [x] `TestMacroToolDescription_DocumentsTemplateSyntax` — Description contains `§`, the `{{` counter-example, all four `override_*` names, `initial_vars`, and `url_encode`
- [x] `TestSchemaMacroResource_EnumsMatchGoConstants` — `schema/macro` `action` / `on_error` / `extract.from` / `extract.source` enums equal the Go constants
- [x] `TestSchemaMacroResource_TypesMatchGeneratedSchema` — `schema/macro` step field types equal the generated tool schema (this test caught the `override_body` drift)
- [x] `TestHelpMacroResource_DocumentsExpansionSites` — `help/macro` names every `macro.ListEncoders()` entry and documents `§variable§` on `override_method`
- [x] `internal/mcp/macro_tool_test.go` / `macro_handlers_test.go` pass unmodified
- [x] `make lint`, `make build`, `make test` all green (4119 passing tests, exit 0)

### Acceptance criterion

Without reading any MCP resource, `tools/list` alone now yields:

```
override_headers: "header overrides applied on top of the recorded flow's headers; each key
replaces all existing values for that name. VALUES support §name§ template expansion, keys do
not — e.g. {"Cookie": "PHPSESSID=§session_cookie§"}. {{name}} / ${name} / %name% are NOT
expanded. Expanded keys or values containing CR or LF are rejected"
```

Resolves USK-1034
Linear: https://linear.app/usk6666/issue/USK-1034


usk6666 (commented):

Control-plane-only documentation/schema PR; no data-path file touched, `go.mod` / `go.sum` untouched. Every security-relevant claim in the new `jsonschema` tags and the expanded tool Description was independently verified against `internal/macro/` and found accurate, and no tag violates the `WORD=` prefix rule that would panic `gomcp.AddTool` at server construction. One LOW documentation-precision note.

### WORD_EQUALS_CHECK

**36 `jsonschema` tags inspected** across `internal/mcp/macro_handlers.go` (28, all new) and `internal/mcp/macro_tool.go` (8; 2 new, 6 rewritten). **0 violations** — no tag's first whitespace-delimited token contains `=`.

- Verified statically against the exact upstream rule `disallowedPrefixRegexp` (`^[^ \t\n]*=`) at `jsonschema-go@v0.4.2/jsonschema/infer.go:399`, enforced at `:334-336`.
- Near-miss tags cleared: `Source` (`header|body|body_json|status|url`), `OnError` (`abort|skip|retry;`), `Action` (`REQUIRED`).
- The acceptance example `PHPSESSID=§session_cookie§` is correctly kept mid-sentence in both the `override_headers` tag (first token `header`) and the tool `Description` (not a tag, unconstrained).
- Confirmed dynamically: `make test` constructed the MCP server and all 6 new macro schema tests passed — no `tag must not begin with 'WORD='` panic.

### Factual accuracy audit (verified against `internal/macro/`)

| Claim | Source of truth | Verdict |
|---|---|---|
| Four expansion sites: `override_method` / `override_url` / `override_headers` values / `override_body` | `engine.go:381,390,399,418` — the only `ExpandTemplate`/`ExpandHeaders` call sites in the engine | correct (incl. the `override_method` fix) |
| `override_headers`: values expand, keys do not | `template.go:94-104` (`ExpandHeaders` iterates values only) | correct |
| CR/LF rejected on keys **and** values | `engine.go:405-411` checks both `k` and `v` | correct |
| `extract` / `when` never expanded | no expansion call sites | correct |
| unknown `§name§` left literal; unknown encoder fails the step | `template.go:69-72`, `encoder.go:68-71` | correct |
| 9 encoders listed | `macro.ListEncoders()` — exact match | correct |
| `retry_count` default 3 / max 10; `retry_delay_ms` 1000; `timeout_ms` 60000; `macro_timeout_ms` 300000; `steps` max 50 | `types.go` constants; enforced in `validateMacro` / `validateStepErrorPolicy` | correct + enforced |
| regex pattern ≤ 1024 B, input ≤ 1 MiB; `body_match` first 1 MiB | `MaxRegexPatternLen` / `MaxRegexInputSize` / `MaxStepBodySize`, enforced in `matchRegex` / `matchHeaders` / `matchBody` | correct + enforced |
| non-HTTP `flow_id` rejected | `checkMacroStepsProtocol` at both define and run (TOCTOU-closed, USK-877) | correct + enforced |
| `on_error` / `extract.from` / `extract.source` enums | `validateStepErrorPolicy` / `validateExtractRules` | correct + enforced |
| `vars` merged over `initial_vars` | `Engine.Run` (`engine.go:43-50`) | correct |
| `{{}}` / `${}` / `%%` → status `warning` + `warnings[]` | `detector.go` + `engine.go:350-352` | correct |

**No documented-but-unenforced constraint found** — every limit named in a tag has a corresponding server-side check.

Also confirmed while tracing the claims: `macroSendFunc` re-checks target scope **and** runs the SafetyFilter input check on the **post-expansion** URL/body (`macro_handlers.go:601-614`), so advertising `§var§` in `override_url` does not open a scope-bypass path. No new text encourages disabling safety filters or exfiltrating recorded credentials. Examples use only RFC 2606 `example.com` and placeholder variables — no real-looking credential, token, key, internal hostname, or path is introduced.

### FINDINGS

|---|---|---|---|---|---|---|
| S-1 | LOW | `internal/mcp/macro_handlers.go` | 51 | MCP/API | CWE-1078 (Inappropriate Source Code Style/Comment) | `override_headers` description says "each key replaces all existing values for that name", but the match is an exact Go map-key comparison, not case-insensitive |

**S-1 detail.** The recorded template headers come from `storeFlowFetcher.GetFlowRequest` → `flow.Flow.Headers`, which preserves wire casing by design (`keyValuesToMap` in `internal/pipeline/record_step.go:1245` does no canonicalization — MITM Principle #1). HTTP/2 flows therefore record `cookie` (lowercase, RFC 9113). An agent following this description and sending `override_headers: {"Cookie": "PHPSESSID=§session_cookie§"}` does **not** replace the recorded `cookie` entry: `buildRequest` writes a second map key (`engine.go:413`), and `macroSendFunc` then calls `httpReq.Header.Set(...)` for each (`macro_handlers.go:591-598`), which canonicalizes both to `Cookie` — so the winner is decided by Go map iteration order.

*Impact:* an operator can non-deterministically replay the **stale recorded** session cookie instead of the intended override, silently invalidating an auth-chaining macro. Not attacker-triggerable; operator-facing test-fidelity issue only.

*Remediation (documentation only):* add one clause, e.g. "…the key is matched case-sensitively against the recorded header name; HTTP/2-recorded flows use lowercase header names." This is explicitly **not** a request to change `map[string]string` — that lossy-type decision is settled and out of scope.

### Verification

- `make lint` — pass (gofmt, go vet, staticcheck v0.7.0, ineffassign, gocyclo)
- `make build` — pass
- `make test` — all 6 new macro schema tests pass. Two packages (`internal/flow`, `internal/mcp`) hit the 4m per-package `-timeout` alarm locally with **zero assertion failures**; the stacks show slow unrelated sqlite tests (`TestLargeExportImport` 1m55s) under `-race` on a machine running a concurrent `make test` in a sibling worktree. PR CI is green (`test` 3m12s, `test-e2e-smoke` 3m55s), so this is environmental, not a PR defect. Non-blocking note for the author: the 6 new tests each spin up a fresh in-memory MCP server/client pair (~8s added to `internal/mcp`, which runs under a 4m cap); sharing one session across them would reclaim most of that headroom.

### STATS

CRITICAL: 0 · HIGH: 0 · MEDIUM: 0 · LOW: 1

---

usk6666 (commented):
_(Duplicate submission — the previous `gh pr review` call returned a transient GitHub 5xx/GraphQL error but had already been accepted server-side, so the retry created a second copy. See the full security review immediately above; this entry carries no additional content.)_

usk6666 (commented):
## Code Review: CHANGES REQUESTED ❌

This is a well-executed documentation PR: the `§var§` syntax now appears in the tool `Description`, on all **four** expansion sites (including `override_method`, which the Issue body got wrong), and on both KV-Store entry points. I independently fact-checked all 43 documented defaults/limits/enums/behaviours against `internal/macro/` — 41 are exactly right. `make lint` / `make build` are green and all 8 new tests pass locally. The `WORD=` landmine is cleanly avoided (verified mechanically against `disallowedPrefixRegexp` on all 36 tags — zero violations).

One factual error blocks merge: the `flow_id` tag points agents at a JSON path that does not exist in the `query` response.

### Findings

| F-1 | HIGH | internal/mcp/macro_handlers.go | 49 | Correctness | `flow_id` tag says `items[].id`; the `flows` resource returns `flows[]` |
| F-2 | MEDIUM | internal/mcp/macro_handlers.go | 52 | Correctness | `override_headers` "replaces all existing values for that name" is true only on an exact-case key match |
| F-3 | LOW | internal/mcp/macro_handlers.go | 74 | Correctness | `required` says "the macro fails"; the *step* fails and `on_error` then applies |
| F-4 | LOW | internal/mcp/macro_schema_test.go | 206-247 | TestQuality | Drift guard covers 5 of 8 documented numeric constants |
| F-5 | NIT | internal/mcp/macro_schema_test.go | 214-218 | TestQuality | Substring assertions on bare digits (`"3"`, `"10"`) are weak |
| F-6 | NIT | internal/mcp/macro_schema_test.go | 24-76 | CodeHealth | Every helper call spins a fresh in-memory MCP server (~9 s added to a 3m12s CI job with a 4 m cap) |
| F-7 | NIT | internal/mcp/macro_schema_test.go | 114-123, 193-201 | CodeHealth | `sortedSchemaKeys` / `sortedObjectNames` are the same function twice |

---

#### F-1 — HIGH · `flow_id` points at a non-existent response key

```go
StreamID string `json:"flow_id" jsonschema:"... take it from query resource=\"flows\" items[].id ..."`
```

The `flows` resource returns `{"flows": [...], "count": …, "total": …}` — `queryFlowsResult.Flows` has `json:"flows"` (`internal/mcp/query_tool.go:812-815`), and `internal/mcp/resources/help_query.md:88` documents it as `flows[]`. `items[]` is the **`intercept_queue`** shape (`help_query.md:153`). Every e2e harness decodes `json:"flows"` (e.g. `internal/mcptest/resend_smoke_integration_test.go:110`).

This is the same failure mode the PR exists to eliminate — a schema string that sends an agent looking for something that isn't there. The *value* is right (`queryFlowsEntry.ID` is the Stream ID that `GetStream` / `GetFlows` take); only the key name is wrong.

**Suggestion**: `items[].id` → `flows[].id`.

---

#### F-2 — MEDIUM · header override replacement is case-sensitive

```go
jsonschema:"header overrides applied on top of the recorded flow's headers; each key replaces all existing values for that name. ..."
```

`buildRequest` does `req.Headers[k] = []string{v}` (`internal/macro/engine.go:412`) against a map keyed by the **wire-observed** header name. Per MITM Principle #2 an HTTP/2-recorded flow stores lowercase names (RFC 9113), so an override key of `Cookie` does not replace `cookie` — it adds a second map entry. `macroSendFunc` then calls `httpReq.Header.Set(key, v)` per key (`macro_handlers.go:591-599`), which canonicalises both to `Cookie`, so the winner is decided by Go map iteration order — i.e. nondeterministically. The headline example in the tool `Description` is exactly `{"Cookie": …}`, so this is the case agents will hit.

Not asking to change the `map[string]string` type (settled). This is the "document the limitation honestly" half of that decision.

**Suggestion**: "…each key replaces all existing values for that **exact** header name — the match is case-sensitive against the recorded flow's header names, which are lowercase for HTTP/2-recorded flows."

---

#### F-3 — LOW · `required` overstates the blast radius

```go
Required bool `json:"required,omitempty" jsonschema:"when true, the macro fails if extraction finds nothing"`
```

A failed required extraction returns an error from `ExtractValues` (`extract.go:17-19`) → `doStepExecution` (`engine.go:326-330`) → `handleStepFailure` (`engine.go:251-275`), where the step's `on_error` decides. With `on_error: "skip"` the step is skipped and **the macro continues**. Only the default `abort` matches the current wording.

**Suggestion**: "when true, the step fails if extraction finds nothing (the step's `on_error` policy then applies)".

---

#### F-4 — LOW · drift guard is partial

`TestMacroToolSchema_StepDefaultsMatchEngine` pins 5 documented numbers (3 / 10 / 1000 / 60000 / 300000). Three more are now documented in tags with no guard at all:

- `steps` → "maximum 50" vs `macro.MaxSteps`
- `extract.regex`, `when.header_match`, `when.body_match` → "1024 bytes" vs `macro.MaxRegexPatternLen`
- `extract.regex` → "at most 1 MiB", `when.body_match` → "first 1 MiB" vs `macro.MaxRegexInputSize` / `macro.MaxStepBodySize`

`help_macro.md`'s own copies of the defaults (`3` / `1000` / `60000` / `300000`, lines 29-35) are also unguarded, so the file can drift silently even when the Go-tag test fires.

**Suggestion**: add those three constants to the table + the constant-equality block, and extend `TestHelpMacroResource_DocumentsExpansionSites` with the same default strings it already applies to the encoder list.

---

#### F-5 — NIT · weaken-proof the numeric assertions

`{"retry_count", []string{"3", "10"}}` with `strings.Contains` also passes for "default 30, maximum 100". The companion constant-equality check catches engine-side changes, but not a typo on the prose side.

**Suggestion**: assert the exact phrase, or build it from the constant — `"default " + strconv.Itoa(macro.DefaultRetryCount)`.

---

#### F-6 — NIT · one MCP server per assertion

`macroToolSchema` / `macroToolDescription` / `readMacroSchemaResource` each call `setupResourceTestSession`, and `TestMacroToolSchema_StepDefaultsMatchEngine` calls `macroToolSchema` twice. Locally the 8 tests cost ~8.9 s (`StepDefaultsMatchEngine` alone 3.38 s) in a `-race` package whose CI `test` job already runs 3m12s against the `-timeout 4m` cap in `Makefile:30`. The schema is static, so the cost buys nothing.

**Suggestion**: memoise the marshalled schema / description once per package (`sync.OnceValues`), keeping `t` only for `Fatalf`.

---

#### F-7 — NIT · duplicate helper

`sortedSchemaKeys(map[string]any)` and `sortedObjectNames(map[string]map[string]any)` differ only in value type.

**Suggestion**: `func sortedKeys[V any](m map[string]V) []string`.

---

### Acceptance criterion — does it hold?

**Yes.** From `tools/list` alone the agent gets, in the tool `Description`: the codepoint (`U+00A7 SECTION SIGN on BOTH sides`), the four expansion sites by name, the verbatim target example `override_headers {"Cookie": "PHPSESSID=§session_cookie§"}`, the explicit negative list (`{{name}}`, `${name}`, `%name%` are NOT expanded → sent literally), the encoder-chain form, and where values come from. `override_headers`' own property description repeats the example plus the "VALUES support…, keys do not" caveat. An agent can write the target macro first try.

F-1 does not break that criterion — it breaks the *adjacent* step (obtaining a `flow_id`) that the same tag is trying to teach.

### Verification

`make lint` PASS (gofmt / vet / staticcheck / ineffassign / gocyclo) · `make build` PASS · all 8 new tests PASS. Two local packages (`internal/flow`, `internal/mcp`) hit the `-timeout 4m` cap at exactly 240 s under CPU contention from concurrent worktree test runs; CI is green (`test` job 3m12s), so that is a local artefact, not a PR regression.

### FACT_CHECK_TABLE

| Documented item | PR claims | Source | Source ref | |
|---|---|---|---|---|
| `on_error` enum | `abort\|skip\|retry` | same 3 | types.go:47-51 | OK |
| `on_error` default | `abort` | `OnErrorAbort` when empty | engine.go:195-197 | OK |
| `retry_count` default | 3 | `DefaultRetryCount = 3` | types.go:19, engine.go:198-200 | OK |
| `retry_count` maximum | 10 | `MaxRetryCount = 10`, rejected above | types.go:22, engine.go:499-501 | OK |
| "retries" (not attempts) | N retries | `maxAttempts = retryCount + 1` | engine.go:211 | OK |
| `retry_delay_ms` default | 1000 | `DefaultRetryDelayMs = 1000` | types.go:24, engine.go:201-203 | OK |
| `timeout_ms` (step) default | 60000 | `DefaultStepTimeoutMs = 60_000` | types.go:15, engine.go:192-194 | OK |
| `macro_timeout_ms` default | 300000 | `DefaultMacroTimeoutMs = 300_000` | types.go:13, engine.go:82-85 | OK |
| `macro_timeout_ms` actually wired | — | `MacroTimeout` → `TimeoutMs` | macro_tool.go:113 | OK |
| `steps` maximum | 50 | `MaxSteps = 50` | types.go:17, engine.go:453-455 | OK (unguarded, F-4) |
| regex pattern cap | 1024 bytes | `MaxRegexPatternLen` | types.go:34, extract.go:169 | OK |
| regex input cap | at most 1 MiB | `MaxRegexInputSize = 1<<20` | types.go:29, extract.go:178-181 | OK |
| `body_match` input | first 1 MiB | `MaxStepBodySize` + regex cap | engine.go:334-337, guard.go:183-186 | OK |
| `header_match` pattern cap | ≤1024 bytes each | per-pattern check | guard.go:146-149 | OK |
| `extract.from` enum | `request\|response` | same 2 | types.go:74-77 | OK |
| `extract.source` enum | `header\|body\|body_json\|status\|url` | same 5 | types.go:58-67 | OK |
| `source: status` reads response | always response | `From` ignored | extract.go:127-132 | OK |
| `group` 0 = full match | yes | `matches[0]` | extract.go:183-192 | OK |
| `default` ignored when `required` | yes | required returns before default | extract.go:17-22 | OK |
| `required` → "the **macro** fails" | macro fails | **step** fails; `on_error` applies | engine.go:326-330, 251-275 | **WRONG (F-3)** |
| four `§var§` expansion sites | method/url/header values/body | exactly those four | engine.go:379-423 | OK |
| header **keys** not expanded | not expanded | values only | template.go:92-105 | OK |
| CR/LF rejection | keys **and** values | both checked | engine.go:405-411, helpers.go:38-48 | OK |
| header override replacement | "replaces all values for that name" | exact-case key only | engine.go:412 | **WRONG (F-2)** |
| unknown `§var§` left literal | yes | returns `§expr§` | template.go:70-73 | OK |
| unknown encoder fails step | yes | `ApplyEncoders` errors | encoder.go:394-396 | OK |
| encoder chain syntax | `§name \| e1 \| e2§` | split on `\|` + TrimSpace | template.go:60-86 | OK |
| encoder list (help, 9) | url_encode, base64, base64_decode, hex, html_encode, lower, upper, md5, sha256 | identical | encoder.go:372-386 | OK |
| detector scope | url / header values / body | + method NOT scanned | detector.go:235-284 | OK |
| detector → status `warning` + `warnings[]` | yes | `status = "warning"` | engine.go:345-360, types.go:170-185 | OK |
| help: body scanned first 64 KiB | 64 KiB | `MaxUnresolvedScanBytes = 64<<10` | detector.go:202 | OK |
| `when.step` required unless `extracted_var` alone | yes, else error | exactly that | guard.go:44-51 | OK |
| not-run / skipped step never matches | yes | returns false | guard.go:53-61 | OK |
| `status_code_range` `[0,0]` = unset | yes | both-zero short-circuit | guard.go:100-102 | OK |
| guard AND + `negate` | yes | all checkers must pass | guard.go:22-25, 63-86 | OK |
| `extract` / `when` never expanded | yes | only the four sites | engine.go:371-426 | OK |
| `flow_id` source path | `query resource="flows" items[].id` | response key is `flows[]` | query_tool.go:812-815, help_query.md:88 | **WRONG (F-1)** |
| non-HTTP flows rejected | use resend_ws / grpc / raw | identical gate + message | macro_handlers.go:302-354 | OK |
| `name` required for every action | yes | all 3 handlers check | macro_handlers.go:137, 204, 385 | OK |
| `define_macro` upserts | yes | `SaveMacro` + `created` flag | macro_handlers.go:163-194 | OK |
| `vars` merged over `initial_vars` | vars win | initial then vars | engine.go:42-51 | OK |
| KV keys case-sensitive | yes | plain map lookup | template.go:68 | OK |
| `WORD=` prefix rule | all 36 tags | none match `^[^ \t\n]*=` | infer.go:334-336, 399 | OK |

**STATS** — CRITICAL: 0 · HIGH: 1 · MEDIUM: 1 · LOW: 2 · NIT: 3

---

usk6666 (commented):
## Code Review (re-review, round 1): APPROVED ✅

All four LOW-and-above findings from the initial review are resolved in `991c0b6e`. Re-verified each against source; CI is green on the fix commit.

| ID | Severity | Status | Verification |
|----|----------|--------|--------------|
| F-1 | HIGH | **FIXED** | `flow_id` tag now reads `query resource="flows" flows[].id`. Confirmed against `queryFlowsResult.Flows` (`json:"flows"`, `internal/mcp/query_tool.go:811`) and `help_query.md:88`. |
| F-2 / S-1 | MEDIUM / LOW | **FIXED** | Tag now states the key is "matched case-sensitively against the recorded header name (HTTP/2 flows record lowercase names, so use cookie not Cookie there)". Accurate against `buildRequest`'s `req.Headers[k] = []string{v}` over a `map[string][]string` keyed by wire-observed names, and `keyValuesToMap` (`internal/pipeline/record_step.go`), which does no canonicalization per MITM Principle #1. A fuller paragraph was added to `help_macro.md`. The `map[string]string` type is correctly left unchanged. |
| F-3 | LOW | **FIXED** | `required` now reads "the step fails if extraction produces no value; on_error then decides whether the macro aborts" — matches `ExtractValues` → `doStepExecution` → `executeWithRetry` → `handleStepFailure`. `default` reworded from "finds nothing" to "produces no value", which also covers non-empty extraction errors. |
| F-4 | LOW | **FIXED** | `TestMacroToolSchema_StepDefaultsMatchEngine` gained three subtests (`extract.regex`, `when.header_match`, `when.body_match`) plus constant pins for `MaxSteps`, `MaxRegexPatternLen`, `MaxRegexInputSize`, `MaxStepBodySize`. Numeric drift coverage is now 9 of 9 documented constants (3 / 10 / 1000 / 60000 / 300000 / 50 / 1024 / 1 MiB regex input / 1 MiB body). |
| F-5, F-6, F-7 | NIT | Not fixed (by policy) | NIT findings are excluded from the fix scope. F-6 was partially mitigated — a redundant second `macroToolSchema(t)` construction was removed from the defaults test. |

### WORD= prefix rule — re-verified after the edits

The `override_headers` tag grew substantially and two more tags were rewritten, so the rule was re-checked mechanically across all **426** `jsonschema` tags in `internal/mcp/`: **0 violations**. First tokens of the touched tags are `header`, `id`, `fallback`, `when` — none contains `=`. The `PHPSESSID=§session_cookie§` example remains mid-sentence. Dynamically confirmed: `make build` and all 8 macro schema tests construct the MCP server without a `gomcp.AddTool` panic.

### Note on F-6

The per-test in-memory MCP session cost was estimated at ~9 s in the initial review. Measured from the test output it is ~2.3 s across all 8 tests, and the CI `test` job runs 3m29s on the fix commit against a 4 m per-package cap, so the headroom concern is smaller than first assessed. Left as a NIT.

### Verification

`make lint` PASS · `make build` PASS · 8/8 macro schema tests PASS (incl. 3 new subtests). CI green on `991c0b6e`: `test` 3m29s, `test-e2e-smoke` 4m6s, `lint`, `build`, `CodeQL` all pass.

### STATS

CRITICAL: 0 · HIGH: 0 · MEDIUM: 0 · LOW: 0 · NIT: 3

---

usk6666 (commented):
## Code Review (round 2): APPROVED ✅

All four LOW-and-above findings from round 1 are fixed in `991c0b6e`, and I re-verified each one independently against the source rather than taking the commit message's word for it. The three NITs (F-5/F-6/F-7) were deliberately skipped per fixer policy — that is fine; F-6 was in fact partially addressed for free by dropping the redundant second `macroToolSchema(t)` call (`StepDefaultsMatchEngine` 3.38 s → 1.35 s locally).

### Finding resolution

| ID | Severity | Status | Verification |
|----|----------|--------|--------------|
| F-1 | HIGH | ✅ Fixed | `macro_handlers.go:49` now reads `flows[].id`, matching `queryFlowsResult.Flows` `json:"flows"` (`query_tool.go:813`) and `help_query.md:88` |
| F-2 | MEDIUM | ✅ Fixed | Case-sensitivity caveat added to the tag **and** `help_macro.md:138-143`; claim traced end-to-end below |
| F-3 | LOW | ✅ Fixed | `required` → "the step fails … `on_error` then decides whether the macro aborts", matching `engine.go:326-330` → `handleStepFailure` (`engine.go:251-275`) |
| F-4 | LOW | ✅ Fixed | Three new subtests + constant pins for `MaxSteps` / `MaxRegexPatternLen` / `MaxRegexInputSize` / `MaxStepBodySize` |

### Specific confirmations requested

**(a) `WORD=` prefix rule still holds.** Re-ran the mechanical check over all 36 `jsonschema` tags in `macro_handlers.go` + `macro_tool.go` against `disallowedPrefixRegexp` (`^[^ \t\n]*=`, `jsonschema-go@v0.4.2/jsonschema/infer.go:399`): **36 tags checked, 0 violations**. The grown `override_headers` tag still starts with `header overrides applied…`, and both `PHPSESSID=§session_cookie§` and the new parenthetical stay mid-sentence. `make build` (which constructs the Server via `AddTool`) passes, which is the executable proof.

**(b) The case-sensitivity wording is factually correct.** Traced the full path, and the claim holds at every hop — there is no canonicalisation anywhere between the wire and `buildRequest`:

1. `keyValuesToMap` (`internal/pipeline/record_step.go:1245-1253`) keys the map on `kv.Name` **verbatim** — its own doc comment says duplicate-name order is preserved and no name folding happens.
2. `internal/flow/sqlite.go:742-743` / `:1318` round-trip the map through plain `json.Marshal` / `json.Unmarshal`, which preserves key case exactly.
3. `storeFlowFetcher.GetFlowRequest` (`macro_handlers.go:663-667`) assigns `msg.Headers` straight through; `copyHeaders` (`engine.go:429-440`) deep-copies without touching names.
4. `buildRequest` writes `req.Headers[k] = []string{v}` (`engine.go:412`).

I also grepped for `CanonicalMIMEHeaderKey` / `textproto.Canonical` / `http.Header(` across `internal/flow/` and `record_step.go` — zero hits. So on an HTTP/2-recorded flow (lowercase names per RFC 9113 §8.2.1 and MITM Principle #2) an override key of `Cookie` genuinely does add a second entry rather than replacing `cookie`, exactly as the new text says. The `help_macro.md` paragraph says the same thing in the same direction. ✅

One deliberate tension worth noting, not a finding: the inline example is still `{"Cookie": "PHPSESSID=§session_cookie§"}`. That is correct for HTTP/1.1-recorded flows and is the acceptance-criterion string that has to survive verbatim, and the parenthetical immediately scopes the lowercase advice to h2. Keeping it is the right call.

**(c) The F-4 assertions pin what they claim.** Checked each navigation path and each pin:

- `extractProps` = `steps.items.properties.extract.items.properties`, `whenProps` = `steps.items.properties.when.properties` — both resolve (confirmed by the subtests passing, and `when` is `["null","object"]` yet still carries `properties`).
- `extract.regex` pins `"1024"` + `"1 MiB"` against `"Maximum 1024 bytes, matched against at most 1 MiB of input"` → `MaxRegexPatternLen` (`types.go:34`, enforced `extract.go:169`) and `MaxRegexInputSize` (`types.go:29`, enforced `extract.go:178-181`). ✅
- `when.header_match` pins `"1024"` → `matchHeaders` per-pattern check (`guard.go:146-149`). ✅
- `when.body_match` pins `"1024"` + `"1 MiB"` → `matchBody` (`guard.go:176-186`) plus the `MaxStepBodySize` truncation at `engine.go:334-337`. ✅
- `steps` pins `"50"` → `MaxSteps` (`types.go:17`, enforced `engine.go:453-455`). ✅
- The constant-equality block `MaxRegexPatternLen != 1024 || MaxRegexInputSize != 1<<20 || MaxStepBodySize != 1<<20` closes the other direction, so neither the prose nor the constant can move alone. ✅

### Residual, non-blocking

Neither of these blocks merge; recording them so they are not lost:

1. `help_macro.md:29-35` still carries its own copies of `3` / `1000` / `60000` / `300000` with no drift guard — the Go-tag half of F-4 is now covered, the Markdown half is not. `TestHelpMacroResource_DocumentsExpansionSites` is the natural home if it is ever worth closing.
2. `TestMacroToolSchema_StepDefaultsMatchEngine` is now ~76 lines carrying two table loops and four constant-pin blocks. Still under the gocyclo 15 threshold, but a future `_LimitsMatchEngine` split would read better.

### Verification

`make lint` PASS (gofmt / vet / staticcheck / ineffassign / gocyclo) · `make build` PASS · all 8 macro schema tests and all 7 subtests PASS locally with **zero assertion failures** across the whole fast tier. The two local package failures are the `-timeout 4m` wall-clock cap at exactly 240 s on `internal/flow` and `internal/mcp` — reproduced with no competing test processes, so this machine is simply slower than the CI runners. CI on `991c0b6e` is green: `test` 3m29s, `test-e2e-smoke` 4m6s, `lint`, `build` all pass.

**STATS** — CRITICAL: 0 · HIGH: 0 · MEDIUM: 0 · LOW: 0 · NIT: 0 (3 round-1 NITs deliberately unfixed per fixer policy)

---

PR #101, opened 2026-09-13
feat(macro): loud failure for unresolved template variables

usk6666:
## Summary
- **Pre-expansion `§var§` scanner** (`internal/macro/unresolved.go`): `templateVarNames` mirrors `ExpandTemplate`'s `strings.Index` pairing by construction, so `§ name §`, `§name | base64§` and `§a§§b§` agree with the expander while `§unclosed` / `§§` / a lone `§` never warn. `UnresolvedVars(s, kv)` returns the sorted, deduped set of names that will reach the wire literally.
- **`internal/mcp/fuzz_http_helpers.go:collectUnresolvedFuzzTokens` now delegates** to that scanner — the second private copy of the pairing loop is gone.
- **Run time**: `DetectUnresolvedVars(step, kvStore)` runs next to `DetectUnresolvedTemplates` in `doStepExecution`, sets `StepResult.Status = "warning"`, appends a `StepResult.Warnings` line per (location, category), and emits `slog.Warn`.
- **Define time**: `macro.ValidateMacroTemplates(*Macro)` statically scans the four `override_*` fields; `handleDefineMacro` returns the result in the new `macroDefineMacroResult.Warnings`. Never an error — the macro is still saved.
- **Remediation in the text**: foreign-syntax samples render as `{{session}} → §session§ — not expanded; macro templates use the §name§ form only (U+00A7 SECTION SIGN on both sides)`; unresolved `§var§` warnings list the KV Store keys that *are* available (**names only**, capped at 10; tokens capped at the existing 5 + `...(N more)`).
- **Docs**: `help_macro.md` gains an `### Unknown §variable§ names` section appended after `### False positives`, both `Returns:` lines mention `warnings`, and the stale "Unknown variables are left literally on the wire" sentence is amended.

Acceptance criteria, both covered by explicit tests:
1. `override_headers: {"Cookie": "PHPSESSID={{session}}"}` → `define_macro` returns a warning containing the literal `§session§` (`TestExecute_DefineMacro_ForeignSyntaxWarns`).
2. `§sesion_cookie§` → `run_macro` returns a warning listing available variable names and no values (`TestExecute_RunMacro_UnresolvedVarWarns`).

## Design decisions ratified by the user (Phase 1.5)
- Pre-expansion shared scanner instead of a 4th regex in `detector.go` (exact; agrees with `ExpandTemplate` by construction; immune to the 64 KiB window and to hostile response values containing `§x§`)
- Only the four `override_*` fields are scanned — a `§name§` in the recorded base flow is not expanded and is not warned about (documented instead)
- `detector_test.go`'s `TestDetectUnresolvedTemplates_SectionSignNotFlagged` inverted — it asserted the exact behaviour this Issue removes (USK-846 scope statement, not a considered policy). It is now `TestDetectUnresolvedTemplates_SectionSignHandledByVarScanner`: still pins that the foreign-syntax detector stays out of `§`-land (it has neither the KV Store nor the pre-expansion template), and additionally asserts `DetectUnresolvedVars` fires on the same input.

## Test plan
- [x] Residual `§ident§` detected: plain, encoder pipe (`§name | base64§`), encoder chain, inner whitespace (`§ name §`), adjacent pairs (`§a§§b§`)
- [x] **Agreement test** (`TestUnresolvedVars_AgreesWithExpandTemplate`): 23-entry corpus checked against an *independent* `strings.Split`-based reference implementation, plus a liveness check that supplying a reported name changes `ExpandTemplate`'s output (proves no false positives)
- [x] False positives that must not fire: bare `§`, `§123`, `§unclosed`, `§§`, whitespace-only interior, resolved variables, and `§x§` inside a KV Store **value**
- [x] **Security regression**: sentinel KV value asserted absent from every warning on both the define-time and run-time paths (`TestDetectUnresolvedVars_NeverLeaksKVValues`, `TestEngine_Run_UnresolvedVarWarning`, `TestExecute_RunMacro_UnresolvedVarWarns`)
- [x] `define_macro` with foreign syntax / unresolved vars returns `warnings[]` and does **not** error
- [x] A variable warned about at define time and supplied via `run_macro`'s `params.vars` executes successfully with the value on the wire (`TestExecute_DefineMacro_WarnedVarStillRunnableViaRunVars`)
- [x] Values resolved from a *preceding* step's `extract` produce no warning; own-step and forward references do
- [x] Determinism: 20 repeated runs byte-identical on both paths (guard against map iteration)
- [x] MITM Principle #1: `TestDetectUnresolvedVars_DoesNotMutateStep` + wire-copy assertions — the literal token is sent verbatim, and no "best match" is written into the KV Store
- [x] Every other test in `detector_test.go` stays green
- [x] `make lint`, `make build`, `make test` all pass (4136 tests, exit 0); 100% statement coverage on every new function, `internal/macro` package at 94.5%

## Note for reviewers
Runs in parallel with PR #98 (USK-1034). Overlapping files are `internal/mcp/macro_handlers.go`
(distinct hunks) and `internal/mcp/resources/help_macro.md` (this PR appends a new section after
"False positives"; #98 inserts before "Foreign syntaxes"). Whichever merges second rebases.
`internal/mcp/macro_tool.go` and `internal/mcp/resources/schema_macro.json` are untouched.

Resolves USK-1035
Linear: https://linear.app/usk6666/issue/USK-1035


usk6666 (commented):

The new `§var§` scanner is regex-free, bounds-safe on arbitrary bytes, read-only with respect to both the `Step` and the KV Store, and deterministic; the KEY-NAMES-ONLY invariant holds on every new path, so the server-log boundary (no KV Store values in logs) is preserved. Two LOW observations about the *availability hint* and the 64 KiB body scan window are recorded below for a follow-up; neither is exploitable enough to block.

### Verified claims (checked against source, not taken on trust)

| Claim | Verdict |
|---|---|
| No KV Store **value** can reach a warning string or `slog` attribute | **TRUE.** `availableVarsHint` (`unresolved.go:282-292`) iterates `for name := range kvStore` and appends only `name`. `formatVarTokens` renders operator-authored template names. `DetectUnresolvedVars`'s `Sprintf` interpolates only `loc.label` / tokens / two constants. The `slog.WarnContext` at `engine.go:319-324` passes `macroName`, `step.ID`, and that same string. No `%v`/`%+v` on the map anywhere. No KV Store key embeds its own value (`hooks.go:426` builds `__response_headers<name>__` from the header **name**; the value goes to the map value). |
| Sentinel regression test genuinely covers **both** paths | **TRUE.** `TestDetectUnresolvedVars_NeverLeaksKVValues` (`unresolved_test.go:295`) asserts the sentinel is absent from `DetectUnresolvedVars` output **and** from `ValidateMacroTemplates(&Macro{InitialVars: kv})`. The sentinel is unique and appears only as a value, so it would be caught if either path started echoing values. `TestExecute_RunMacro_UnresolvedVarWarns` repeats the assertion through the real MCP boundary. |
| **Zero new regexes** | **TRUE.** The diff on `internal/macro/detector.go` shows the `var(...)` block byte-identical; grepping `regexp.MustCompile` under `internal/macro/` returns exactly the 3 pre-existing patterns, all with the bounded `[A-Za-z_][A-Za-z0-9_]{0,63}` interior. RE2 → no backtracking, no ReDoS. |
| `FindAllString` → `FindAllStringSubmatch` cannot panic on `m[1]` | **TRUE.** Each of the three patterns has exactly one capture group, so every submatch slice has `len == 2`. A non-participating group would yield `""`, not a panic. |
| `macroRunMacroResult.KVStore` already returns values **unmasked** | **TRUE** — this was the premise for skipping the Output Filter, and it holds. `buildRunMacroResult` (`macro_handlers.go:375`) assigns `KVStore: result.KVStore` verbatim; grepping `filterOutput` in `macro_handlers.go` / `macro_tool.go` returns no hits, and there is no generic filter in the tool dispatch. Key names are therefore a strict subset of existing disclosure, and the no-filter decision stands. |

### Bounds / panic audit (MITM Principle #5)

`templateVarNames` (`unresolved.go:63-86`) does **no rune decoding** — it is pure `strings.Index` on the 2-byte UTF-8 `§`, so there is no byte-vs-rune index mixing to get wrong. Every slice expression is provably in range: `remaining[openIdx+2:]` (an `Index` hit guarantees ≥ 2 bytes remain) and `after[closeIdx+2:]` (same). Hand-checked degenerate inputs — `"§"`, `"abc§"`, `"§§"`, `"§  §"`, a bare `0xC2`, `"\xa7\xc2\xa7"`, `"\xc2\xc2\xa7"` — all terminate without panic. `templateVarName` uses `SplitN(..., 2)[0]`, which always yields ≥ 1 element. `joinSamples`' `samples[:limit]` is guarded by `total > limit` and both call sites pass positive constants; `strings.Join(nil, …)` is safe on the empty path. The walk advances monotonically, so it is **O(n)**, not quadratic.

### MITM Principle #1 (no mutation of operator bytes)

Confirmed read-only end to end: `stepTemplateLocations` copies field values into a local struct, nothing writes back to `Step` or to `kvStore`, and the `→ §ident§` rendering exists only in warning *text*. `DetectUnresolvedVars` is called **after** `buildRequest`, so it cannot influence the request being built. Pinned by `TestDetectUnresolvedVars_DoesNotMutateStep`, which additionally asserts no "best match" is written into the KV Store.

### Other checks

- **Determinism** — no map iteration reaches output: `sortedUnresolved`, `stepTemplateLocations` (header names), and `availableVarsHint` all `sort.Strings`. Pinned by two 20-iteration tests.
- **Injection into the warning channel** — a template token or key name may carry raw CR/LF/ANSI, but both sinks escape: `slog`'s Text and JSON handlers (`internal/logging/logging.go:57-59`) quote any value containing control characters or spaces (every warning contains spaces, so quoting always applies), and the MCP response is JSON-encoded. Not a finding.
- **Concurrency** — no new shared state. The fuzz variant loop is sequential (no goroutines in `fuzz_http_helpers.go`), and `DetectUnresolvedVars` reads `kvStore` on the goroutine that owns it, the same one `ExpandTemplate` already used.
- **Control-flow / availability** — `Result.Status` is still unconditionally `"completed"` (`engine.go:144`); `StepResult.Status == "warning"` is consumed nowhere outside the MCP response, so an attacker who flips a step to `"warning"` (by omitting a referenced response header) cannot abort a macro or a fuzz job.
- **`collectUnresolvedFuzzTokens` delegation** — semantically identical to the deleted private loop (same pairing, same `TrimSpace(SplitN(…,"|",2)[0])`, same `!= ""` guard, same set dedup). `strings` is still used at `fuzz_http_helpers.go:506`, so the import is live.
- **Dependencies** — the diff on `go.mod` / `go.sum` is empty. No new dependencies, no license review needed.
- The expected rebase conflict with PR #98 was not treated as a finding, per the ratified design notes.

### Security Findings (LOW — informational, not merge-blocking)

| S-1 | LOW | internal/macro/unresolved.go | 282-292 | MCP/API | CWE-779 (Logging of Excessive Data) | `availableVarsHint` emits attacker-influenced KV key names with no per-name length bound and no reserved-key exclusion |
| S-2 | LOW | internal/macro/unresolved.go | 203-218, 242-266 | InputValidation | CWE-770 (Allocation Without Limits) | The 64 KiB body scan window is not carried over to the two new `override_body` scan paths |

#### S-1 — attacker-influenced key names in the availability hint

In the fuzz `post_macro` path with `pass_response: true`, `injectResponseVars` (`internal/mcp/hooks.go:426`) writes keys of the form `__response_headers<lowercased upstream header name>__`. Header **values** are truncated to 8 KiB and the count is capped at 256, but the **name** is neither length-capped nor filtered, and it comes straight off the wire (`fetchFuzzVariantResponse` returns the recorded flow's headers; `maxHeaderSize` is 4 MiB, `internal/layer/http1/parser/parser.go:34`). `availableVarsHint` sorts every kvStore key and renders the first 10 verbatim into the warning, which `engine.go:319` writes to `slog.Warn`.

Two consequences:

1. **The hint becomes useless exactly where it is needed.** `_` (0x5F) sorts before every lowercase letter, so 256 `__response_headers…` keys deterministically fill the entire 10-name list and none of the operator's real names (`session_cookie`, `csrf_token`) are ever shown. The upstream chooses the header names, hence the sort order.
2. **Log-volume amplification.** Up to ~10 × (attacker-chosen name length) bytes per warning location, per step, per fuzz iteration.

Notably this does **not** require operator error to trigger: referencing `§__response_headers_set_cookie__§` in a `post_macro` and the upstream simply *not* sending that header is enough to fire the warning. Severity is held at LOW because there is no value disclosure (the header value is not part of the key), no log injection (slog quotes), the MCP-response side is unaffected (`executePostMacro` discards the hook's `StepResult`, and a direct `run_macro` has a clean kvStore), and the amplification factor over attacker bandwidth is only ~10x.

**Remediation:** in `availableVarsHint`, skip `IsReservedKey(name)` entries — they are never a valid suggestion for a typo'd operator variable, and `ValidateMacroTemplates` already treats them as always-resolved — or sort them last; and truncate each rendered name to a small fixed budget (e.g. 64 bytes) before `joinSamples`. Note `TestDetectUnresolvedVars_TypoWarningShape` currently asserts `__nonce` is present, so that test would need to move with the decision.

#### S-2 — 64 KiB body window not carried to the new scan paths

The existing window is intact: `DetectUnresolvedTemplates` still truncates to `MaxUnresolvedScanBytes` before calling `scanForResiduals` (`detector.go:99-107`). But the invariant it encodes is not carried over to the two new call sites — `ValidateMacroTemplates` calls `scanForResiduals(loc.value)` and `templateVarNames(loc.value)` on the full `*step.OverrideBody`, and run-time `DetectUnresolvedVars` → `UnresolvedVars` does the same. `scanForResiduals` accumulates **every** distinct match into `samples` with one `fmt.Sprintf` allocation each *before* `joinSamples` truncates to 5, so an `override_body` of distinct `{{aN}}` tokens allocates transient strings proportional to body size; `validateMacro` bounds `MaxSteps` but imposes no per-field size limit.

This is LOW rather than MEDIUM because the input is MCP control plane (operator/agent), not wire traffic — it is self-inflicted, the cost is linear (RE2, monotonic walk), and the same body is already JSON-parsed and persisted to SQLite.

**Remediation:** truncate the `body` location to `MaxUnresolvedScanBytes` in `stepTemplateLocations`, and/or stop appending in `scanForResiduals` / `sortedUnresolved` once `limit` is exceeded while continuing to count for the `...(N more)` suffix.

### STATS

```
CRITICAL: 0
HIGH:     0
MEDIUM:   0
LOW:      2
```

---

usk6666 (commented):
## Code Review: APPROVED ✅

The load-bearing `templateVarNames` ⟺ `ExpandTemplate` contract holds: I read both loops line by line and additionally ran a 300k-case differential harness over an adversarial alphabet (`§`, lone `0xC2`, lone `0xA7`, `|`, CR, LF, spaces, `{`/`}`/`$`/`%`) — **0 false positives, 0 under-detections**. Determinism is clean (every new `map` range is followed by a `sort`), the fuzz-path delegation is set-identical to the deleted private loop, and both PR-body claims I was asked to verify are factually correct against source. Findings are LOW/NIT only.

### Verification notes (claims checked against source, not taken on trust)

| Claim | Verdict |
|---|---|
| go-sdk publishes no output schema, so `Warnings` needs no `schema_macro.json` change and still reaches the agent | **TRUE.** `registerMacro` passes `Out = any` and a `Tool` literal with no `OutputSchema`, so `toolForErr`'s `if t.OutputSchema != nil \|\| reflect.TypeFor[Out]() != reflect.TypeFor[any]()` is false → `outputResolved` stays nil → `applySchema` is a no-op, and `res.Content == nil` → the SDK auto-fills the `TextContent` block (go-sdk v1.4.1 `mcp/server.go:307-390`). `schema_macro.json` is input-only (`"title": "macro input"`). Empirically confirmed: `unmarshalExecuteResult` reads `Content[0].(*TextContent)` and the new tests pass through the real in-memory MCP transport. |
| `macroRunMacroResult.KVStore` already returns KV **values** unmasked | **TRUE.** `macro_handlers.go:103` + `:375` assign `result.KVStore` straight through; no `filterOutput*` call exists anywhere on the macro tool path (all Output Filter call sites are in `query_tool.go` / `resend_*_helpers.go` / `manage_tool.go`). The "names ⊂ existing disclosure" rationale is sound. |
| Scope discipline: `macro_tool.go` / `schema_macro.json` untouched | **TRUE.** `git diff --name-only main...HEAD` confirms 12 files, neither present. |
| `detector_test.go` inversion is honest | **TRUE.** The rewrite keeps the original assertion (`DetectUnresolvedTemplates` still returns nothing for `§…§`) and *adds* a `DetectUnresolvedVars` len-3 pin. No coverage deleted. No other test in the repo asserts that an unresolved `§var§` is silent — the `internal/mcp/fuzz_*_macro_integration_test.go` hits assert the literal reaches the **wire**, which this PR preserves. |
| No second copy of the pairing loop survives | **TRUE.** `DelimOpen`/`DelimClose` now appear only in `template.go` (authority), `unresolved.go` (mirror), `detector.go` (rendering) and the reference impl in `unresolved_test.go`. `strings` is still used in `fuzz_http_helpers.go:506`, so no orphaned import. |
| 23-entry corpus, caps of 5 / 10, `maxSamplesPerLocation = 5` | **TRUE**, counted and read. Every new function is exercised by at least one test; the coverage percentages themselves I did not independently re-measure. |

Adversarial cases traced by hand and confirmed to agree with the expander: `§§`, `§  §`, `§ name §`, `§name | base64§`, `§name | §`, `§a§§b§`, `§a§b§c§`, `§unclosed`, trailing lone `§`, `§x§` inside a substituted KV value, CRLF inside the expression, byte-vs-rune indexing (`len(DelimOpen) == 2`, both sides byte-indexed via `strings.Index`, so a lone `0xA7` can never pair). The one benign divergence — `§§§foo§` makes the scanner report `foo` while `ExpandTemplate` hard-errors — is documented on `templateVarNames` and fails loudly either way.

### Findings

| F-1 | LOW | internal/macro/unresolved.go | 204, 207, 262 | CodeHealth | New body scans bypass the documented `MaxUnresolvedScanBytes` (64 KiB) bound |
| F-2 | LOW | internal/macro/unresolved.go | 282-292 | Architecture | `availableVarsHint` picks its 10 names by plain alphabetical sort; reserved `__*` keys crowd out real names in the fuzz post-macro path |
| F-3 | LOW | web/src/lib/mcp/types.ts | 1342-1356 | Architecture | Bundled WebUI silently drops the new `warnings[]` |
| F-4 | NIT | internal/macro/unresolved.go | 163-164 | CodeHealth | `availableVarsHint(kvStore)` recomputed and repeated verbatim on every warning line of a step |
| F-5 | NIT | internal/mcp/resources/help_macro.md | 216-220 | Correctness | Two "Scope and limits" bullets are true only of the run-time half |
| F-6 | NIT | internal/macro/unresolved.go | 242-266 | Correctness | A `§var§` in an `override_headers` **key** is never expanded and never warned |
| F-7 | NIT | internal/macro/unresolved.go | 34-36 | GoConventions | `templateVarName` lives in the mirror, not in the authority |

### Suggestions

**F-1** — `MaxUnresolvedScanBytes` is documented as bounding "the body region scanned … to keep detector cost bounded for large bodies" (`detector.go:10-14`), and `help_macro.md` states "Body content is scanned only within the first 64 KiB to bound CPU cost" immediately above the new section. Both new body paths — `ValidateMacroTemplates` → `scanForResiduals(*step.OverrideBody)` and `DetectUnresolvedVars` → `UnresolvedVars(*step.OverrideBody, …)` — pass the full body with no truncation, and `validateMacro`/`validateStep` impose no cap on `override_body` length. The run-time path additionally re-runs per retry attempt. Mitigating: the scanned bytes are always the *pre-expansion, operator-authored* template (never a KV value inflated from a hostile upstream response), RE2 is linear, and the single-pass `§` walk is cheap — so this is cost consistency, not an exploitable amplification. Given this codebase's explicit CWE-770 discipline elsewhere (`maxResponseBodyKVBytes` / `maxResponseHeaderValueKVBytes` / `maxResponseHeaderKVCount` in `hooks.go`), suggest either truncating to `MaxUnresolvedScanBytes` in `stepTemplateLocations` for the `body` location, or amending the `MaxUnresolvedScanBytes` godoc + the `help_macro.md` sentence to scope the bound to the foreign-syntax detector only.

**F-2** — In the fuzz post-macro path (`hooks.go:588` `engine.RunWithKV` with the response projection), the kvStore carries `__iteration`, `__nonce`, `__response_status`, `__response_body` and up to `maxResponseHeaderKVCount` (256) `__response_headers__<lower(name)>__` entries. `_` is `0x5F`, so `__*` sorts ahead of every lowercase operator variable — all 10 hint slots are consumed by reserved projection keys and `session_cookie` / `csrf_token` land behind `...(N more)`. The hint then carries zero actionable information in exactly the path where the KV Store is largest, which is the acceptance criterion (b) payload. `macro.IsReservedKey` is already used two functions up; suggest partitioning so non-reserved names fill the budget first (e.g. sort non-reserved, then reserved, then `joinSamples`), or dropping reserved keys from the hint entirely. Note this is about *selection order*, not the ratified names-only / no-Output-Filter decision, which I agree with and am not relitigating.

**F-3** — `MacroDetailPage.tsx:456` calls `define_macro` through the MCP tool and types the response as `MacroDefineResult`, which has no `warnings` field, so the handler shows only `Macro "<name>" saved`. `MacroStepResult` likewise has no `warnings` (pre-existing since USK-846, so the run-time half was already invisible — this PR does not regress it). An operator driving the bundled UI gets exactly the silence the Issue exists to remove. Out of the Issue's stated MCP-agent scope, so a follow-up rather than a blocker: add `warnings?: string[]` to both interfaces and surface a warning toast on save plus a per-step list in the run result.

**F-4** — `DetectUnresolvedVars` calls `availableVarsHint(kvStore)` once per warning location (up to one per override header), re-sorting the whole key set each time and repeating the identical `available: …` list verbatim on every line. For a step with several unresolved headers that is pure duplicated tokens in a product whose stated principle is AI-agent token efficiency. Suggest hoisting `hint := availableVarsHint(kvStore)` above the loop and, ideally, emitting it once per step rather than once per location.

**F-5** — "A step skipped by its `when` guard is never scanned" is true at run time (`executeStep` returns before `doStepExecution`) but false at define time — `ValidateMacroTemplates` walks every step including guarded ones. Likewise "this check has no false positives: a name is reported if and only if the expansion engine would have left it in place" is exact for the run-time check, but the define-time check deliberately reports names that `params.vars` supplies later — its own remediation string says "ignore this warning if that is the plan". Suggest scoping both bullets to the run-time path.

**F-6** — `buildRequest` writes `req.Headers[k] = []string{v}` with the **raw** key, so `override_headers: {"§hdr§": "v"}` puts `§hdr§` on the wire as a header name with no signal from either detector. This matches `DetectUnresolvedTemplates`' documented scope ("names are NOT scanned") and `stepTemplateLocations`' godoc is accurate about it, so it is consistent rather than wrong — but it is the same silent-send class the Issue targets. Cheap fix if wanted: also run `UnresolvedVars` over the header *name* and label it `header-name:<k>`. Otherwise worth one sentence in the `help_macro.md` scope list.

**F-7** — `templateVarName` is the primitive `expandExpression` (`template.go:64`) depends on, but it lives in `unresolved.go`. Since the BINDING CONTRACT comment designates `ExpandTemplate` as the authority and `templateVarNames` as the mirror, hosting the shared definition in the mirror inverts that direction. Suggest moving it next to `expandExpression` in `template.go` and keeping the cross-reference comment in `unresolved.go`.

---

PR #105, opened 2026-09-13
ci(scorecard): add OpenSSF Scorecard workflow + Scorecard/npm README badges

usk6666:
## Summary

Fills the badge slot vacated by the sunset Go Report Card badge (removed in #100, now merged) with two badges that carry real information, plus the workflow that produces one of them.

| Badge | Why |
|---|---|
| **OpenSSF Scorecard** | Grades supply-chain hygiene — pinned dependencies, branch protection, signed releases, dangerous workflow patterns, token permissions. For a MITM proxy this is a more meaningful public signal than a code-style grade. |
| **npm version** | `npm i -g @usk6666/yorishiro-proxy` is the README's primary install path, so "which version would I get" is the question a reader actually has. Verified rendering: `npm: v0.18.2`, matching the latest release. |

The repo already satisfies most of what Scorecard checks — every action is SHA-pinned, the weekly Dependabot security-update workflow runs, releases are tag-driven with npm OIDC Trusted Publishing, and `SECURITY.md` exists — so the initial score should be reasonable rather than embarrassing.

## Workflow design notes

- **`publish_results: true`** is what feeds the badge. It requires `id-token: write` so the OpenSSF API can verify via OIDC that results really came from a run in this repository.
- **Not a `pull_request` workflow.** A fork PR must not be able to publish a score for this repo. Results publish only from `main` (schedule + push) plus manual `workflow_dispatch`.
- **Branch-Protection will score as inconclusive.** Scoring it requires a PAT with `administration: read` held as a long-lived CI secret. That is a worse supply-chain trade than one unscored check, so it is deliberately left unscored — documented in the workflow header so the low sub-score isn't mistaken for a real gap.
- **Cron `41 5 * * 2`** (Tuesdays 05:41 UTC) is clear of `nightly-e2e` (17:00 daily) and `dependabot-security` (Mondays 06:17).
- **Annotated-tag gotcha:** `ossf/scorecard-action` and `github/codeql-action` publish annotated tags, so `git/ref/tags/<tag>` returns the *tag object* SHA. Actions requires a *commit* SHA, so the pins here are the dereferenced commit SHAs. Getting this wrong produces a workflow that fails to resolve the action.

## Known: the Scorecard badge is blank until this merges

`api.scorecard.dev` currently returns `openssf scorecard: invalid repo path` for this repo, because no results have been published yet. The `push: branches: [main]` trigger means the first run happens on merge, and the badge populates from that run. This is expected, not a misconfiguration.

## Test plan

- [x] `make lint` — gofmt / go vet / staticcheck / ineffassign / gocyclo all pass (exit 0)
- [x] `scorecard.yml` parses as valid YAML
- [x] All four action pins verified as real commit SHAs via the GitHub API (annotated tags dereferenced)
- [x] npm badge endpoint verified live: renders `npm: v0.18.2`
- [x] Scorecard badge endpoint reachable; returns the pre-publication placeholder as expected
- [ ] Post-merge: confirm the `OpenSSF Scorecard` workflow run succeeds and the badge resolves to a score

## Related

- Follow-up Issue for the third proposal (golangci-lint consolidation): USK-1043

https://claude.ai/code/session_017CwQqNRansWdoLPNwtPS3f


usk6666:
### Note for review: interaction with CodeQL default setup

This repo runs CodeQL via **default setup**, not a checked-in workflow — the `Analyze (go / actions / javascript-typescript)` jobs on this PR come from `dynamic/github-code-scanning/codeql`, and there is no `codeql.yml` under `.github/workflows/`.

That's worth naming explicitly because default setup *does* reject SARIF uploads — but the documented restriction is specific to SARIF produced by a **CodeQL advanced-setup workflow**. Scorecard's SARIF is third-party (`tool.driver.name: "Scorecard"`), which is a different case, and the upstream OpenSSF template pairs `upload-sarif` with default setup on a very large number of repos.

I could not verify this directly: reading the default-setup configuration needs `code-scanning:read`, and this token gets `403 Resource not accessible by integration`.

Deliberately **not** adding `continue-on-error: true` to the upload step as a precaution. That would be designing around a failure mode whose preconditions I have not confirmed (CLAUDE.md MITM principle 6), and it would mask a genuine rejection rather than surface it. If the upload step does turn out to be rejected after merge, the fix is one line, and the badge itself is unaffected either way — `publish_results` is what feeds it and is independent of the SARIF upload.

The post-merge checkbox in the PR description covers verifying both.

PR #106, opened 2026-09-13
feat(mcp): docs tool for in-band documentation delivery

usk6666:
## Summary

- **New `docs` MCP tool** (the 19th) — `docs()` returns a topic index, `docs(topic="X")` the full document, `docs(topic="X", section="Y")` one heading-anchored section. The topic registry is derived from the existing `helpResources` table via a new explicit `topic` field, so there is no second copy of the corpus and `registerResources()` is untouched (the `yorishiro://help/*` resource URIs keep working).
- **Exported surface for USK-1037** (CLI): `DocsTopic`, `DocsIndex()`, `DocsDocument(topic)`, `DocsSection(topic, section)` in `internal/mcp`. The tool handler is a thin wrapper. No new package.
- **Replaced every `yorishiro://help/` pointer** with `call docs(topic="X")` — 21 sites in `internal/mcp/*.go` (18 tool Descriptions + 2 Go comments + 1 `jsonschema` tag), 1 in `cmd/yorishiro-proxy/client.go`, and **10** cross-links inside the help documents themselves. A URI an agent cannot fetch is the dead end this change removes.
- **Three new documents**: `help_getting-started.md` (end-to-end first session), `help_docs.md` (the docs tool itself), `help_template-syntax.md` (pointer-only `§var§` reference).
- **CLI tables updated**: `clientToolList` (docs leads — it is the discovery entrance), `clientToolDescriptions`, `clientToolHelp`, and `positionalArgMapping["docs"] = {"topic","section"}` so `yorishiro-proxy client docs macro "Variable substitution syntax"` works with bare words.
- **Drive-by fix**: `help_fuzz_ws.md` carried a raw NUL byte inside a JSON example. Replaced with the JSON escape and guarded by `TestHelpResources_ValidUTF8NoControlBytes`.

## Design decisions carried from the Phase 1.5 design review

- **TextContent, not structured output.** `Out = any` + `out = nil` provably skips the go-sdk's `StructuredContent` branch (`mcp/server.go`: `elemZero` stays nil when `Out == any` and no `OutputSchema` is set). Returning a struct would ship a 20 KB document **twice** — once JSON-escaped into `StructuredContent`, once duplicated into a `TextContent` block by the SDK. Pinned by `TestDocsTool_ReturnsSingleTextContent`.
- **Output SafetyFilter deliberately NOT applied**, with a code comment on `handleDocs` saying so and why. MITM principle #7 scopes the Output Filter to *recorded wire data*; these are static in-repo documents. Applying it would mask the sample card numbers and e-mail addresses in `help_security.md` — corrupting the documentation that explains the masking. The comment exists so a future security review does not "fix" this.
- **No output size cap.** The corpus is compile-time-bounded; the largest document is ~5k est tokens, far below the project's own 256 KiB oversize threshold. Truncating a reference mid-sentence is worse agent behaviour than the token cost. `section=` and the index's `est_tokens` are the mitigation.
- **Help-only topic namespace.** The 17 `schema_*` resources duplicate the `InputSchema` that `tools/list` already delivers in full, so including them would cost index tokens and force a namespace disambiguator for zero new information. `topic` is populated on `helpResources` only; a test asserts `topic == ""` for every `schemaResources` entry.
- **H2-only index, full outline on the error path.** Measured: index ≈ 1.6k tok; an H2+H3 index would be ≈ 4.2k — more than reading the largest document. An unknown `section` instead returns that *one* topic's H2–H4 outline (~450 tok worst case, for `configure`).
- **Fence-aware heading scanner.** A `#` inside a ``` or `~~~` block is neither promoted to a section nor able to truncate one. Handles nested fences (a ``` inside a ````-fence does not close it) and `#tag` / `#!/bin/sh`.
- **`est_tokens` is `len/4`**, computed once behind `sync.OnceValue`, and labelled a heuristic in the index header. No tokenizer dependency — the approved-dependency list is closed.

## Corrections to the design review (code was right, prompt was wrong)

- The design review said *"no test asserts that any tool Description contains `yorishiro://help/`"*. **False** — `proxy_start_tool_test.go` `TestProxyStartTool_DescriptionGuidesToConfigure` has a `phrase: "yorishiro://help/proxy_start"` case. Updated to assert `docs(topic="proxy_start")` instead.
- The in-document cross-link count was **10 across 7 files**, not 9 across 6. The 10th (`help_fuzz_ws.md`) was invisible to `grep` because the NUL byte made the file binary — fixing the NUL surfaced it.

## Deferred

- **D1**: exhaustive per-protocol reserved-key table (16 protocol-divergent `__response_*` keys across `fuzz_http`/`fuzz_ws`/`fuzz_grpc`/`fuzz_raw`) — **re-open after USK-1035 merges**. Not low-cost, and it would immediately drift against USK-1035's unresolved-variable semantics. The pointer-only `template-syntax` topic closes the discoverability gap without the drift surface.

## Test plan

- [x] `make lint`
- [x] `make build`
- [x] `make test`
- [x] `make test-e2e-smoke` (incl. `TestRegression_ClientToolList_MatchesServer`, the CLI merge gate)

New tests (`docs_tool_test.go`, 21 top-level + table rows; `resources_test.go` +3):
`tools/list` contains `docs` and its Description names every topic · index enumerates all topics with description + est_tokens + H2 sections · index omits H3 · full document is byte-identical to the embed · **acceptance criterion**: `docs(topic="macro", section="Variable substitution syntax")` reaches `§name§` with its H3 children · slug / case-insensitive / whitespace matching · H3 + H4 addressable · section stops at the next same-level heading · duplicate heading returns the first with a counted note · unknown topic lists topics · unknown section lists the H2–H4 outline · section without topic errors · single `TextContent`, no `StructuredContent` · `DocsIndex()` returns independent copies · fence-aware scanner (9 cases) · slugify (8 cases) · no `yorishiro://help/` in any tool Description or InputSchema · bidirectional embed-FS ↔ registry consistency · `topic` populated on help / empty on schema, and URI + filename agree with it · valid UTF-8, no control bytes.

Resolves USK-1036
Linear: https://linear.app/usk6666/issue/USK-1036/featmcp-docs-tool-for-in-band-documentation-delivery


usk6666 (commented):

Entirely control-plane change (`internal/mcp/` + `cmd/`) serving a compile-time `//go:embed` corpus. The two primary risks for a content-selecting tool — path traversal and output injection — are both structurally closed: topic resolution is an exact map lookup that never reaches a filesystem path, and every error site escapes user input with `%q`. Zero new dependencies, no `unsafe`, no randomness, no goroutines. One non-exploitable LOW on redundant per-heading work in the section matcher.

### Verification of the questions raised

**1. Path traversal — not reachable.** `DocsDocument` resolves via `reg.files[strings.TrimSpace(topic)]`, a Go map lookup keyed on the exact string. A miss errors out; a hit yields a `filename` that is a compile-time string literal from `helpResources` (`resources/help_X.md`). `resourcesFS.ReadFile` never receives caller-supplied data — no `filepath.Join`, no concatenation. `../`, absolute paths, NUL bytes and `%2e%2e%2f` all simply miss the map. `resourcesFS` is an `embed.FS`, so there is no host-filesystem reach even in principle.

**2. Error reflection — mitigated.** All four error sites (`unknownDocsTopicError`, the `DocsDocument` read-error wrap, `docsBody`'s section-without-topic, `extractDocsSection`'s unknown-section) format caller input with `%q`, which escapes newline, carriage return, quote, backslash and non-printables. That defeats the newline/quote-delimiter shapes used for output injection. The success path echoes no caller input at all. The only other value in an error string is the embed-internal path `resources/help_X.md` (public in this repo, and on an unreachable branch since `embed.FS` is immutable and `buildDocsRegistry` already read the file). Caller is the agent itself over a loopback + bearer-token endpoint — self-injection at worst. No finding.

**3. Unbounded output — bounded in practice.** The corpus is 22 `.md` + 17 `.json`, embedded at compile time and not extendable at runtime; the largest document is 19,844 bytes. Index output is bounded by registry size; section output is a substring of one document. Repeated `section` requests return the same bounded slice with no accumulation or per-request caching. The no-cap decision holds.

**4. Section slicer complexity — linear, no regex, no recursion.** `docs_section.go` imports only `fmt`/`strings`/`unicode`, so catastrophic backtracking is not possible. `scanDocsHeadings` is a single pass over lines with inner loops bounded by line length — O(total bytes). No recursion anywhere. Slice bounds are all guarded: `trimmed[level]` behind `level < len(trimmed)`, `rest[0]` behind `rest != ""`, `line[0]` behind the `len(line) < 3` check, `text[j-1]` behind `j > 0`, and in `extractDocsSection` `h.start` comes from a `range` index while `end` is either `len(lines)` or a strictly later heading's `start`, so `lines[h.start:end]` cannot go out of range. `headings[first+1:]` slices at most at `len`. No panic path found.

**5. NUL fix — correct and fully covered.** The raw `0x00` was replaced with the six-character JSON escape `\u0000`, which is the correct textual representation and preserves the fuzz payload's intent inside the JSON example. A corpus-wide scan at the PR head finds zero control bytes across all 39 files. `TestHelpResources_ValidUTF8NoControlBytes` iterates `resourcesFS.ReadDir("resources")` — the whole corpus, both extensions — permits only tab/LF/CR, and additionally asserts `utf8.Valid`. Coverage is complete.

**6. `sync.OnceValue` shared state — correctly handled, not a finding.** `docsRegistry` is unexported, so no external caller can obtain the pointer. The one exported accessor `DocsIndex()` allocates a fresh outer slice and deep-copies each entry's `Sections` via `append([]string(nil), ...)`. Every in-package consumer (`renderDocsIndex`, `unknownDocsTopicError`, `docsBody`, `DocsDocument`) is read-only, and the value is immutable after build. `TestDocsIndex_ReturnsIndependentCopies` pins the invariant directly.

**Settled decision #1 — comment confirmed accurate, deliberately not flagged.** `help_security.md` does contain `1234-5678-9012-3456`, `1234567890123456` and `user@example.com`. The `credit-card:separated` pattern `\b(\d{4})[-\s](\d{4})[-\s](\d{4})[-\s](\d{4})\b` matches the first verbatim and `email:standard` matches the address, so applying the Output Filter here would corrupt the documentation that explains the Output Filter. The block comment on `handleDocs` states this correctly and is consistent with CLAUDE.md Principle #7 / RFC-001 §3.7.

**Dependencies.** `go.mod` / `go.sum` are untouched — zero new dependencies, no license exposure.

**Other checks.** No `unsafe`, no `math/rand` (no randomness introduced), no goroutines or channels; the single `defer` is a correct debug-log closure. All 22 `docs(topic="X")` pointers introduced across the Go descriptions, CLI help and markdown resolve to real registry topics; the three surviving `yorishiro://help/` mentions are intentional explanatory prose, not stale links. The new `help_getting-started.md` CA guidance correctly scopes trust to the client under test and does not suggest disabling verification.

### Findings

| S-1 | LOW | internal/mcp/docs_section.go | 142-152 | InputValidation | CWE-407 (Inefficient Algorithmic Complexity) | `docsSectionMatches` recomputes the caller-supplied selector's lowercase and slug forms once per heading |

**S-1 detail.** `docsSectionMatches` computes `strings.ToLower(strings.TrimSpace(selector))` on entry and, on a miss, `slugifyDocsHeading(selector)` — both derived purely from the selector, both recomputed for every heading. `extractDocsSection` calls it once per heading with `level >= 2`, up to 57 in `help_examples.md`. An L-byte `section` argument therefore costs roughly O(H x L) with H <= 57.

*Not exploitable, and reported only because the complexity question was raised explicitly.* The growth is linear in input with a constant factor of at most ~114, not quadratic; each iteration's copies become garbage immediately so peak live memory stays O(L). Reaching it requires an authenticated caller on the loopback-bound, bearer-token-gated `/mcp` endpoint to send a selector large enough that transmitting it already dominates the cost — and such a caller has strictly more direct options. **Optional cleanup, not a merge blocker**: hoist the two selector-derived values out of the per-heading loop into `extractDocsSection` and pass them in.

---

usk6666 (commented):
## Code Review: APPROVED ✅

Solid, well-reasoned implementation. The markdown section slicer is correct against the real corpus (verified empirically: no cross-level heading shadowing, no slug collisions, no empty slugs, all 22 documents have balanced code fences, no indented headings/fences, no H5/H6), the `Out = any` + `out = nil` shape genuinely skips the go-sdk `StructuredContent` branch (confirmed at `go-sdk@v1.4.1/mcp/server.go:307,356-386`), the `sync.OnceValue` registry is properly defended against caller mutation, and every `yorishiro://help/` reference outside the legitimate resource registrations is gone. One MEDIUM on the CLI rendering surface this PR advertises, plus documentation-accuracy and test-invariant nits.

### Findings

| F-1 | MEDIUM | cmd/yorishiro-proxy/client.go | 23-44 | Correctness | CLI `docs` output is JSON-escaped markdown in both default formats |
| F-2 | LOW | internal/mcp/docs_section.go | 9-14 | CodeHealth | "worst case around 150 tokens" is measurably wrong (actual worst case ~567) |
| F-3 | LOW | internal/mcp/docs_section.go | 14 | TestQuality | `docsOutlineMaxLevel = 4` rests on an unguarded "no H5/H6 today" claim |
| F-4 | LOW | internal/mcp/docs_section.go | 214 | Correctness | Duplicate-heading note is a dead end - no way to reach the 2nd match |
| F-5 | NIT | internal/mcp/docs_tool.go | 1 | GoConventions | Package comment deviates from the package's house form |
| F-6 | NIT | internal/mcp/docs_tool_test.go | 394-407 | TestQuality | Aliasing test poisons the global registry if the copy ever regresses |

### Suggestions

#### F-1 (MEDIUM) - CLI `docs` renders as an escaped JSON blob

`docs` is the first tool in `internal/mcp` to return non-JSON `TextContent` - `grep -n 'TextContent{' internal/mcp/*.go` finds only `docs_tool.go:281` and `prompts.go:233`; every other tool returns a typed `out` that the SDK serialises to JSON, which the CLI then parses. `printToolResult` has no path for markdown:

- `--format` defaults to `""` (`client.go:820`) which goes to `resolveFormat`: **TTY gives `"json"`**, **pipe gives `"raw"`**.
- Both call `json.Marshal`/`MarshalIndent` on the whole `CallToolResult`, so a 16 KB document comes back as one `"text": "# macro\n\n## Parameters\n..."` escaped string.
- `--format table` does print the prose, but only by falling through `printResultTable`'s failure path, which first emits `warning: could not parse tool response as JSON; printing raw text output` to stderr.

The CLI entries were forced by `client_toollist_regression_integration_test.go` (lockstep with `tools/list`), but the four example invocations added at `client.go:40-44` promise a readable result they do not produce:

```
  yorishiro-proxy client docs
  yorishiro-proxy client docs macro
  yorishiro-proxy client docs macro "Variable substitution syntax"
  yorishiro-proxy client docs topic=fuzz_http section="Macro hook scopes"
```

Suggested fix (either is fine):

1. Special-case text-returning tools in `printToolResult` - for `toolName == "docs"`, write `extractTextContent(result)` directly regardless of format (the helper already exists at `client_format.go:54`); or
2. If CLI rendering is intentionally deferred to USK-1037, drop the four examples from `clientToolHelp["docs"]` or append a line such as `Pass --format table until the dedicated 'docs' subcommand lands (USK-1037).`

#### F-2 (LOW) - outline-size claim in the `docsOutlineMaxLevel` comment

```go
// H2-H4 keeps the worst case around 150 tokens for a single topic.
```

Measured over the shipped corpus (H2-H4 only, `"  " + "#"*level + " " + text`):

| Topic | Headings | Outline chars | est tokens |
|---|---|---|---|
| examples | 57 | 2,269 | **567** |
| configure | 46 | 1,726 | 431 |
| proxy_start | 38 | 1,627 | 406 |
| query | 49 | 1,448 | 362 |

Only 8 of 22 topics land under 150 est tokens. The *decision* is still right - 567 tok on an error path is an order of magnitude below the 4,961 est tok of the largest document - so only the number needs correcting, e.g. "keeps the worst case under ~600 tokens (help_examples.md), roughly an eighth of reading the largest document." (The companion claim "help_configure.md has 43 H2+H3 headings" is correct: 46 H2-H4 minus 3 H4s.)

#### F-3 (LOW) - no corpus guard for the H5/H6 assumption

`extractDocsSection` matches any `h.level >= 2`, and `help_docs.md:18` advertises *"Addressable levels are `##` through `######`"*. But H5/H6 appear in **neither** discovery surface: the index is H2-only (`docsH2Sections`) and the unknown-section outline caps at `docsOutlineMaxLevel = 4`. The comment's "No help document uses H5 or H6 today" is true right now (verified), but nothing enforces it - the first `#####` someone adds ships a section that exists yet cannot be discovered, which is precisely the failure class USK-1036 exists to close.

Cheap guard, alongside the new `TestHelpResources_ValidUTF8NoControlBytes` in `resources_test.go`:

```go
// TestHelpResources_NoHeadingsBelowOutlineDepth pins the assumption behind
// docsOutlineMaxLevel: H5/H6 are addressable via section= but appear in
// neither the index nor the unknown-section outline, so they would be
// undiscoverable. Raise docsOutlineMaxLevel (and docsH2Sections) before
// adding one.
func TestHelpResources_NoHeadingsBelowOutlineDepth(t *testing.T) {
	for _, rd := range helpResources {
		data, err := resourcesFS.ReadFile(rd.filename)
		if err != nil {
			t.Fatalf("read %s: %v", rd.filename, err)
		}
		for _, h := range scanDocsHeadings(strings.Split(string(data), "\n")) {
			if h.level > docsOutlineMaxLevel {
				t.Errorf("%s:%d: H%d %q exceeds docsOutlineMaxLevel=%d; it is addressable but listed nowhere",
					rd.filename, h.start+1, h.level, h.text, docsOutlineMaxLevel)
			}
		}
	}
}
```

#### F-4 (LOW) - the duplicate-heading note is itself a dead end

`help_security.md` has `## Actions` (line 45) and `### Actions` (line 347). `docs(topic="security", section="Actions")` returns the first plus:

```
(note: 2 sections in this topic share this heading; this is the first)
```

Returning the first match is the settled design and I am not challenging it - but `section=` has no disambiguator, so an agent that wanted the SafetyFilter `### Actions` is told a collision exists and given no way forward. That is the "agent gives up" shape this Issue targets. Suggest making the note actionable, e.g.:

```go
body += fmt.Sprintf("\n\n(note: %d sections in this topic share this heading; this is the first. "+
	"Call docs(topic=%q) without section= for the whole document, or select the intended one "+
	"by its parent heading.)", len(matches), topic)
```

(`topic` is already a parameter of `extractDocsSection`, so this is a one-line change.)

#### F-5 (NIT) - package comment form

30+ files in this package open with `// Package mcp <filename>.go <does X>` (`resend_http.go`, `fuzz_grpc.go`, `grpc_schema_tool.go`, `prompts.go`, ...). `docs_tool.go:1` instead opens with:

```go
// Package mcp exposes this server's own documentation through the docs tool.
```

`go/doc` concatenates every package comment in the package, so this one reads as an authoritative one-line summary of the entire `mcp` package - which it is not. The body of the comment is excellent and worth keeping; just retitle to the house form: `// Package mcp docs_tool.go exposes this server's own documentation through the docs tool.`

#### F-6 (NIT) - aliasing test can poison the package

`TestDocsIndex_ReturnsIndependentCopies` writes `"MUTATED"` into the returned `Sections` slice. The copy works today, so this is inert - but if it ever regresses, the write lands in the process-global `sync.OnceValue` registry and cascades into `TestDocsTool_IndexListsEveryTopic`, `TestDocsSection_Matching` and the rest, turning one clear failure into a noisy pile. Suggest restoring `original` immediately after the second read, or asserting non-aliasing without writing into the shared value.

### Verified, no action needed

- **Section slicer**, checked against the shipped corpus rather than only fixtures: fence toggling handles both markers, info strings, longer-fence nesting and unterminated fences; the `next.level <= h.level` terminator correctly carries H3/H4 children (`TestDocsSection_StopsAtSameLevelHeading`); `lines[h.start:end]` includes the heading line and excludes the next heading's line; slug derivation round-trips every advertised H2 (no collisions, no empty slugs across all 22 documents); the only duplicate is the known `help_configure.md` `## Usage Examples` pair and the note fires correctly.
- **Design decision 1** is genuinely achieved: with `Out = any`, `go-sdk@v1.4.1/mcp/server.go:307` leaves `elemZero` nil and line 356 skips the `StructuredContent` + duplicate-`TextContent` branch for a nil `out`. `TestDocsTool_ReturnsSingleTextContent` pins it.
- **No path traversal**: `DocsDocument` resolves through `reg.files[topic]` (map lookup on a compile-time registry), never string-concatenation of caller input into a path.
- **Concurrency**: `sync.OnceValue` + read-only registry + per-call `embed.FS.ReadFile` copies; `DocsIndex` deep-copies `Sections`. `-race` clean.
- **Index cost** measured at 6,209 chars (~1,552 est tokens) for 22 entries, consistent with the H2-only decision; largest document is `help_security.md` at 19,844 bytes (~4,961 est tokens).
- **Doc factual accuracy** spot-checked against source: encoder list matches `internal/macro/encoder.go:48-56`, `macro.ReservedKeyPrefix`/`IsReservedKey` exist in `internal/macro/reserved.go`, `query resource=ca_cert` / `fuzz_results` exist in `query_tool.go:269`, `install mcp --user-scope` and `server -mcp-http-addr` / `-open-browser` all exist.
- `gofmt` clean; no `t.Logf`, `TODO`, `FIXME` or `HACK` in the new code; errors wrapped with `%w` where a cause exists; all 4 exported symbols have godoc; `Out = any` + `return nil, nil, err` matches the package convention (`query_tool.go` uses it 30x); the stray NUL byte in `help_fuzz_ws.md` is correctly replaced with its JSON escape and is now guarded by the new `TestHelpResources_ValidUTF8NoControlBytes`.
- Settled decisions 2, 3, 4, 5, 6, 7, 8 are all implemented as described and are **not** flagged.

STATS - CRITICAL: 0 | HIGH: 0 | MEDIUM: 1 | LOW: 3 | NIT: 2

---

PR #109, opened 2026-09-13
ci(lint): consolidate linters into golangci-lint

usk6666:
## Motivation

`make lint` and the CI lint job depended on three separately hand-installed binaries: `staticcheck`, `ineffassign`, `gocyclo`.

**Accurate framing, given PR #107:** that PR already froze all three at explicit versions, so `@latest` is *not* currently armed. This PR is not "defuse `@latest`" — that was already done. It removes the two problems pinning alone cannot fix:

1. **Three independently hand-bumped pins.** Each is a `go install` from source, so each can independently raise its Go floor at bump time and take down the entire lint job on the `1.25.x` runner. That is the failure class PR #96 paid to diagnose (`staticcheck@latest` began requiring go1.26), and PR #107's own comment names USK-1043 as its successor.
2. **Local/CI divergence.** `make lint` ran whatever happened to be on each developer's PATH, with no version relationship to CI at all.

After this change the lint job installs **one** prebuilt binary via `golangci-lint-action`, and `make lint` **hard-fails** if the local binary is not the pinned version — parity becomes structural rather than conventional.

## What changed

| File | Change |
| -- | -- |
| `.golangci.yml` (new) | `version: "2"`, `linters.default: none`, exactly 5 linters, `run.timeout: 10m` |
| `.golangci-lint-version` (new) | `v2.13.2` — single source of truth |
| `Makefile` | `lint` target: keeps `gofmt -l .`, replaces the 4 tool invocations with `golangci-lint run`, adds presence + version guards. `fmt` untouched |
| `.github/workflows/ci.yml` | `lint` job only; `Install linters` step **and its comment block** deleted |
| `CLAUDE.md`, `.claude/skills/implement/SKILL.md`, `.claude/agents/implementer.md` (×2), `.claude/agents/fixer.md`, `.claude/settings.json` | 6 hardcoded descriptions of the linter set updated |

`.claude/settings.json`: dropped the now-dead `Bash(gocyclo:*)` / `Bash(ineffassign:*)`, added `Bash(golangci-lint:*)`, **kept `Bash(gofmt:*)`**. Without this the next implementer/fixer agent hits a permission prompt on its first `make lint`.

**Untouched:** `build`, `test`, `test-e2e-smoke`, `setup`, `depsec-tool` jobs. No Go source changed — `git diff --name-only` contains zero `.go` files.

## Deviation from the Issue's mapping table: gofmt stays in the Makefile

The Issue proposed `gofmt -l .` → golangci-lint's `gofmt` linter. **Not done**, for two verified defects:

**1. Reach (a P6 under-detection).** `golangci-lint run` only loads the default build configuration. Demonstrated on a scratch module containing one unformatted plain file and one unformatted `//go:build e2e` file:

```
--- toolchain `gofmt -l .` ---
badfmt.go
e2e_gap_test.go          <-- //go:build e2e

--- `golangci-lint run` with the gofmt formatter ENABLED ---
badfmt.go:6:1: File is not properly formatted (gofmt)
1 issues:
* gofmt: 1                <-- e2e_gap_test.go never loaded
```

This repo has **137** `//go:build e2e` files plus all of `tools/depsec`. Migrating gofmt would have silently stopped format-checking every one of them.

**2. Fidelity.** golangci-lint vendors `github.com/golangci/gofmt`, a fork that disagrees with the toolchain's gofmt on composite-literal indentation, so `make fmt` (`gofmt -w .`) would no longer guarantee `make lint` passes. Its `simplify` also defaults to `true` (= `gofmt -s`).

Incidentally, in golangci-lint **v2** `gofmt` is a *formatter*, not a linter — `linters.enable: [gofmt]` is rejected outright with `can't load config: gofmt is a formatter`, so the Issue's table row is not even expressible as written.

gofmt ships with the Go toolchain, so keeping it costs **zero** `go install`. **All three acceptance criteria still hold** — the criterion is "no `go install` in the CI lint job", and gofmt was never installed.

## Non-obvious addition: `staticcheck.checks`

golangci-lint's `staticcheck` linter is **not** a one-to-one swap for the `staticcheck` binary. It bundles staticcheck + gosimple + stylecheck + **quickfix**, and skips the non-default filter `cmd/staticcheck`'s `lintcmd` applies. Without an explicit `checks:` list, current `main` yields **14 findings** that `staticcheck ./...` never reported:

```
14 issues:
* staticcheck: 14
  QF1001 ×2  QF1002 ×2  QF1003 ×2  QF1008 ×1  QF1011 ×2  QF1012 ×2
  SA9003 ×2  ST1023 ×1
```

The `checks:` list in `.golangci.yml` reconciles this. With it, `main` reports **`0 issues.`** (golangci-lint bundles `honnef.co/go/tools v0.8.1` vs the pinned standalone v0.7.0, so this delta is expected and had to be closed explicitly.)

Pleasant confirmation: golangci-lint v2.13.2 bundles **gocyclo v0.6.0** and **ineffassign v0.2.0** — byte-identical to the versions `ci.yml` pinned.

## Rule-equivalence evidence (acceptance criterion #3)

Scratch module built **outside** the repo (`/tmp`, nothing committed; `git status` clean). Fixture complexities verified with `gocyclo -over 0` first:

```
16 equivscratch TestComplexity16InTestFile complexity_test.go:9:1
16 equivscratch Complexity16              complexity.go:55:1
15 equivscratch Complexity15              complexity.go:6:1
```

### BEFORE — old hand-installed tools

```
--- gofmt -l . ---
badfmt.go
e2e_gap_test.go
--- go vet ./... ---
vetbug.go:9:22: fmt.Sprintf format %d has arg "not-an-int" of wrong type string
--- staticcheck ./...  (v0.7.0) ---
ineff.go:8:2: this value of v is never used (SA4006)
scheck.go:10:9: should use strings.Contains(s, "x") instead (S1003)
vetbug.go:9:21: Printf format %d has arg #1 of wrong type string (SA5009)
--- ineffassign ./...  (v0.2.0) ---
/tmp/usk1043-equiv/ineff.go:8:2: ineffectual assignment to v
/tmp/usk1043-equiv/scheck.go:15:2: ineffectual assignment to x
--- gocyclo -over 15 -ignore '_test\.go$' .  (v0.6.0) ---
16 equivscratch Complexity16 complexity.go:55:1
```

### AFTER — `gofmt -l .` + golangci-lint v2.13.2 with this `.golangci.yml`

```
--- gofmt -l .  (still in make lint) ---
badfmt.go
e2e_gap_test.go
--- golangci-lint run ---
complexity.go:55:1: cyclomatic complexity 16 of func `Complexity16` is high (> 15) (gocyclo)
vetbug.go:9:22: printf: fmt.Sprintf format %d has arg "not-an-int" of wrong type string (govet)
ineff.go:8:2: ineffectual assignment to v (ineffassign)
scheck.go:15:2: ineffectual assignment to x (ineffassign)
ineff.go:8:2: SA4006: this value of v is never used (staticcheck)
scheck.go:10:9: S1003: should use strings.Contains(s, "x") instead (staticcheck)
vetbug.go:9:21: SA5009: Printf format %d has arg #1 of wrong type string (staticcheck)
7 issues:
* gocyclo: 1
* govet: 1
* ineffassign: 2
* staticcheck: 3
```

Identical on the axes this fixture exercises. The negative controls both hold:

- **Complexity exactly 15** (`Complexity15`) — reported by **neither**. Confirms `min-complexity: 15` ≡ `gocyclo -over 15` (strictly `>`).
- **Complexity 16 inside `_test.go`** — reported by **neither**. Confirms the `path: _test\.go` exclusion ≡ `gocyclo -ignore '_test\.go$'`.

> The AFTER block was produced with `--uniq-by-line=false` to show per-linter attribution. Under golangci-lint's **default** `uniq-by-line: true`, the run shows 5 issues instead of 7: `ineff.go:8` and `vetbug.go:9` are each reported by two linters, and only the first survives. **No line escapes the gate** — the set of failing lines is identical either way — but the credited linter can differ. Flagged for reviewer awareness; the default is kept.

## Other verified points

- **`install-mode: binary` is load-bearing.** The action defaults to `binary` (confirmed in `action.yml` at the pinned SHA) and downloads a prebuilt release. Setting `install-mode: goinstall` would reintroduce the exact bug being fixed: golangci-lint v2.13.x declares `go 1.26.0` in its own `go.mod`, and `actions/setup-go` exports `GOTOOLCHAIN=local`, so `go install` would deterministically fail on the 1.25.x runner. A workflow comment documents this.
- **The action SHA is the dereferenced *commit* SHA of an annotated tag.** `git/ref/tags/v9.3.0` → `d583c34f…` (type `tag`) → `git/tags/d583c34f…` → commit **`ba0d7d2ec06a0ea1cb5fa41b2e4a3ab91d21278a`**. Pinning `d583c34f…` would break action resolution — the PR #105 gotcha.
- **Supply-chain age gate:** v2.13.2 was published 2026-08-27, 17 days old — clears the 7-day window.
- **`needs: setup` + the `ui-dist` download are mandatory.** Verified by removing `dist/`: `golangci-lint run` fails with `pattern dist/*: no matching files found (typecheck)`.
- **`run.timeout: 10m`** — **corrected in review round 1.** v2 has **no** timeout by default; the `1m` default was v1's. `golangci-lint run --help` on v2.13.2 prints `Timeout for total work. Disabled by default`. So this value does not raise a too-low ceiling — it *introduces* one on purpose, so a pathological run fails as a named golangci-lint timeout instead of being killed opaquely when the job's own `timeout-minutes: 15` expires. The two are deliberately ordered lint < job. The figures previously quoted here ("2m10s end-to-end", "2m46s package loading") were mutually inconsistent and have been dropped rather than replaced.
- **Bare `golangci-lint run` ≡ `golangci-lint run ./...`** — verified against a nested package; no under-linting from omitting the path argument.
- **`tools/depsec` deliberately out of scope** (the Issue's deferred "decide at implementation time" question, now decided): it is a separate module — `golangci-lint run ./tools/...` errors with `no go files to analyze`, v2 does not walk nested modules from a parent run — it is stdlib-only, tops out at cyclomatic complexity **14** (under the threshold), and is already gated by the `depsec-tool` job, which is untouched. It also remains covered by `gofmt -l .`.

### S-1 (MEDIUM, blocking) — the `unused` / U1000 check was silently lost

The equivalence claim above was **wrong on one axis**, and the scratch fixture could not
expose it because it contained no dead code.

`cmd/staticcheck` v0.7.0 calls `cmd.AddAnalyzers(unused.Analyzer)` unconditionally
(`cmd/staticcheck/staticcheck.go:30` — only the quickfix set is flag-gated), and
`unused.Analyzer` carries no `NonDefault` marker, so `staticcheck ./...` reported **U1000
by default**. In golangci-lint v2, `unused` is a **separate linter**, not part of the
`staticcheck` linter — so `linters.default: none` left it off. `make lint` was
staticcheck's only caller, so unlike `govet` nothing else backstopped it.

Reproduced on a scratch module (`/tmp`, nothing committed) with one unused field and one
unused func:

```
--- staticcheck ./...            (old gate, v0.7.0) ---
main.go:7:2:  field unused is unused (U1000)
main.go:10:6: func helperNeverCalled is unused (U1000)
exit=1

--- golangci-lint run            (this PR, before the fix) ---
0 issues.
exit=0                            <-- regression

--- golangci-lint run            (this PR, after adding `- unused`) ---
main.go:7:2:  field unused is unused (unused)
main.go:10:6: func helperNeverCalled is unused (unused)
2 issues:
* unused: 2
exit=1                            <-- restored
```

`- unused` is now in `linters.enable`, with a comment recording *why* so it is not later
"tidied away" as part of golangci-lint's default set. This **restores** the replaced rule
set rather than adopting the default one, so the stated non-scope still holds. It is
zero-churn on this branch: `make lint` reports `0 issues.`

### F-2 (LOW) — the version-mismatch branch printed no install instructions

The install command lived only on the missing-binary path, so the mismatch branch ended
with "Install the pinned version (see above)" pointing at nothing — and that is the branch
the drift guard actually exists to hit. The two checks now share one hint block:

```
$ make lint GOLANGCI_LINT_VERSION=v9.9.9
==> gofmt check
golangci-lint version mismatch: have v2.13.2, pinned v9.9.9.
Local results can diverge from CI.
Install the pinned version -- do NOT use 'go install ...@latest':
golangci-lint v9.9.9 declares 'go 1.26.0', so building it from source
needs go1.26+, which is exactly the breakage USK-1043 removes.

  curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/v9.9.9/install.sh \
    | sh -s -- -b /home/user/go/bin v9.9.9
```

### F-4 + S-5 (NIT/LOW) — degenerate `.golangci-lint-version` content

Previously an empty/missing file produced `...have v2.13.2, pinned .`, and the unquoted
`$(shell cat ...)` substitution was executed textually at four points. The `lint` recipe
now validates the file's shape (`^v[0-9]+\.[0-9]+(\.[0-9]+)?$`, exactly one line — the same
shape `golangci-lint-action`'s `src/version.ts` enforces) *before* the value is used, and
names the specific defect:

```
empty      -> .golangci-lint-version must hold exactly one version tag, e.g. v2.13.2.
              -> the file is empty.
missing    -> ... -> the file is missing.
CRLF       -> ... -> it has CRLF line endings; re-check it out with LF (.gitattributes pins this).
malformed  -> ... -> got: v2.13.2"; touch /tmp/pwned_usk1043; echo "
```

All four exit non-zero, and the injection-shaped payload no longer reaches a substitution
point (verified: `/tmp/pwned_usk1043` was not created). To be explicit, this is a
**robustness and error-message fix, not a vulnerability fix** — anyone who can write that
file can equally write the Makefile recipe. A new `.gitattributes` pins the file to
`eol=lf`.

### S-3 (LOW) — workflow comment understated the action

The comment said the action "just places the pinned binary on PATH". At the pinned commit,
`src/run.ts` calls `restoreCache()` (:237) and `plugins.install()` (:239) **before** the
`install-only` early return (:247-248), and `postRun()` calls `saveCache()` unconditionally
(:261). The step therefore also restores/saves a GitHub Actions cache of
`$HOME/.cache/golangci-lint` (`GOLANGCI_LINT_CACHE` — analysis results, not the binary).
Comment corrected; the load-bearing `install-mode: binary` warning is untouched.

### S-4 (LOW) — license-policy scope

golangci-lint is GPL-3.0 (verified at `v2.13.2`), while `CLAUDE.md` prohibits "All GPL
variants". There is **no** compliance issue — it is in no `go.mod`/`go.sum`, there is no
`tools.go`, nothing GPL links into the shipped Apache-2.0 binary or the npm artifacts, and
a separate executable's diagnostic output is not a derivative work. But a future reader
would get a clean `/license-check` while a GPL-3.0 binary gates the build, so the policy now
states it governs **distributed** dependencies and that build-time-only tools are out of
scope, with the two caveats (never vendor/redistribute it; `golangci-lint custom` would
produce a combined GPL-3.0 binary that must not be shipped).

### S-2 — not fixed, by design

Absence of download-checksum verification in `golangci-lint-action` is an already-accepted
trade-off: `install-mode: goinstall` would deterministically reintroduce the
go1.26 / `GOTOOLCHAIN=local` break this PR exists to remove. Left as is.

## Verification

Re-run after the round-1 fixes:

```
make build                exit=0
make lint                 exit=0   ==> gofmt check / ==> golangci-lint v2.13.2 / 0 issues.
make test                 exit=0
golangci-lint config verify  exit=0
```

`make lint` guard rails re-exercised after the restructure: **both** the missing-binary and
the version-mismatch paths now print the full install instructions and exit non-zero; and
empty / missing / CRLF / malformed `.golangci-lint-version` each produce a named error and
exit non-zero (`.golangci-lint-version` restored byte-identical afterwards). `version
--short` emits no leading `v`, so the Makefile's `v` prefix is correct.

`grep 'go install' .github/workflows/ci.yml` now matches only the explanatory comment warning against it.

## Note for the reviewer

On the first `make test` run, `internal/mcp` hit the 4m package timeout (`panic: test timed out after 4m0s`) with **no assertion failures** — every listed test had PASSed. Standalone it takes **219s** against the 240s cap, so it tips over under full-suite parallel load. The rerun passed cleanly (exit 0). This is a **pre-existing borderline-timeout flake unrelated to this PR** (which changes zero `.go` files); possibly worth a follow-up to raise the timeout or split the package.

Resolves USK-1043
Linear: https://linear.app/usk6666/issue/USK-1043

https://claude.ai/code/session_01LPjkMZA8ADA3GaErVTeyYA


usk6666 (commented):
## Code Review: APPROVED ✅

A well-executed, well-evidenced CI-reliability PR. I re-measured every load-bearing claim rather than taking the PR body at its word, and the substantive ones all hold: `golangci-lint run` reports **0 issues** on this branch, the legacy trio (`staticcheck` v0.7.0 / `ineffassign` v0.2.0 / `gocyclo -over 15 -ignore '_test\.go$'`) also reports 0, the "14 findings without `checks:`" figure reproduces **exactly** (QF1001×2 QF1002×2 QF1003×2 QF1008×1 QF1011×2 QF1012×2 SA9003×2 ST1023×1), and all three Makefile guard rails fail closed. The one real defect is a factually wrong statement about golangci-lint's default `run.timeout` baked into `.golangci.yml`, plus two minor accuracy/reach gaps. No Go source changed; scope is clean.

### Verification performed (read-only, in an isolated worktree)

| Claim | Result |
| -- | -- |
| `make lint` exit 0, `0 issues.` | ✅ reproduced (19s warm cache) |
| `golangci-lint config verify` | ✅ exit 0 — schema valid for `version: "2"` |
| `staticcheck ./...` (v0.7.0) / `ineffassign ./...` (v0.2.0) | ✅ 0 findings each — matches golangci-lint |
| `gocyclo -over 15 -ignore '_test\.go$' .` (v0.6.0) | ✅ 0 findings — matches `min-complexity: 15` |
| golangci-lint bundles `fzipp/gocyclo v0.6.0` + `gordonklaus/ineffassign v0.2.0` | ✅ byte-identical to the old pins (go.mod at `v2.13.2`) |
| `min-complexity: 15` ≡ `-over 15` (strictly `>`) | ✅ both call `fzipp/gocyclo` `SortAndFilter(top, over)`; an `-over 10` probe shows a large cohort at exactly 15 that neither reports |
| `_test\.go` exclusion scoped to `gocyclo` only | ✅ proven: in the no-`checks:` run, staticcheck still fired inside `layer_test.go` / `frame_test.go` / `harness_test.go` / `session_wsh2_test.go` |
| `-QF*` wildcard actually works | ✅ 11 QF findings present without the list, 0 with it, while SA/S/ST checks keep firing |
| `linters.default: none` is the v2 key | ✅ `config verify` accepts it; v1's `disable-all` is not used |
| `govet` default set ≡ `go vet` | ✅ `defaultAnalyzers` in `pkg/golinters/govet/govet.go` mirrors `go1.26.1 cmd/vet/main.go` |
| golangci's staticcheck bundles staticcheck+stylecheck+simple+quickfix | ✅ `slices.Concat(...)` in `pkg/golinters/staticcheck/staticcheck.go` |
| cmd/staticcheck filters `NonDefault` checks | ✅ `lintcmd/cmd.go` builds `["all", "-<each NonDefault>"]` |
| golangci-lint v2.13.2 `go.mod` declares `go 1.26.0` | ✅ |
| golangci-lint vendors the `github.com/golangci/gofmt` fork | ✅ present in go.mod |
| `actions/setup-go` exports `GOTOOLCHAIN=local` | ✅ `setGoToolchain()` in `src/main.ts` at the pinned SHA |
| action `version-file` supports `.golangci-lint-version`, tolerates the `v` prefix + trailing newline | ✅ `src/version.ts`: `"v" + content.trim().replace(/^v/gi, "")` |
| `install-only: true` still puts the binary on PATH | ✅ `src/run.ts` calls `core.addPath(...)` **before** the early return |
| action SHA `ba0d7d2e…` is the dereferenced commit of annotated tag `v9.3.0` | ✅ `d583c34f…` (type `tag`) → `ba0d7d2e…` (type `commit`) — the PR #105 gotcha avoided |
| 137 `//go:build e2e` files | ✅ exact |
| `tools/depsec` tops out at complexity 14 | ✅ `fetchAlerts` = 14 |
| Guard rail: binary absent | ✅ prints install command, exits 1 |
| Guard rail: version mismatch (`GOLANGCI_LINT_VERSION=v9.9.9`) | ✅ exits 1 |
| Guard rail: empty version variable | ✅ fails closed (see F-4) |

**Shell correctness**: both new guard blocks are single logical lines (backslash-continued), so `exit 1` terminates the recipe shell and fails the target — confirmed by running them. `{ ...; }`, `command -v`, `[ ]`, `if/fi` are all POSIX; nothing bash-specific. `$$(...)` → `$(...)` and `\\` → `\` expand correctly (verified in the printed output). `lint` remains in `.PHONY`.

**Workflow**: `needs: setup` preserved, the `ui-dist` download retained (mandatory — `//go:embed dist/*`), `timeout-minutes: 15` > `run.timeout: 10m` is coherent ordering, and `ci.yml` is the **only** `make lint` caller in the repo, so no other workflow is affected. No `go install` remains in the lint job (AC #2 ✅). Leaving the duplicate `go vet ./...` in the `build` job is **acceptable, not a defect** — that job never runs `make`, so it keeps a standalone vet signal even if the lint job is skipped.

**Docs**: I grepped the whole repo for `staticcheck|ineffassign|gocyclo`; no site was missed. The remaining in-source hits are gocyclo-threshold comments, which stay accurate (same tool, same threshold 15). `.claude/settings.json` is valid JSON, `Bash(golangci-lint:*)` is correctly ordered between `goimports` and `grep`, the removed `gocyclo`/`ineffassign` entries are genuinely dead, and `Bash(gofmt:*)` is correctly retained.

### Findings

| F-1 | MEDIUM | `.golangci.yml` | 22-24 | Documentation | "golangci-lint's own default is 1m" is wrong for v2 — the default is *disabled*; the two measurements in the same comment also contradict each other |
| F-2 | LOW | `Makefile` / `.claude/agents/*.md` | 87-92 / 118-120, 178-181 | Documentation | The version-mismatch branch says "(see above)" but never prints the install command; the agent docs claim it does |
| F-3 | LOW | `.golangci.yml` | 12-14 | Config | gocyclo silently loses reach over `tools/depsec`; the in-repo "Reach" note covers only gofmt |
| F-4 | NIT | `Makefile` | 14 | Makefile | Missing/empty/CRLF `.golangci-lint-version` fails closed but with a confusing `pinned .` message |

### Suggestions

**F-1 — `.golangci.yml:22-24`**

```yaml
run:
  # golangci-lint's own default is 1m, which is not enough for this module on a
  # cold CI runner (measured 2m10s end-to-end with an empty Go build cache;
  # package loading alone logged 2m46s).
  timeout: 10m
```

Measured against the pinned binary:

```
$ golangci-lint run --help | grep timeout
      --timeout duration   Timeout for total work. Disabled by default
```

and `pkg/commands/run.go` at `v2.13.2`:

```go
const defaultTimeout = 0 * time.Minute
```

`1m` was **v1**'s default. In v2 there is no timeout at all unless you set one, so `timeout: 10m` does not *raise* a too-low ceiling — it **introduces** one where none existed. The setting itself is still a defensible choice (it makes the failure self-naming rather than an opaque job kill, exactly as the `ci.yml` comment argues, and the effective budget is larger than the 10-minute *job* cap this PR replaces), so no behaviour change is needed — only the justification. Separately, "measured 2m10s end-to-end … package loading alone logged 2m46s" cannot both be wall-clock: package loading cannot exceed end-to-end. Suggested rewrite:

```yaml
run:
  # golangci-lint v2 has NO timeout by default (`--timeout`: "Disabled by
  # default"). We set an explicit one so a pathological run fails as a named
  # golangci-lint timeout rather than an opaque job kill at `timeout-minutes`.
  # Cold-cache end-to-end on a CI runner measured ~2m10s wall, so 10m is ~5x
  # headroom. (The ~2m46s package-loading figure is summed across parallel
  # loaders, not wall time.)
  timeout: 10m
```

Please re-check the two numbers against the actual run before rewording — I could not verify them, only that as stated they are mutually inconsistent. This repo has been bitten by confidently-stated-but-unverified values before (USK-1034), which is why this is MEDIUM rather than a NIT.

**F-2 — `Makefile:87-92` + `.claude/agents/implementer.md:178-181` + `.claude/agents/fixer.md:118-120`**

Actual output of the mismatch path (reproduced with `make lint GOLANGCI_LINT_VERSION=v9.9.9`):

```
==> gofmt check
golangci-lint version mismatch: have v2.13.2, pinned v9.9.9.
Local results can diverge from CI. Install the pinned version (see above).
```

"(see above)" refers to the missing-binary block, which is *not* printed on this path — a developer who has the wrong version gets no `curl` command at all. Both agent docs assert the opposite: "hard-fails if `golangci-lint` is missing from PATH **or** is not the version pinned …; it prints the install command." Since the whole point of the guard is to stop local/CI drift, this is the branch most likely to be hit in practice. Factor the instructions into a make variable and use it in both branches:

```make
define GOLANGCI_INSTALL_HINT
echo "Install the pinned version -- do NOT use 'go install ...@latest':"; \
echo "golangci-lint $(GOLANGCI_LINT_VERSION) declares 'go 1.26.0', so building"; \
echo "it from source needs go1.26+, which is exactly the breakage USK-1043 removes."; \
echo ""; \
echo "  curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/$(GOLANGCI_LINT_VERSION)/install.sh \\"; \
echo "    | sh -s -- -b $$(go env GOPATH)/bin $(GOLANGCI_LINT_VERSION)"
endef
```

then call `$(GOLANGCI_INSTALL_HINT);` from both the `command -v` block and the mismatch block. Simply repeating the two `echo` lines is fine too. Either way, no doc change is then needed.

**F-3 — `.golangci.yml:12-14`**

The "Reach" note is framed entirely around gofmt, but `gocyclo` has the *same* reach loss and it is the one that actually regressed:

- **Before**: `gocyclo -over 15 -ignore '_test\.go$' .` is a filesystem walk from the repo root, so it covered `tools/depsec/{main,plan,semver,render,registry}.go`.
- **After**: `golangci-lint run` cannot see a nested module, so those five files get no complexity check at all.
- The `depsec-tool` job does **not** close the gap — it runs `gofmt -l .`, `go vet ./...`, `go test -race ./...`, no complexity check.

(The 137 e2e files are *not* affected: every one is `*_test.go`, so `-ignore '_test\.go$'` already excluded them — I verified there are zero non-`_test.go` files carrying `//go:build e2e`.)

Currently benign — I measured `tools/depsec`'s ceiling at complexity 14 (`fetchAlerts`, `main.go:105`) — and the commit message does justify it. But the commit message is not discoverable from the config; a reader of `.golangci.yml` or CLAUDE.md sees only the gofmt reach note. One line in the config is enough:

```yaml
#   1. Reach. `gofmt -l .` walks every .go file; `golangci-lint run` only sees
#      the default build configuration of this module, which excludes all 137
#      `//go:build e2e` files and all of tools/depsec. Note gocyclo shares this
#      blind spot: tools/depsec (separate module, currently max complexity 14)
#      is no longer complexity-checked. Accepted -- it is stdlib-only and gated
#      by the `depsec-tool` job.
```

**F-4 — `Makefile:14`**

```make
GOLANGCI_LINT_VERSION := $(shell cat .golangci-lint-version 2>/dev/null)
```

Verified with `make lint GOLANGCI_LINT_VERSION=`:

```
golangci-lint version mismatch: have v2.13.2, pinned .
```

It **fails closed**, which is the important property, but the message is confusing and does not point at the real cause (missing/empty file). A checkout with CRLF line endings degrades the same way (`$(shell)` strips the trailing `\n` but keeps `\r`), and there is no `.gitattributes` in the repo. Optional hardening:

```make
ifeq ($(strip $(GOLANGCI_LINT_VERSION)),)
$(error .golangci-lint-version is missing or empty)
endif
```

(guard it so it does not fire for unrelated targets), plus a one-line `.gitattributes`:

```
.golangci-lint-version text eol=lf
```

### Acceptance criteria

1. **`make lint` via golangci-lint, zero findings on `main`** — ✅ `0 issues.`, exit 0. The branch is one commit behind `main` (`#108`), but that commit touches only `go.mod` / `go.sum` / `web/`, no `.go` files.
2. **No `go install` in the CI lint job** — ✅ only the explanatory comment warning against it remains.
3. **Rule equivalence confirmed** — ✅ independently reproduced against the repo itself (not only the scratch module): identical 0-finding result for all four linters, identical bundled gocyclo/ineffassign versions, `-QF*` / `-SA9003` / `-ST1023` shown to be exactly the necessary and sufficient negations, and the `_test\.go` exclusion shown not to leak onto staticcheck. The one intentional difference is `honnef.co/go/tools v0.8.1` (bundled) vs `v0.7.0` (old pin) — a staticcheck *version bump*, correctly called out in the PR body and currently a no-op on this codebase.

Non-scope honoured: `linters.default: none`, exactly four linters, zero `.go` files touched, `tools/depsec` untouched.

---

usk6666 (commented):
## Security Review: CHANGES REQUESTED ❌

No runtime code is touched (zero `.go` files), so the proxy's TLS / network / wire-parsing attack surface is unchanged — those categories are N/A. The action SHA pin is **correct** (annotated tag properly dereferenced to the commit SHA), workflow token scope is unchanged and minimal, and no Scorecard check regresses. One verified defect: the new `.golangci.yml` silently drops the `unused` (U1000) check that the standalone `staticcheck ./...` it replaces ran by default, so the lint gate loses coverage while the PR body asserts the rule set is "identical on every axis".

### Security Findings

| S-1 | MEDIUM | `.golangci.yml` | 27-33 | CIIntegrity | CWE-1127 | `unused` (U1000) is not enabled; `cmd/staticcheck` ran it by default, so the lint gate silently lost a check with no backstop elsewhere in CI |
| S-2 | LOW | `.github/workflows/ci.yml` | 68-72 | SupplyChain | CWE-494 | Binary install path performs no checksum/signature verification, replacing a `go install` path backed by the Go checksum transparency log (already-accepted trade-off) |
| S-3 | LOW | `.github/workflows/ci.yml` | 60-61 | CIIntegrity | CWE-1104 | `install-only: true` is not install-only: the action also restores/saves a GH Actions cache and would build a custom binary if a `.custom-gcl.*` file existed. The comment's claim is inaccurate |
| S-4 | LOW | `.golangci-lint-version` | 1 | Dependencies/License | CWE-710 | golangci-lint is **GPL-3.0**, replacing MIT/MIT/BSD-3-Clause tools. Policy as written prohibits GPL; policy as operationalized covers only `go.mod`. No actual legal risk — but the divergence is undocumented |
| S-5 | LOW | `Makefile` | 14, 81-93 | SupplyChain | CWE-78 | `.golangci-lint-version` content is interpolated unquoted into the `lint` recipe; crafted content executes arbitrary shell. Not a trust-boundary crossing — hardening only |

---

### S-1 — MEDIUM — `unused` (U1000) silently dropped

`cmd/staticcheck` at the pinned v0.7.0 registers `unused.Analyzer` unconditionally, and `unused.Analyzer`'s `RawDocumentation` carries **no** `NonDefault: true` — so `staticcheck ./...` reported U1000 by default. In golangci-lint v2 `unused` is a **separate linter**, not part of the `staticcheck` linter, and `linters.default: none` leaves it off.

Reproduced on a scratch module (`/tmp`, nothing committed):

```go
type widget struct { used int; neverRead int }
func neverCalled() { fmt.Println("dead") }
```

```
--- staticcheck v0.7.0 ./...              (the gate this PR removes)
unusedcode.go:7:2:  field neverRead is unused (U1000)
unusedcode.go:10:6: func neverCalled is unused (U1000)
exit=1

--- golangci-lint v2.13.2 + this PR's .golangci.yml
0 issues.
exit=0
```

Two things make this worth a round rather than a footnote:

1. **No backstop.** `go vet ./...` still runs independently in the `build` job, in `make vet`, and in `depsec-tool`, so any `govet`-set delta is covered. `staticcheck` had no second invocation anywhere — `make lint` was its only caller (`grep -rn staticcheck .github/ Makefile` confirms).
2. **The fix is free.** Adding `- unused` to `linters.enable` is **`0 issues.`** on this branch today (verified: `golangci-lint run --default=none --enable=gocyclo --enable=govet --enable=ineffassign --enable=staticcheck --enable=unused` → exit 0), and the same invocation correctly reports both U1000 findings on the fixture above, proving the linter fires.

The scratch equivalence matrix in the PR body is otherwise sound — I built an independent positive control and confirmed `SA5009`, `S1003`, `ineffassign` ×2 and `govet printf` all still fire under this config, so `-QF*` does not over-exclude. And the three `checks:` exclusions beyond golangci-lint's own defaults are **faithful**, verified in `dominikh/go-tools@v0.7.0` source:

- `SA9003` → `NonDefault: true` in `staticcheck/sa9003/sa9003.go`
- `ST1023` → `NonDefault: true` in `stylecheck/st1023/st1023.go`
- `QF*` → `cmd/staticcheck/staticcheck.go` adds `quickfix.Analyzers` only behind the `-debug.run-quickfix-analyzers` flag

The fixture set simply had no unused-code case, so the equivalence test could not have caught this.

**Remediation** — either closes it:
- Add `- unused` to `linters.enable` in `.golangci.yml`; or
- If the omission is deliberate, say so in `.golangci.yml` alongside the existing gofmt rationale and correct the PR body's "Identical on every axis" claim, since U1000 is an axis where it is not.

---

### S-2 — LOW — no download integrity verification (accepted trade-off, recorded)

Read `src/install.ts` at the pinned commit: `installBin()` builds the asset URL and calls `tc.downloadTool(assetURL)` with **no digest argument**, then extracts. There is no checksum, signature, or attestation check anywhere in the path.

The delta versus what it replaces is real: `go install honnef.co/go/tools/cmd/staticcheck@v0.7.0` verifies module content against `sum.golang.org`, a public transparency log. The new path's only integrity guarantee is TLS to `github.com` plus the mutability properties of a GitHub release asset — and release assets can be replaced in place at a fixed tag, so `v2.13.2` pins a *name*, not *content*.

I am **not** disputing the accepted reasoning. I confirmed `install-mode` defaults to `binary` in `action.yml` at the pinned SHA, and `goinstall` would indeed hit the `go 1.26.0` / `GOTOOLCHAIN=local` break this PR exists to remove. The trade-off is correct.

Two verified notes for the record:

- **The developer path is stronger than CI.** The install command `make lint` prints fetches `install.sh` from a **tag-pinned** URL (`/v2.13.2/install.sh`, not `HEAD`), and that script *does* `hash_sha256_verify` the tarball against `golangci-lint-2.13.2-checksums.txt`. The CI path does not. Note that `make lint` only *prints* this — it never executes a network fetch. That part is clean.
- **Attestations exist.** `gh api repos/golangci/golangci-lint/attestations/sha256:2277d43b98ec0054280f2ac26b53268bae97682444678a59a657dd565da021d6` returns **2 build provenance attestations** for `golangci-lint-2.13.2-linux-amd64.tar.gz`. So if this is ever revisited, `gh attestation verify` is available without giving up the prebuilt binary.

**Release-age gate: PASSES.** v2.13.2 published `2026-08-27T23:23:58Z` → 18 days at review time, clearing the 7-day window in CLAUDE.md.

---

### S-3 — LOW — `install-only: true` does more than the comment says

The workflow comment states the action "just places the pinned binary on PATH". Reading `src/run.ts` at the pinned commit, the `installOnly` early-return happens *after*:

```ts
await core.group(`Restore cache`, restoreCache)
const binPath = await core.group(`Install`, () => install().then(plugins.install))
core.addPath(path.dirname(binPath))
if (installOnly) return
```

and `postRun()` unconditionally calls `saveCache()`.

- **Cache**: `src/cache.ts` caches `$HOME/.cache/golangci-lint` (analysis results only — not the binary, which `installBin` re-downloads each run), keyed by OS + interval + `go.mod` sha1. This is a **new** surface for the lint job. Not exploitable: GH Actions cache scope isolation means a fork or feature-branch PR cannot write into `main`'s scope, and `main`'s scope is behind branch protection.
- **`plugins.install`**: scans for `.custom-gcl.{yml,yaml,json}` and, if present, runs `golangci-lint custom` — building and installing a binary from arbitrary Go modules named in that file. No such file exists in this repo. Even if a fork PR added one, `ci.yml` has no `secrets.*` references and inherits `contents: read`, and the `test` job already executes PR-authored Go code by design, so this grants nothing above baseline.

I also confirmed golangci-lint v2 has **no** in-config `.so` plugin loading — `golangci-lint run` cannot load code from `.golangci.yml` alone. So a malicious config file by itself is not a code-execution vector.

**Remediation**: purely a comment accuracy fix — either reword to note the cache restore/save, or add `skip-cache: true` if the caching is unwanted in an install-only step.

---

### S-4 — LOW — GPL-3.0 tool under a policy that prohibits GPL (conclusion: policy does not apply)

Licences verified from upstream this session, not from memory:

| Tool | Licence | Role |
|---|---|---|
| golangci-lint | **GPL-3.0** | new |
| honnef.co/go/tools (staticcheck) | MIT | replaced |
| gordonklaus/ineffassign | MIT | replaced |
| fzipp/gocyclo | BSD-3-Clause | replaced |
| golangci/golangci-lint-action | MIT | new |

**Conclusion: the Dependency License Policy does not apply, and there is no legal or compliance risk.** Reasoning:

1. golangci-lint is not a dependency. It is absent from `go.mod` and `go.sum` (`grep -i golangci go.mod go.sum` → no match), there is no `tools.go`, and no Go source references it. Nothing GPL is linked into the shipped Apache-2.0 binary or the npm artifacts.
2. It is invoked as a **separate executable** that reads source and emits diagnostics. GPL-3.0 copyleft attaches to distributing the covered work or a work based on it; a tool's diagnostic output is not a derivative work of the tool — the same reason GCC's licence does not infect compiled programs.
3. The policy's operational definition confirms the scope. `/license-check` Step 1 is `go list -m -json all` — it enumerates Go module dependencies and would never surface golangci-lint at all.

Flagging at LOW purely because CLAUDE.md states "**All** GPL variants" with no dependency-versus-build-tool carve-out, and the repo demonstrably cares about this dimension. A future reader running `/license-check` will get a clean report while a GPL-3.0 binary is a required part of the build gate — an undocumented divergence, not a violation.

**Remediation**: one sentence under "Dependency License Policy" scoping it to distributed dependencies and noting that build-time-only tools (golangci-lint, GPL-3.0) are out of scope because their output is not a derivative work. Two caveats worth capturing at the same time: the conclusion holds only while golangci-lint is never vendored or redistributed in a release artifact, and `golangci-lint custom` would produce a combined GPL-3.0 binary if one were ever built and shipped.

---

### S-5 — LOW — `.golangci-lint-version` content reaches the shell unquoted (hardening only)

`GOLANGCI_LINT_VERSION := $(shell cat .golangci-lint-version)` is substituted textually into the `lint` recipe at four points, including `if [ "$$have" != "$(GOLANGCI_LINT_VERSION)" ]`. Confirmed executable:

```
$ make lint 'GOLANGCI_LINT_VERSION=v1.0"; touch /tmp/usk1043-pwned; echo "'
/bin/sh: 2: [: missing ]
...
$ ls /tmp/usk1043-pwned
-rw-rw-r-- 1 user user 0 ... /tmp/usk1043-pwned     <-- executed
```

**This is not a trust-boundary crossing and I do not consider it exploitable.** Anyone who can write `.golangci-lint-version` can equally write the `Makefile` recipe itself; a fork PR already gets arbitrary execution via `go test ./...` in the `test` job, against a read-only token with no secrets on a public repo; and there is no automated path that edits this file (no `.github/dependabot.yml`, and `tools/depsec` has no golangci reference). The pre-existing `$(COMMIT)` / `$(VERSION)` recipe substitutions have the same shape, so this is not a regression in kind.

Notably the **CI** side is already hardened for free: `src/version.ts` normalises the file content and validates it against `/^v(\d+)\.(\d+)(?:\.(\d+))?$/`, rejecting anything else with a hard error — so the `version-file` input introduces no delta over a workflow-file pin (both require push access). Only the local Makefile path is unvalidated.

**Remediation (optional)**: apply the same shape check in the Makefile, e.g. guard on `echo "$(GOLANGCI_LINT_VERSION)" | grep -qE '^v[0-9]+\.[0-9]+\.[0-9]+$'` before use.

---

### Verified clean — no findings

- **Action SHA pin is correct.** `git/ref/tags/v9.3.0` → `d583c34f0599d37dbac4a198b9c83201be380893` with `"type":"tag"` — an **annotated** tag, the exact PR #105 trap. Dereferenced via `git/tags/d583c34f…` → commit **`ba0d7d2ec06a0ea1cb5fa41b2e4a3ab91d21278a`**, matching the pin byte for byte. Cross-checked against the `/tags` listing. Both the tag object and the commit report `verification.verified: true`, tagger `ldez`.
- **Token scope unchanged.** `ci.yml` keeps top-level `permissions: contents: read`; the `lint` job adds no job-level `permissions`. The action's `github-token` (default `${{ github.token }}`) is used only by `fetchPatch()` for `only-new-issues`, which is not reachable under `install-only: true`.
- **No secrets exposure.** `grep 'secrets\.' .github/workflows/ci.yml` → no match.
- **No workflow injection.** `grep '\${{' .github/workflows/ci.yml` → **zero occurrences** in the entire file. No interpolation into any `run:` block.
- **Scorecard: no regression.** Pinned-Dependencies — every `uses:` across all five workflows remains SHA-pinned with a version comment, the new one included. Token-Permissions — top-level read-only preserved. The PR also removes three `go install` lines from a `run:` block, which is neutral-to-positive.
- **Fail-closed everywhere it matters.** Empirically: missing binary → `exit 2`; version mismatch → `exit 2`; missing/empty `.golangci-lint-version` → `exit 2` (message reads `pinned .`, a cosmetic nit); malformed `.golangci.yml` → `exit 3`; unknown linter name → `exit 3`. Nothing silently skips. (A typo'd *check ID* inside `staticcheck.checks` is silently ignored, but since `all` is the only positive entry, the only reachable direction is stricter, never looser.)
- **`version-file` is a non-finding.** As above — validated by a strict regex action-side, and equally attacker-controlled as the workflow file it substitutes for.

---

usk6666 (commented):
Round-1 fixes pushed in 59b3042c. Per finding:

**S-1 (MEDIUM, blocking) — fixed.** Reproduced the lost `unused`/U1000 check before fixing it. On a scratch module (`/tmp`, nothing committed) with one unused field + one unused func: `staticcheck ./...` (v0.7.0) → 2 U1000, exit 1; `golangci-lint run` with the merged config → `0 issues.`, exit 0; with `- unused` added → 2 findings, exit 1. Source-confirmed the mechanism rather than taking it on faith: `cmd/staticcheck/staticcheck.go:30` calls `cmd.AddAnalyzers(unused.Analyzer)` unconditionally (only the quickfix set is flag-gated), and `unused/` contains no `NonDefault` marker. `- unused` is now in `linters.enable` with a comment saying why, so it is not later tidied away. Zero-churn as predicted: `make lint` still reports `0 issues.` on this branch.

**F-1 (MEDIUM) — fixed.** Confirmed against the installed v2.13.2: `golangci-lint run --help` prints `Timeout for total work. Disabled by default`. The comment now says v2 has no default timeout and that `10m` *introduces* a ceiling so a pathological run fails as a named golangci-lint timeout rather than an opaque job kill. Both inconsistent figures dropped, not replaced — I did not measure a cold-cache CI run this session, so I quote no number.

**F-2 (LOW) — fixed.** The two guards now share one hint block, so both emit install instructions. Verified both paths:
- `make lint GOLANGCI_LINT_VERSION=v9.9.9` → mismatch message **plus** the full curl command, exit 1.
- `make lint` with golangci-lint off PATH → not-found message plus the same block, exit 1.

**F-3 (LOW) — fixed.** Added to the "Reach" bullet, using your precise framing: gocyclo shares only the `tools/depsec` half of the blind spot (the old `-ignore '_test\.go$'` filesystem walk reached that nested module; `golangci-lint run` cannot). Verified independently: `tools/depsec` has 5 non-test `.go` files and tops out at **14** (`fetchAlerts`, `main.go:105`), and 0 non-`_test.go` files carry `//go:build e2e` (137 total, all tests) — so the e2e files are explicitly noted as *not* a gocyclo regression.

**F-4 + S-5 (NIT/LOW) — fixed.** The `lint` recipe validates the file's shape (`^v[0-9]+\.[0-9]+(\.[0-9]+)?$`, exactly one line — matching the action's `src/version.ts`) *before* the value is used, and names the defect: empty / missing / CRLF / malformed each get a distinct line and exit 1. The injection-shaped payload `v2.13.2"; touch /tmp/pwned_usk1043; echo "` is now rejected and did **not** create the file. Added `.gitattributes` with `eol=lf`. As you framed it, this is committed as a robustness/error-message fix — no comment claims it closes a security hole.

**S-3 (LOW) — fixed.** Verified the action at the pinned commit myself: `src/run.ts` calls `restoreCache()` (:237) and `plugins.install()` (:239) before the `install-only` early return (:247-248), and `postRun()` calls `saveCache()` unconditionally (:261); `src/cache.ts:96` sets `GOLANGCI_LINT_CACHE` to `$HOME/.cache/golangci-lint`. Comment corrected; the `install-mode: binary` warning left intact.

**S-4 (LOW) — fixed.** Confirmed GPL-3.0 at `v2.13.2` and absence from `go.mod`/`go.sum` (no `tools.go`). Added a short **Scope** line to the policy: it governs *distributed* dependencies; build-time-only tools are out of scope — with both caveats (never vendor/redistribute it; `golangci-lint custom` would produce a combined GPL-3.0 binary that must not be shipped). Policy structure otherwise untouched.

**S-2 — not fixed, as directed.** Left as the already-accepted trade-off.

Also updated the PR body: the "Identical on every axis" equivalence claim and the `1m`-default timeout claim are both corrected there, and the linter count is now 5.

### Verification

```
make build                    exit=0
make lint                     exit=0   ==> gofmt check / golangci-lint v2.13.2 / 0 issues.
make test                     exit=0   49 packages ok, 0 FAIL
golangci-lint config verify   exit=0
```

Still zero `.go` files in the diff (`git diff --cached --name-only | grep '\.go$'` → empty).

One note on `make test`: the first run failed, and the evidence pins it to the **known pre-existing `internal/mcp` flake** rather than anything here. On the clean re-run `internal/mcp` was the *only* package to execute fresh — **215.9s against the 240s cap** — while the other 48 came from cache. Go never caches failed results, so those 48 had passed in run 1 and `internal/mcp` was precisely the failure. Run 1 overlapped my concurrent lint runs (load avg ~15), which tipped it over. This matches the 219s/240s flake already documented in the PR body and cannot originate from this commit, which changes no Go code.

usk6666 (commented):
## Security Re-Review (Round 2): APPROVED ✅

The blocking finding (S-1) is genuinely resolved — verified with an independent three-way fixture, not from the config text. The restructured `Makefile` lint gate fails closed on every path I could construct, including two shell-injection attempts, and every factual claim the fix commit added checks out against primary sources. One LOW cosmetic defect in a diagnostic branch; not blocking.

### Round-1 Finding Disposition

| ID | Disposition | Evidence |
|----|-------------|----------|
| **S-1** | **RESOLVED** | Independent fixture in `/tmp` (unused func + unused field + unused test helper). **Control A** — `staticcheck v0.7.0` (the binary this PR replaces, installed locally): 3× `U1000`, exit 1. **Control B (negative)** — pre-fix config `961785d4`: `0 issues.`, exit 0, confirming the regression was real. **C** — fixed config at `59b3042c`: 3 findings tagged `(unused)`, exit 1. Full parity including the `_test.go` case (`--tests` defaults true), so the gocyclo `_test.go` exclusion correctly does not leak into `unused`. |
| **S-2** | **ACCEPTED (unchanged, by design)** | Informational only, as agreed. `install-mode: goinstall` would reintroduce the go1.26 / `GOTOOLCHAIN=local` break — confirmed: golangci-lint `v2.13.2`'s `go.mod` declares `go 1.26.0`, CI pins `1.25.x`. |
| **S-3** | **RESOLVED** | Verified against the action source at the pinned commit `ba0d7d2e`: `run()` calls `restoreCache` (src/run.ts:237) and `install().then(plugins.install)` (:239) **before** the `install-only` early return (:247-249), and `postRun()` calls `saveCache()` (:261). The corrected comment is accurate. |
| **S-4** | **RESOLVED** | `GPL-3.0` confirmed via GitHub API; `grep -i golangci go.mod go.sum` → absent. CLAUDE.md now scopes the policy to *distributed* dependencies and records both correct caveats (never vendor; never ship a `golangci-lint custom` build). |
| **S-5** | **RESOLVED** | Injection empirically blocked. `v2.13.2'; touch /tmp/usk1043-PWNED; echo '` and `v2.13.2 $(touch ...)` both rejected at `Makefile:84`; no side-effect file created. `make` aborts the recipe at the failing line, so `pinned='$(GOLANGCI_LINT_VERSION)'` (`Makefile:100`) is unreachable with unvalidated content — **including under `make -k`**. `:=` + `$(shell)` is not re-expanded, so there is no parse-time injection either. |

### Fail-closed matrix (all verified by execution against the real `Makefile`)

| Input | Result |
|---|---|
| valid `v2.13.2` | pass, `0 issues.` |
| missing / empty / CRLF / garbage / two-line version file | **fail closed**, each with a named diagnostic |
| quote-breakout & `$(...)` injection | **fail closed**, no execution |
| shape-valid but version mismatch (`v2.99.0`) | **fail closed** with install hint |
| binary absent from PATH | **fail closed** |
| unknown linter / malformed `.golangci.yml` | **fail closed** (golangci-lint exit 3) |

Confirmed the fix comment's claim that the check deliberately lives in the recipe rather than a parse-time `$(error)`: `make clean` and `make vet` still succeed with a missing **or** malformed version file.

Also confirmed: `vX.Y` without a patch component is **intentional, not a weakening** — the action's `versionRe = /^v(\d+)\.(\d+)(?:\.(\d+))?$/` (src/version.ts:15) is the same shape as the recipe's `^v[0-9]+\.[0-9]+(\.[0-9]+)?$`, and it still fails closed at the version comparison. `.gitattributes` is correctly scoped: `git check-attr -a` sets `text`/`eol=lf` on `.golangci-lint-version` only and returns **no** attributes for `Makefile`, `go.mod`, `.golangci.yml`, `web/package.json`, `internal/session/session.go`, or `ci.yml` — no side effects.

### Supply-chain fundamentals re-confirmed

- **Action pin correct.** `git/ref/tags/v9.3.0` → `d583c34f…` type `tag` (annotated); dereferenced via `git/tags/d583c34f…` → commit **`ba0d7d2ec06a0ea1cb5fa41b2e4a3ab91d21278a`**, exactly the pinned value. All 19 `uses:` in `ci.yml` are SHA-pinned.
- **Permissions unchanged and minimal**: top-level `contents: read`, no job-level escalation, no `${{ }}` interpolation anywhere in the `lint` job. `scorecard.yml` untouched — no regression.
- `.claude/settings.json` is a **net reduction** in the Bash allowlist (`gocyclo:*`, `ineffassign:*` removed; `golangci-lint:*` added; no stale `staticcheck:*` left behind).

### Spot-check of the new comment text

Given this PR's failure mode was confidently-stated-but-wrong claims, I verified each added assertion against a primary source. All correct:

- `run --help` on v2.13.2 prints *"Timeout for total work. Disabled by default"* — the old "1m default" claim was indeed v1's; the correction is right.
- `gocyclo -top 3 -ignore '_test\.go$' tools/depsec` → `14 main fetchAlerts …/tools/depsec/main.go:105:1` — matches the comment exactly, including the line number.
- 137 files carry `//go:build e2e`, and **all 137** are `_test.go` — the "not a gocyclo regression" claim holds.
- `tools/depsec` is a genuine nested module (`module depsec`), so the blind spot is structural; the `depsec-tool` job does gate it with gofmt + vet + test.
- The staticcheck `checks:` exclusion list is load-bearing: `--no-config` on this repo yields exactly **14** staticcheck findings (plus 50 errcheck), matching the stated figure.

### New Findings

- **ID: S-6**
  **Severity:** LOW
  **File:** `Makefile`
  **Line:** 92-93
  **Category:** CIIntegrity
  **CWE:** CWE-1078
  **Description:** The CRLF diagnostic branch uses `grep -Eq '^v…$'`, which is line-oriented, so the anchors match *any single line* rather than the whole input. A two-line file (`v2.13.2\nv9.9.9`) therefore takes the CRLF branch and reports *"it has CRLF line endings; re-check it out with LF"* — a misdiagnosis. Reproduced. Secondary nit on the same block: the trailing *"CI enforces the same shape"* is slightly over-broad for CRLF specifically, since the action `.trim()`s the file content (src/version.ts:109) before applying `versionRe`, so CI would actually accept a CRLF file that `make lint` rejects.
  **Impact:** Cosmetic only. **The gate itself is unaffected** — the `[ "$lines" -ne 1 ]` test on line 85 short-circuits first, so multi-line input is still rejected with exit 1. The asymmetry is in the fail-safe direction (local stricter than CI), and `.gitattributes` makes the CRLF case unreachable in practice. Worst case is a developer chasing line endings when the real defect is a stray second line.
  **Remediation:** Optional. Add a `[ "$lines" -ne 1 ]` branch ahead of the CRLF `elif` to report "it has more than one line", or match with `grep -Ecq`/a whole-input comparison instead of a line-oriented match.

### Stats

| Severity | Count |
|---|---|
| CRITICAL | 0 |
| HIGH | 0 |
| MEDIUM | 0 |
| LOW | 1 |

`make lint` on the branch: **`0 issues.`, exit 0** — with `unused` enabled and **zero** suppressions (no `nolint` added by this PR; the only `exclusions` rule remains the pre-existing gocyclo/`_test.go` one). The fix restored the check without over-correcting. Still zero Go source changes.

---

PR #110, opened 2026-09-13
feat(cli): docs subcommand

usk6666:
## Summary

- **`yorishiro-proxy docs [topic] [section]`** — a new offline subcommand (`cmd/yorishiro-proxy/docs.go`) that prints the embedded help corpus without contacting an MCP server. USK-1036 (#106) made the same documents reachable as an MCP tool and via `client docs topic=X`, but both need a running server. This closes the CLI-driven / server-not-running case, the last M50 gap.
- **Index rendered locally from `mcp.DocsIndex()`**; documents and sections come back byte-identically from `mcp.DocsDocument` / `mcp.DocsSection`. **No symbol in `internal/mcp` was exported or changed** — only `resources/help_docs.md` gained a section. Only the index *header* differs per surface, and it must: the MCP header instructs the reader to `Call docs(topic="X")`, which an offline CLI user cannot do.
- **Leading positionals are stripped before `flag.Parse`.** Go's `flag` stops at the first non-flag argument, so `docs macro --section "X"` — the Issue's own headline example — would otherwise silently drop the flag with no error. Positional-section + `--section` is an error rather than a silent preference; a `key=value`-shaped topic gets a grammar hint pointing at the bare-word spelling.
- **19 `clientToolHelp` entries** now point at `yorishiro-proxy docs <tool>`, guarded by a fast-tier invariant so a partially-applied edit cannot pass silently. `macro`'s "use JSON input or MCP client" dead-end is replaced with a pointer naming the `§var§` syntax; `grpc_schema`'s MCP-idiom pointer is brought into CLI idiom; `printClientUsage` gains an offline-docs trailer.
- **`help_docs.md` gains `## From the CLI`** so the `yorishiro-proxy docs docs` self-reference is truthful. Plus a `pentest-proxy` SKILL.md Quick Reference row and a CLAUDE.md Package Layout drift fix (`serverjson.go` / `browser.go` moved to `internal/mcpserver/` back in #723).

### Three things a reviewer should know

1. **`yorishiro-proxy docs macro` currently starts a proxy server.** `run()` has no `default:` case and `mcpserver.Run` never inspects `fs.Args()`, so today both words are silently ignored and the server boots. Adding `case "docs":` closes that footgun for the `docs` word specifically. Adding a general `default:` is a behavior change, deliberately left to a follow-up.
2. **`internal/mcp` was already linked into this binary** via `internal/mcpserver` (`init.go`, `server.go`), which `main.go` and `client.go` already import. The new import adds zero bytes, zero dependencies and zero init cost (`docsRegistryOnce` is `sync.OnceValue`; the package has no non-test `init()`). **"Works offline" means *does not open a socket*, not *slim binary*.**
3. **`docs --help` exits 0 to stdout**, diverging from `install` / `upgrade`, which let `flag.ErrHelp` escape to `main` and exit 1 with `error: flag: help requested`. A *documentation* subcommand that exits non-zero when probed with `--help` teaches an agent the command is unsupported — self-defeating for the one command whose job is discoverability. The other subcommands are deliberately **not** changed here; unifying them is deferred.

The Output SafetyFilter is deliberately **not** applied, inheriting USK-1036's decision (`help_security.md` documents the PII presets with sample card numbers the credit-card preset would mask).

## Test plan

- [x] `make build`
- [x] `make lint` (gofmt, go vet, staticcheck, ineffassign, gocyclo) — clean
- [x] `make test` — fast tier green; 40 new subtests in `cmd/yorishiro-proxy/docs_test.go`
- [x] `make test-e2e-smoke` — green, including the new `TestBinarySmoke_DocsSubcommandWorksOffline`
- [x] Arg-grammar table test covers every row: `docs` / `docs macro` / `docs macro "Variable substitution syntax"` / `docs macro --section "…"` / `docs macro -section "…"` / `--section "…" macro`, plus both conflict cases (section twice, three positionals) and section-without-topic
- [x] Offline oracle is three-layered: fast-tier `runDocs` with no env/network/fixtures; exec'd binary with `HOME=t.TempDir()` (no `server.json` auto-discovery) **and** a bound-then-closed `YP_CLIENT_ADDR` + bogus `YP_CLIENT_TOKEN` — inert today, red the moment `docs` starts dialling
- [x] `--section` is asserted to actually *narrow* the output, which is what catches `flag` silently dropping it
- [x] Unknown topic → exit 1, every valid topic on stderr (via the `*exec.ExitError` idiom); unknown section → the H2–H4 outline
- [x] `TestHelpResources_HeadingDepthWithinOutline` and `TestDocsTool_IndexListsEveryTopic` still pass after the new H2
- [x] **Acceptance criterion**: with no server running, `yorishiro-proxy docs macro` and `yorishiro-proxy docs macro --section "Variable substitution syntax"` both reach the `§name§` syntax (`TestRunDocsCLI_AcceptanceCriterion`, named after USK-1036's `TestDocsTool_AcceptanceCriterion`)

Resolves USK-1037
Linear: https://linear.app/usk6666/issue/USK-1037/featcli-docs-subcommand


usk6666:
## Automated review gate: APPROVED (0 fix rounds)

| | Verdict | Findings |
|---|---|---|
| Code Review | **APPROVED** | 0 |
| Security Review | **APPROVED** | 1 LOW, accepted as-is |

Two parallel worktree-isolated reviewers plus independent verification against the built binary.

### Code review

- **Argument grammar traced by hand, all eight shapes correct.** The critical one: `docs macro --section X` genuinely reads the flag. Go's `flag` package stops parsing at the first non-flag argument, so the obvious implementation would have dropped `--section` silently and printed the whole document — breaking the Issue's own headline example with no error. `splitDocsArgs` strips leading positionals before `Parse`, mirroring `parseInstallTarget` (`install.go:84-102`).
- A positional section **plus** `--section` is an explicit "section given twice" error, not a silent preference. Three positionals error with a copy-pasteable fix.
- `runDocs` contains no `os.Exit`; `main.go:20` remains the only one in `cmd/`.
- `renderDocsCLIIndex` derives every topic, description, est_tokens and section list from `mcp.DocsIndex()` — **no hardcoded topic list**. The PR also *removes* a hand-maintained 22-topic enumeration from `clientToolHelp["docs"]` rather than guarding it.
- All 19 `clientToolHelp` pointer lines landed. `macro`'s dead end was **replaced** (keeping the `query resource=macros` continuation), and `grpc_schema`'s pre-existing MCP-idiom pointer was rewritten in place, not duplicated.
- Offline oracle is as designed: `HOME=t.TempDir()` (no `server.json` auto-discovery) plus deliberately-inert `YP_CLIENT_ADDR`/`YP_CLIENT_TOKEN` at a closed port, so the test goes red if a refactor ever routes `docs` through the connection path. Non-zero case uses the `*exec.ExitError` idiom. No `t.Logf` violations.

### Security review

Path traversal is **structurally impossible**: the topic is a key into a compile-time `map[string]string`, never joined into a path, so `docs ../../etc/passwd` is a map miss.

**S-1 (LOW, accepted as-is)** — `docs.go:155-156` and `:228-231` interpolate an argv fragment with `%s` rather than `%q`. Reproduced, then deliberately not fixed: the same message already prints the topic `%q`-escaped immediately before the raw fragment, so the neutralization is present, and the raw rendering is an intentional copy-paste affordance. `%q` would degrade the primary hint to ``did you mean 'yorishiro-proxy docs "macro"'?`` — the exact string this milestone exists to make usable. argv is operator-controlled; no attacker in the loop. **Re-open trigger:** if `docs` argv ever becomes non-operator-controlled.

**One reviewer finding rejected as a false positive.** The security agent flagged the `CLAUDE.md` hunk for dropping `serverjson.go` / `browser.go`, having checked global existence rather than location. Both moved to `internal/mcpserver/` in #723; the edited block is the `cmd/yorishiro-proxy/` listing, which now matches `ls` exactly.

### Verification

`lint` / `build` / `test` / `test-e2e-smoke` — **all green on CI**.

For the record: `make test` exited non-zero on both *local* review runs (`panic: test timed out after 4m0s`), which was isolated to the 4-minute **per-package** budget under machine contention (4 cores, load average 5–12, ~8 concurrent worktrees) — different packages timed out each run and no test failed on its merits. CI has now confirmed this: `test` passes in 4m04s. The thin headroom is pre-existing and unrelated to this PR; tracked as USK-1048.

### Related issues filed during this work

- **USK-1047** — `main.go`'s switch has no `default:`, so *any* unknown subcommand silently starts a proxy server. This PR closes it for the word `docs`; the general fix is deferred.
- **USK-1046** — `.claude/skills/pentest-proxy/SKILL.md`'s `fuzz_http` example does not match the real schema on any of three axes. Pre-existing; this PR only adds a Quick Reference row and does not touch the broken example.
- **USK-1048** — fast-tier per-package timeout headroom.

PR #111, opened 2026-09-14
fix(session): treat upgrade-pending ctx-cancel as ErrUpgradePending in clientToUpstream

usk6666:
## Summary
- `clientToUpstream`'s `dispatchClientAction` error exit now defers to a pending upgrade instead of treating the handoff's ctx-cancel as a genuine failure, so the cascade-close defer no longer fires during an upgrade swap.
- Extracted `upgradeExitOrErr` (internal/session/session.go) so the "upgrade-pending takes precedence" policy and its ordering/synchronisation rationale are documented once, and `clientToUpstream` stays under the gocyclo-15 gate.
- Added `TestClientToUpstream_DispatchErrorVsUpgradePending` — a deterministic, table-driven, bidirectional regression test that drives the real `clientToUpstream`.

## Root cause
`errgroup.WithContext` cancels the shared ctx on the FIRST non-nil error, and `ErrUpgradePending` — the signal for a perfectly normal SSE/WS upgrade — is non-nil. When that cancel lands while `clientToUpstream` is inside `dispatchClientAction` → `upstream.Send`, Send returns `context canceled`. The `if perr != nil { return perr }` exit had no upgrade-pending check (unlike the `client.Next` exit right above it), so the handoff artifact was misclassified as a genuine error. The cascade-close defer then called `uh.ch.Close()` on a channel with `sentEnd=false / recvEnd=false`, which on HTTP/2 emits `RST_STREAM(CANCEL)` — killing the still-live SSE response body — while simultaneously stamping `markTerminated(io.EOF)`. The detach drain starts moments later, finds only the already-buffered events, reads `Err() == io.EOF`, closes the pipe cleanly, and the truncation surfaces as a graceful EOF with `state="complete"`. Every layer reports success; the client just silently receives fewer events.

Ordering is safe: the peer latches the notice BEFORE returning `ErrUpgradePending`, and the errgroup cancel happens after that return. `UpgradeNotice`'s accessors are mutex-guarded, so the cross-goroutine read is properly synchronised.

## Test plan
- [x] Deterministic regression test, both directions (upgrade-pending → returns `ErrUpgradePending` and the upstream channel is NOT closed; dispatch error without a pending upgrade → error propagates and the cascade close still fires). Verified to FAIL on the pre-fix tree and PASS after.
- [x] `make lint` / `make build` / `make test` / `make test-e2e-smoke`
- [x] Timing-widened SSE-over-h2 probe (throwaway, not committed): **4 truncations / 300 runs before the fix, 0 / 300 after**.

Resolves USK-1042
Linear: https://linear.app/usk6666/issue/USK-1042


usk6666 (commented):

Minimal, correctly-scoped concurrency fix. `UpgradeNotice`'s accessors are genuinely mutex-guarded and the claimed happens-before edge (latch → `ErrUpgradePending` return → errgroup cancel → cancel observed in `Send` → `Pending()` read) holds, so `upgradeExitOrErr` is race-free and cannot be *missed* under the documented interleaving. I traced all four swap paths (`runUpgradeSSE`, `runUpgradeSSEOverH2`, `runUpgradeWS`, `runUpgradeWSOverH2`) for the "parent ctx already dead at swap entry" case that this change makes newly reachable at the dispatch branch, and confirmed no connection, goroutine, or reader leak. Two LOW notes below; neither is exploitable and neither blocks merge.

### Verification performed

**Leak analysis (CWE-404 / CWE-772) — implementer's claim confirmed.**
- `runUpgradeSSEOverH2`: `uClose`, `cW.Close`+`cClose`, and `ReleaseStreamSubStack` defers are all registered **before** `driveSSEEventLoop`, which returns on its leading `ctx.Err()` check. `ReleaseStreamSubStack` (`internal/connector/connection_stack.go:234-259`) closes the registered `sse` adapter. Nothing survives.
- `runUpgradeSSE` (h1): `defer upBody.Close()` and `defer clientCloser.Close()` registered before the loop; conn ownership moved to `upBody` at detach, so the h1 conn is closed. `http1.Layer.DetachStreamingBody`/`DetachStream` take no ctx, so a dead ctx introduces no new pre-defer failure mode.
- `runUpgradeWSOverH2`: deferred `ReleaseStreamSubStack` closes both `ws.Layer`s (→ `detachCloserAdapter` → `cClose`/`uClose`); `pullWSChannels`/`assertSubStackLookups` failures land after that defer.
- `runUpgradeWS`: recurses into `RunStackSession` → `RunSession`, which on a non-`ErrUpgradePending` result runs both `client.Close()` and `uh.ch.Close()` defers. `keepaliveStop()` deferred.
- `sse.Wrap` (`internal/layer/sse/layer.go:70`) spawns no goroutine — no orphaned reader.

**Resource-pinning primitive (remote-triggerable) — not present.** `trySetPending` fires only from `processUpgradeReceive`, i.e. only on a **Receive**-direction envelope, so a latched notice implies the upstream was already dialed and answered. Consequences: (a) the `upstreamCh == nil` branch in `RunStackSessionExchange` is unreachable via this new path, and (b) keeping the upstream open when a notice is latched *is* the correct semantic — ownership transfers to the swap orchestrator, whose lifetime is bounded by ctx cancel / stream end / relay error. A hostile upstream returning `text/event-stream` to pin a connection into an SSE relay is pre-existing (`isSSEResponse` already latched before this PR) and unchanged.

**Caller survey.** All production entry points (`internal/proxybuild/builder.go:727,879,1131`, `tcp_forward.go:507,612`, `tcp_forward_l7.go:232,325`, `tcp_forward_h2.go:345`) go through `RunStackSession`/`RunStackSessionExchange`, which own the `UpgradeNotice` and handle `ErrUpgradePending`. There is no direct `RunSession` production caller that could receive `ErrUpgradePending` and leave both wires unclosed.

**Test-side `-race` safety.** `sendHook` and `sendCalls` are written and read only from the single test goroutine (`clientToUpstream` is called synchronously; `mockChannel.Send` runs inline). `sendHook` is set before first use, matching the existing unguarded `sendErr`/`blockNext` pattern. No new race surface.

**Build/test.** `gofmt`, `go vet`, `staticcheck`, `gocyclo -over 15` clean on `internal/session`. `internal/session` `-race -count=3` green; `internal/proxybuild`, `internal/connector`, `internal/layer/sse`, `internal/layer/ws` `-race` green. Diff is exactly the two declared files — no scope creep, no dependency changes.

### Low-severity notes (non-blocking, no action required)

| S-1 | LOW | internal/session/session.go | 2518-2523 | GoSecurity | CWE-755 | `upgradeExitOrErr` suppresses *every* dispatch error once an upgrade is latched, including a genuine `*layer.StreamError`, dropping `ClassifyError`'s `FailureReason` from the session result |
| S-2 | LOW | internal/session/session.go | 2605-2612 | GoSecurity | CWE-362 | The guard lets `clientToUpstream` win the errgroup error race, so the swap can proceed before `upstreamToClient` delivered the 101/200 to the client (Friction 2-C ordering) |

**S-1.** When a genuine `upstream.Send` failure coincides with a latched notice, the original error is replaced rather than joined, so the session result loses the `layer.ErrorCode` that `ClassifyError` projects into `flow.StreamUpdate.FailureReason`. In practice the failure re-surfaces in the post-swap relay (`driveSSEEventLoop` / `runWSOverH2Relay`) and still projects `state="error"`, so this is a diagnostic-fidelity nit rather than a new silent-success path in the opposite direction. A narrower guard keyed on `errors.Is(perr, context.Canceled)` would avoid it, but it is *less* robust: only the h2 path returns a bare `ctx.Err()` (`internal/layer/http2/channel.go:798-822`); other Channel implementations may wrap the cancel, in which case the narrow form would miss the very case this PR fixes. The unconditional form also keeps symmetry with the pre-existing `client.Next` branch. I consider the current shape the right trade-off — flagging only for the record. (`errors.Join(ErrUpgradePending, perr)` would preserve both, but that touches the errgroup error signalling the Issue explicitly deferred.)

**S-2.** If `clientToUpstream` hits a genuine dispatch error while the notice is latched but before `upstreamToClient` finishes `client.Send(101)`, it now returns `ErrUpgradePending` first, the errgroup cancel aborts the peer's in-flight `client.Send`, and `g.Wait()` still yields `ErrUpgradePending` — so the swap runs even though the client never received the upgrade response. The observable result is a client-side protocol error (raw WS frames / h2 DATA on a stream with no HEADERS), not data exposure or a leak, and the swap tears both wires down through the defers verified above. This is **not newly introduced as a class**: the pre-existing `client.Next` upgrade-pending branch (session.go:2573) has exactly the same property, and the PR extends it to one more branch symmetrically. Reaching it requires an attacker to both force a genuine dispatch error and win a tight race against the peer's return.

---

usk6666 (commented):
## Code Review: APPROVED ✅

Minimal, correctly-scoped one-line fix whose three load-bearing claims (latch-before-return ordering, mutex-guarded `UpgradeNotice` accessors, gocyclo-15 justification for the helper) all check out against the source. The regression test was empirically proven to be genuinely bidirectional: it fails on the pre-fix tree AND fails on two plausible over-corrections, which is exactly what USK-902 failed to do.

### Verification performed

**Test efficacy (3 local experiments, all reverted — nothing committed or pushed):**

| Experiment | Result |
|---|---|
| Revert `return upgradeExitOrErr(notice, perr)` → `return perr` | Row A **FAILS** — `err = upstream.Send: context canceled`, `Close calls = 1` (reproduces the exact RST_STREAM(CANCEL) cascade from the Issue) |
| Delete the cascade-close defer entirely | Rows B **and** C **FAIL** (`upstream closed = false, want true`) |
| Over-correct: defer also skips `context.Canceled` | Row B **FAILS** |

So the answer to "would the test still pass if the cascade-close defer were deleted?" is **no** — the test does the job the Issue requires. It drives the real `clientToUpstream` / `dispatchClientAction` / real cascade defer; only the peer's *latch* and the *ctx cancel* are modelled, which is the environmental precondition, not the producer simulation USK-902 was rejected for.

**Ordering/synchronisation claim (doc comment) — confirmed, not overstated:**
- All three production `trySetPending` call sites are in `processUpgradeReceive` (`upgrade.go:444/450/457`), reachable only from `UpgradeStep.Process` on `Direction == Receive`, i.e. only from `upstreamToClient`'s `runPipelineTracked` — which runs strictly *before* its `if upgradePending(notice) { return ErrUpgradePending }`.
- `trySetPending` and `Pending()` both take `n.mu` (`upgrade.go:168`, `:175`).
- Corollary worth recording: because `pending != ""` implies the peer goroutine latched it and is already returning, suppressing the cascade close here cannot strand a peer parked in `upstreamNext`.

**Self-reported behaviour change (dead parent ctx → swap path) — traced, no leak:**
- Both SSE paths terminate in `driveSSEEventLoop`, which checks `ctx.Err()` at the top of every iteration and returns immediately (`session.go:1759-1761`). The `uClose` / `cW.Close`+`cClose` / `ReleaseStreamSubStack` / `upBody.Close` / `clientCloser.Close` defers are all registered before that call and do run.
- Both WS paths end in a recursive `RunStackSession(ctx, …)`; with a dead ctx the inner `RunSession` returns `context.Canceled` (not `ErrUpgradePending`), so *its* deferred `client.Close()` / `uh.ch.Close()` fire normally. `keepaliveStop()` also runs.
- Note this state was already reachable pre-fix whenever `upstreamToClient` won the race to report the first errgroup error, so the PR introduces no new leak class.

**Symmetry with the sibling exit:** the helper is correctly *not* applied to the `client.Next` branch — that branch must run `checkInterceptReleaseEOF` and map `io.EOF → nil` *after* the upgrade check, which `upgradeExitOrErr`'s signature cannot express. The split is deliberate and justified, not missed unification.

**gocyclo claim: verified.** `Makefile:76-77` really does gate on `gocyclo -over 15` (CLAUDE.md's description of `make lint` omits it). `clientToUpstream` currently measures exactly **15**; an inline `if` would make it 16 and break `make lint`. The helper is warranted, not gratuitous abstraction.

**`mockChannel.sendHook`:** opt-in behind a nil check, no existing test sets it, and the `Send(_ ctx …)` → `Send(ctx …)` rename is inert. `sendCalls++` runs on the test goroutine (synchronous `clientToUpstream` → `dispatchClientAction` → `Send` chain, no background goroutines in this test), and `cancel()` from inside the hook is safe/idempotent. Race-clean.

**Scope discipline: clean.** Only the 2 expected files; `internal/layer/http2/` untouched; no logging/observability added; no TODO/FIXME/HACK; no errgroup refactor.

**Commands run:** `go test -race -count=3 ./internal/session/` ✅ · `./internal/proxybuild/ ./internal/pipeline/` ✅ · `go vet` ✅ · `staticcheck` ✅ · `gofmt -l` clean.

### Findings

| F-1 | NIT | internal/session/session.go | 2499-2521 | CodeHealth | `upgradeExitOrErr`'s doc explains the ctx-cancel case but doesn't acknowledge that it discards *any* `err` when a notice is latched |

**F-1 (NIT)** — `upgradeExitOrErr` returns a bare `ErrUpgradePending` and drops `err` on the floor, so a genuinely broken upstream that fails `Send` at the same instant an upgrade latches is also reported as a clean handoff. This is fine in practice and is the policy the Issue prescribed (it mirrors the `client.Next` branch): the `dial` sub-path is unreachable with a latched notice (the notice only latches on a `Receive`, which requires `uh.ch` to already exist), and the swap degrades gracefully because `DetachStream` / the read loop surfaces the real failure. The doc comment currently frames the input only as "a wire operation in flight [that] fails with context canceled", which understates the guard's breadth.

*Suggestion (doc-only, no behaviour change):* add a sentence such as "Note this deliberately discards err in full, not just context.Canceled — a genuine failure co-occurring with a latched upgrade is reported as the handoff; the swap path surfaces the real fault instead." Explicitly **not** suggesting logging here, per the Issue's deferral of observability work.

---

PR #112, opened 2026-09-14
fix(connector): force-close raw TLS conns in driveSSEOverH2ThroughProxy to deflake HappyPath

usk6666:
## Summary
- `driveSSEOverH2ThroughProxy` (`internal/connector/sse_over_h2_integration_test.go`) relied on
  `tr.CloseIdleConnections()` alone to make the proxy observe the client h2 connection close. That
  races with `xhttp2.Transport`'s internal "is this conn idle?" bookkeeping: under CPU contention
  the response reader may not have released the conn to the pool yet, so the proxy's
  `clientL.Channels()` never observes a remote close, `onHTTP2Stack` never exits its for-select
  loop, and its deferred `wg.Done()` never fires — `waitSessionDone` then trips its 15 s deadline.
- Ported the workaround already documented in
  `TestFullListener_CONNECT_SSE_OverH2_SiblingNonInterference`
  (`sse_over_h2_sibling_integration_test.go:47-59` / `:183-189`): capture the underlying TLS conns
  returned by `DialTLS` and force-close them, which makes the client-side conn close deterministic.
- Registered as a `defer` (after the `tr.CloseIdleConnections()` defer, before the
  `resp.Body.Close()` defer) so the LIFO teardown order is body close → force-close raw conns →
  `CloseIdleConnections`, matching the sibling's explicit ordering, and so every early-return path
  in the helper is covered.
- Test-only change. No production code touched. The helper's read-loop oracle is deliberately left
  as-is (tracked separately as USK-1045).

Affects both callers of the helper: the smoke-tier `TestFullListener_CONNECT_SSE_OverH2_HappyPath`
(`//go:build e2e`, per-PR merge gate) and the exhaustive-tier
`TestFullListener_CONNECT_SSE_OverH2_FrameRecording`.

## Test plan
- [x] `gofmt -l` clean / `make lint` / `make build` / `make test` / `make test-e2e-smoke` all pass
      (`..._HappyPath` and `..._SiblingNonInterference` both PASS, 0 FAIL lines)
- [x] Contended repeat runs — **the original ~1/600 flake did not reproduce**, stated plainly:
      - pre-fix: 1600 iterations, 0 failures (400 @ `GOMAXPROCS=2` + 4 burners; 600 @ `GOMAXPROCS=2`
        + 16 burners, load avg ~29; 600 @ `GOMAXPROCS=1` + 16 burners)
      - post-fix: 1200 iterations at the two heavy profiles, 0 failures
- [x] Because the race never landed, the causal chain was proved **deterministically** instead
      (temporary local test, deleted and never committed): with the client h2 conn deliberately left
      open the session stayed blocked past 5 s — i.e. heading for `waitSessionDone`'s 15 s
      `t.Fatal` — and force-closing the captured raw TLS conn released it in **372 µs**. That is
      exactly the property this fix supplies, independent of the transport's idle bookkeeping.

Indicative side effect (contention varied between runs, so treat as directional only): 600
iterations of the connector package took 124.9 s pre-fix vs 69.0 s post-fix, consistent with
removing real per-iteration teardown latency.

Resolves USK-1044
Linear: https://linear.app/usk6666/issue/USK-1044


usk6666 (commented):

Test-only change (+39/-0 in a single `//go:build e2e` file, no production code, no dependency movement); the new `h2ConnsMu`/`h2Conns` capture is correctly mutex-guarded on both the write (`DialTLS`) and read (deferred close loop) side, and the force-close is safe for double/concurrent invocation. No security findings.

### Checks

1. **Concurrency (CWE-362)** — Every access to `h2Conns` is under `h2ConnsMu`: the `append` inside `DialTLS` (`:196-198`) and the ranged read in the deferred close (`:212-218`). No other reference to the slice escapes the helper, so even if `xhttp2.Transport` invokes `DialTLS` concurrently (retry/redial) or after the helper returns via the still-running `RoundTrip` goroutine, the slice itself is race-free and `-race` clean.
2. **Resource handling** — The `defer` closes every conn captured before it runs, on all return paths (early `status != 200` / Content-Type / body-read-error returns included), which is strictly better coverage than the sibling's inline post-`grp.Wait()` call. Double-close is reachable (force-close loop → then the deferred `tr.CloseIdleConnections()`, or a concurrent `readLoop` teardown) and is harmless: `crypto/tls.Conn.Close` interlocks on `activeCall` via atomic CAS and returns `net.ErrClosed` on the second call — no panic, no race — and the test correctly discards it with `_ =`. The only uncovered window is the `h2 GET: timeout waiting for headers` path, where the orphaned `RoundTrip` goroutine could append a conn after the defer already ran; that path already returns an error that both callers turn into `t.Fatalf`, so no fd accumulates across repeated passing iterations. Not a finding (and the pre-existing `resp` leak on that same path is unchanged by this PR).
3. **Nothing unintended committed** — Branch is exactly one commit ahead of `main`; `git show --stat` and `gh pr diff` both show only `internal/connector/sse_over_h2_integration_test.go`. No scratch files, debug probes, key material, or credential-like strings in the diff; working tree clean; `gofmt -l` clean.
4. **Dependencies** — `go.mod` / `go.sum` untouched. The single new import is stdlib `sync`. No license exposure.

Per scope guidance, the pre-existing `InsecureSkipVerify: true` test scaffolding, the unset upstream `MinVersion`, and the helper's weak read-loop oracle (USK-1045) were reviewed and deliberately not flagged.

Note on verification: this worktree has no `internal/mcp/webui/dist/`, so per CLAUDE.md no bare `go` command was run; compile/behaviour verification is left to CI and the code-review agent. The analysis above is static plus stdlib source confirmation of `tls.Conn.Close` idempotency.

**FINDINGS: None** (CRITICAL 0 / HIGH 0 / MEDIUM 0 / LOW 0)

---

usk6666 (commented):
## Code Review: APPROVED ✅

Clean, minimal, correctly-scoped test-only change that faithfully ports the sibling test's documented workaround. All five load-bearing claims in the PR body were verified against source and empirically; the defer ordering, the locking, and the oracle-preservation all hold. No findings above LOW. One empirical note below: a **second, independent and pre-existing** flake in the same smoke-tier test was reproduced during review — it is not caused by this PR, but it means the merge gate is not fully deflaked yet.

### Verification performed

| Check | Result |
|---|---|
| `gofmt -l` on changed file | clean |
| `go vet -tags e2e ./internal/connector/` | clean (full tier compiles) |
| `go vet -tags "e2e e2e_smoke" ./internal/connector/` | clean (smoke tier compiles) |
| `staticcheck -tags e2e ./internal/connector/` | only a pre-existing `U1000` in `wss_h1_flyedge_integration_test.go`, unrelated |
| `-race`, HappyPath + FrameRecording + Sibling | ~1400 post-fix iterations, **0 DATA RACE** |
| `_FrameRecording` (exhaustive caller) + `_SiblingNonInterference`, `GOMAXPROCS=2` + 6 CPU burners | 200/200 and 200/200 PASS — force-close does not break frame assertions |
| Head commit contents | exactly 1 file, +39/-0; no probe/scratch/debug artifacts |

### Scrutiny points

1. **Defer ordering** — holds exactly. Registration is `tr.CloseIdleConnections` (L202) → force-close (L212) → `resp.Body.Close` (L244), so LIFO teardown is body close → force-close → `CloseIdleConnections`, as claimed. Every `return` in the helper body (L223/240/242/246/249/258/268/272/274) is after L212, so all early-return paths are covered. The `DialTLS` closure appends only after a successful handshake and closes `raw` directly on each failure path, so no conn is ever captured-but-unclosed there. The comment's sibling comparison is precisely worded and does not overclaim (the sibling's `CloseIdleConnections` runs at test exit, i.e. after `waitSessionDone`; here it runs just before the helper returns — the force-close / `CloseIdleConnections` *relative* order is identical in both).
2. **Concurrency** — `h2Conns` is accessed only under `h2ConnsMu` in both the append and the close loop; no data race, confirmed by ~1400 `-race` iterations. Only one `RoundTrip` is issued so at most one dial occurs, but the locking would be correct for concurrent dials too. See NIT-1 for the one lifetime-escape edge.
3. **Both callers** — correct for both. `onHTTP2Stack` (`h2_test_helpers_test.go`) calls `streamWG.Wait()` on both exit branches *before* the deferred `wg.Done()`, so `waitSessionDone` returning implies every per-stream `RunStackSessionExchange` — and therefore every `RecordStep` write — already completed. There is no race between the force-close and the Store reads the assertions perform. `_FrameRecording`'s assertions are all "at least one" / "exactly zero send-side" shaped and cannot be truncated by an earlier client-conn close; 200/200 PASS under contention confirms it.
4. **Oracle unchanged (USK-1045 correctly out of scope)** — confirmed. The force-close is a `defer`, so it can only execute after the read loop has already returned, and `resp.Body.Close()` runs strictly before it. It cannot change how many events the loop observes, nor convert a clean read into an error. The `nil`-on-EOF / per-`Read` `"\n\n"` counting semantics are byte-for-byte untouched.
5. **Honesty of verification claims** — accurate, not overclaiming. The head commit `6cfe465` is a single file at +39/-0 with no temporary test, probe, debug print, or scratch file committed. The "the original flake did not reproduce" framing matches my own experience: I could not reproduce the `waitSessionDone` timeout mode on pre-fix `main` either (0 occurrences in 1500 pre-fix iterations, including 300 at `GOMAXPROCS=1` with 6 CPU burners). Labelling the timing delta as "directional only" is appropriately hedged.

### Findings

| F-1 | LOW | internal/connector/sse_over_h2_integration_test.go | 71-73 | TestQuality | Second, **pre-existing** flake in the same smoke-tier test is not addressed — merge gate remains probabilistically flaky |
| NIT-1 | NIT | internal/connector/sse_over_h2_integration_test.go | 212-218 | CodeHealth | Captured conn can escape the force-close on the 15 s header-timeout path |

#### F-1 (LOW) — residual pre-existing flake, not a regression

While stress-testing this PR I hit a failure mode that is **not** the 15 s `waitSessionDone` timeout:

```
sse_over_h2_integration_test.go:73: expected at least one Stream with Protocol=sse
    after SSE-over-h2 swap; got streams=[0xc0001c65b0]
--- FAIL: TestFullListener_CONNECT_SSE_OverH2_HappyPath (0.06s)
```

The session completes quickly, but zero `Protocol=sse` envelopes reached `RecordStep`, so `maybeRetagProtocol` never fired. `driveSSEOverH2ThroughProxy` still returned `nil` because of the EOF-tolerant oracle.

I confirmed this is **pre-existing and not introduced by this PR** by A/B-ing under identical load (`GOMAXPROCS=2` + 6 CPU burners, `-race`):

- **pre-fix (`main` @ `1fd40c12`): 399 PASS / 1 FAIL out of 400** — same assertion (`:72` pre-fix = `:73` post-fix)
- **post-fix (this PR): 400 PASS / 0 FAIL out of 400**

So this PR is strictly not a regression, and it is correct to merge. But since USK-1044's stated goal is "stop this smoke-tier test from probabilistically blocking the merge gate", note that a second independent mechanism at roughly 1/400 under contention survives this fix. Structurally it also cannot be caught reliably today, because the weak oracle lets a 0-event EOF pass as success — which is exactly what USK-1045 is about.

**Suggestion**: no change to this PR. File a follow-up Issue for the `Protocol=sse` retag flake, with the reproduction above, and note the dependency on USK-1045 (strengthening the oracle to accumulate the body and require `>= numEvents`, as the sibling already does at `sse_over_h2_sibling_integration_test.go:144-169`, would convert this silent 0-event pass into a clear failure).

#### NIT-1 — conn capture can escape the force-close

On the `case <-time.After(15 * time.Second)` path (L241-243) the `RoundTrip` goroutine outlives the helper, so `DialTLS` can append to `h2Conns` *after* the force-close loop has already drained it, leaving that conn (and the goroutine and any later `*http.Response`) unclosed.

This is not a data race (the mutex covers both sides) and it cannot hang anything, because both callers `t.Fatalf` immediately on a non-nil error from the helper and therefore never reach `waitSessionDone`. The goroutine/response leak on that path also predates this PR. Purely cosmetic — flagged only for completeness; fixing it would mean setting a "closed" flag under the mutex and having `DialTLS` close-and-reject when set, which is more machinery than this teardown warrants.

---

PR #116, opened 2026-09-15
fix(config): validate tls_fingerprint at config load

usk6666:
## Summary

`tls_fingerprint` was the only enum-valued `ProxyConfig` field with no `Validate()` coverage (`forward.protocol`, `rotation.policy`, `safety_filter.*.action` and `intercept_queue.timeout_behavior` all have it). A config-file typo such as `"firefx"` therefore reached runtime unrejected: `InitTLSTransport` logged a warning and degraded to a Go-native ClientHello, so every `resend_*` / `fuzz_*` dial made before (or without) `proxy_start` went out un-spoofed while `query config` kept reporting `"firefx"` back to the agent.

For a tool whose entire purpose is fingerprint fidelity, that is a silent fingerprint lie on the resend axis.

## Changes

Add the check to `ProxyConfig.Validate()` (and, for symmetry, to `Config.Validate()`, whose tier-2 `TLSFingerprint` is still read by `InitTLSTransport`), and move the 6-name vocabulary into `internal/config` as the single source of truth:

- `config.ValidTLSFingerprint(string) bool` — case-insensitive, whitespace-trimmed, matching the normalization every consumption seam already applies (`UTLSProfileFor`, `tlslayer.clientUTLS`, `transport.ParseBrowserProfile`, `http2.resolveH2Fingerprint`).
- `config.TLSFingerprintNamesList()` — the shared list for the three site-specific error messages.

`internal/mcp` and `internal/mcpserver` drop their local whitelists and call the shared predicate; both also gain `TrimSpace`, so all three entry points (config file, `-tls-fingerprint` CLI flag, MCP `proxy_start` / `configure`) now accept exactly the same set.

## Design decisions (the two points the Issue flagged for design review)

1. **Case / whitespace → normalize, uniformly.** Previously `internal/mcp`'s `applyTLSFingerprint` lowercased but did not trim, so `" none "` was rejected at the MCP boundary while the config-file path accepted it. All three entry points now trim + case-fold, matching what the consumption seams already do.
2. **`InitTLSTransport` stays fail-soft.** Now that all three entry points reject unknown names before boot reaches it, the `Warn` + `StandardTransport` fallback is defence-in-depth rather than the primary guard. Making it fail-hard would change the signature to return an error and add a new boot-failure mode for an input that is now unreachable. Its godoc records this.

## Invariants preserved

Validation **rejects but never rewrites**: `ProxyConfig.TLSFingerprint` stays verbatim so `EffectiveTLSFingerprint` (H2 send-shape, `query config`) and `EffectiveUTLSProfile` (`tlslayer` dial, `poolKeyForH2`, `ALPNCacheKeyFromConfig`) keep observing the operator's exact bytes (USK-1021). `UTLSProfileFor`, `ResolveTLSFingerprint` and the `tlslayer` / `transport` parrot tables are untouched.

This is operator configuration input, not wire data, so MITM Principle #1 (do not normalize what the wire did not normalize) does not apply.

`Config.Validate`'s log-level / log-format checks moved into `validateLogging` to stay under the gocyclo threshold.

## Test plan

- `TestValidTLSFingerprint` / `TestTLSFingerprintNamesList` — the shared vocabulary predicate and error-message list.
- `TestProxyConfig_Validate_TLSFingerprint` / `TestConfig_Validate_TLSFingerprint` — both validation tiers, including the "unset means default" and trimmed/folded spellings.
- `TestProxyConfig_Validate_TLSFingerprintNotMutated` — guards the USK-1021 no-rewrite invariant.
- `TestLoadFile_InvalidTLSFingerprint` / `TestLoadConfigs_InvalidTLSFingerprint` — the file → `Validate` boot gate rejects a typo.
- `TestLoadConfigs_TLSFingerprintSpellingsAccepted` — accepted spellings stored verbatim.
- `TestApplyTLSFingerprintFlag_TrimsWhitespace` — CLI flag alignment.
- `TestApplyTLSFingerprint_Vocabulary` — MCP entry point, asserting the resulting transport type (`UTLSTransport` vs `StandardTransport`) rather than just the error.

Verified against current `main` (rebased locally onto 66886618): `make vet`, `make test` (fast tier, `-race`) both green, zero failures. `gofmt` clean.

Refs: USK-1032

https://claude.ai/code/session_01Bnppbaxf5ZpypkrqpXRtxN

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -35,11 +35,11 @@ func (c *Config) Validate() error {
 	if c.DialTimeout < 0 {
 		return fmt.Errorf("dial_timeout must be >= 0, got %s", c.DialTimeout)
 	}
-	if c.LogLevel != "" && !validLogLevels[strings.ToLower(c.LogLevel)] {
-		return fmt.Errorf("invalid log level: %q (must be debug, info, warn, or error)", c.LogLevel)
+	if err := c.validateLogging(); err != nil {
+		return err
 	}
-	if lf := strings.ToLower(c.LogFormat); lf != "" && lf != "text" && lf != "json" {
-		return fmt.Errorf("invalid log format: %q (must be text or json)", c.LogFormat)
+	if err := validateTLSFingerprintField(c.TLSFingerprint); err != nil {
+		return err
 	}
 	if c.RetentionMaxFlows < 0 {
 		return fmt.Errorf("retention_max_flows must be >= 0, got %d", c.RetentionMaxFlows)
@@ -55,6 +55,19 @@ func (c *Config) Validate() error {
 	}
 	if err := c.validateMaxRawCaptureSize(); err != nil {
 		return err
+	}
+	return nil
+}
+
+// validateLogging checks the log level / log format enums. Both accept the
+// empty string (meaning "use the default") and compare case-insensitively.
+// Extracted from Validate to keep that function under the gocyclo threshold.
+func (c *Config) validateLogging() error {
+	if c.LogLevel != "" && !validLogLevels[strings.ToLower(c.LogLevel)] {
+		return fmt.Errorf("invalid log level: %q (must be debug, info, warn, or error)", c.LogLevel)
+	}
+	if lf := strings.ToLower(c.LogFormat); lf != "" && lf != "text" && lf != "json" {
+		return fmt.Errorf("invalid log format: %q (must be text or json)", c.LogFormat)
 	}
 	return nil
 }
@@ -388,6 +401,63 @@ func UTLSProfileFor(fingerprint string) string {
 		return ""
 	}
 	return fingerprint
+}
+
+// tlsFingerprintNames lists the accepted tls_fingerprint identity values in
+// documentation order. Five name a uTLS parrot profile; "none" is the opt-out
+// sentinel that UTLSProfileFor resolves to "" at the dial seam.
+var tlsFingerprintNames = []string{"chrome", "firefox", "safari", "edge", "random", "none"}
+
+// validTLSFingerprints is the lookup set derived from tlsFingerprintNames.
+var validTLSFingerprints = func() map[string]bool {
+	m := make(map[string]bool, len(tlsFingerprintNames))
+	for _, name := range tlsFingerprintNames {
+		m[name] = true
+	}
+	return m
+}()
+
+// ValidTLSFingerprint reports whether fingerprint names one of the six
+// accepted TLS fingerprint identities. The comparison is case-insensitive
+// with surrounding whitespace trimmed — the same normalization every
+// consumption seam already applies (UTLSProfileFor, tlslayer.clientUTLS,
+// transport.ParseBrowserProfile, http2.resolveH2Fingerprint) — so a spelling
+// this predicate accepts is a spelling that works coherently end to end.
+//
+// The empty string reports false. Whether "unset" is acceptable is a
+// per-entry-point policy rather than a property of the vocabulary: the
+// file-config validators treat "" as "unset → DefaultTLSFingerprint" and skip
+// the check, while the `-tls-fingerprint` CLI flag and the MCP proxy_start /
+// configure tools only reach this predicate for an explicit override.
+//
+// This is the single source of truth for the fingerprint vocabulary
+// (USK-1032); it validates only. File-config validation must not write the
+// normalized spelling back into ProxyConfig.TLSFingerprint: the stored
+// identity stays verbatim so ResolveTLSFingerprint / UTLSProfileFor observe
+// the operator's exact bytes (USK-1021).
+func ValidTLSFingerprint(fingerprint string) bool {
+	return validTLSFingerprints[strings.ToLower(strings.TrimSpace(fingerprint))]
+}
+
+// TLSFingerprintNamesList returns the accepted tls_fingerprint values joined
+// with ", " for use in error messages. Stable documentation order so the
+// config-file, CLI-flag, and MCP-tool surfaces all enumerate the same list.
+func TLSFingerprintNamesList() string {
+	return strings.Join(tlsFingerprintNames, ", ")
+}
+
+// validateTLSFingerprintField checks a tls_fingerprint config-file field.
+// An empty value is accepted: it is the documented "unset" spelling that
+// ResolveTLSFingerprint maps to DefaultTLSFingerprint. The value is never
+// rewritten — see ValidTLSFingerprint.
+func validateTLSFingerprintField(fingerprint string) error {
+	if fingerprint == "" {
+		return nil
+	}
+	if !ValidTLSFingerprint(fingerprint) {
+		return fmt.Errorf("tls_fingerprint %q is not valid (expected %s)", fingerprint, TLSFingerprintNamesList())
+	}
+	return nil
 }
 
 // validateProjectName checks that a project name contains only safe characters.
@@ -870,7 +940,15 @@ type SafetyFilterRuleConfig struct {
 //
 // Other ProxyConfig sections (WebSocket / GRPC / SSE limits, SafetyFilter)
 // continue to use their dedicated standalone validators called from main.
+//
+// USK-1032: tls_fingerprint is checked here so a config-file typo fails at
+// load time instead of silently falling back to a Go-native ClientHello on
+// the resend / fuzz dial path while `query config` still reports the typo
+// back to the agent. Validation rejects; it never rewrites the field.
 func (c *ProxyConfig) Validate() error {
+	if err := validateTLSFingerprintField(c.TLSFingerprint); err != nil {
+		return err
+	}
 	for i := range c.Plugins {
 		if err := c.Plugins[i].Validate(); err != nil {
 			return fmt.Errorf("plugins[%d]: %w", i, err)
--- a/internal/config/config_tls_test.go
+++ b/internal/config/config_tls_test.go
@@ -3,7 +3,9 @@ package config
 import (
 	"os"
 	"path/filepath"
+	"strings"
 	"testing"
+	"time"
 )
 
 func TestLoadFile_TLSFingerprint(t *testing.T) {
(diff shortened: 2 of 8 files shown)


usk6666 (commented):

This PR closes the USK-1032 / PR #93 S-3 (CWE-636) gap correctly. I traced every writer and reader of `ProxyConfig.TLSFingerprint` / `Config.TLSFingerprint` and **the central reachability claim behind design decision #2 holds**: the `InitTLSTransport` fail-soft fallback is now genuinely unreachable in production. The normalization design (decision #1) is also sound — I verified empirically that the validator predicate and every consumption seam apply byte-identical normalization, so there is no Unicode/case class that the predicate accepts but a consumer resolves differently. One LOW maintenance finding.

### Verification of the reachability claim (design decision #2)

`InitTLSTransport` has exactly **one** production call site: `mcpserver/server.go:361`, inside `assembleAndRunMCPServer`, which is called only from `Run` (`server.go:295`). Both of its inputs are validated earlier in `Run`:

| Input | Writers (non-test) | Validated at |
| -- | -- | -- |
| `cfg.TLSFingerprint` | **none in production** — no flag, no env var, no loader writes it (`LoadFile` returns `*ProxyConfig`). Tier 2 is dead today. | `cfg.Validate()` `server.go:192` (newly covers it, so a future wiring is pre-guarded) |
| `proxyCfg.TLSFingerprint` | `config.LoadFile` JSON unmarshal; `ApplyTLSFingerprintFlag` (`server.go:485`) | `proxyCfg.Validate()` in `LoadConfigs` `server.go:509` (new); `ApplyTLSFingerprintFlag` `server.go:482` |

`YP_TLS_FINGERPRINT` is routed through `ApplyEnvFallback` → `fs.Set("tls-fingerprint", …)` → the same validated flag, so the env axis is covered. `RunOptions` carries no injectable `Config`, so there is no programmatic-construction path into `Run`.

The **USK-809 runtime override** does *not* reach `InitTLSTransport` at all — it is a separate axis (`proxybuild.Manager.SetTLSFingerprint` → `BuildConfig.SetTLSFingerprint`). Its only production writer is `mcp.applyTLSFingerprint` (`proxy_start_tool.go:1166`/`:1178`), which now validates via `config.ValidTLSFingerprint` *before* normalizing and installing. No path bypasses a validator.

Two fail-soft fallbacks remain, and both are now provably dead:
- `InitTLSTransport` (`init.go:768`) — unreachable per the table above, logs at `Warn` (visible at default log level).
- `buildTLSTransport` (`proxy_start_tool.go:1197`) — unreachable because `ValidTLSFingerprint` minus `"none"` is exactly `ParseBrowserProfile`’s domain.

Keeping them as defence-in-depth instead of changing the signature to return an error is a reasonable call.

### Verification of the normalization design (design decision #1)

All four seams normalize identically with `strings.ToLower(strings.TrimSpace(x))` — `transport.ParseBrowserProfile`, `http2.resolveH2Fingerprint`, `tlslayer.clientUTLS`, and now `config.ValidTLSFingerprint`. The only seam using a different primitive is `config.UTLSProfileFor`, which uses `strings.EqualFold` for the `"none"` sentinel.

I ran an exhaustive Unicode scan (U+0001–U+2FFFF) for the two divergence classes the review brief asked about:

1. **`ToLower` lookalikes** — the only non-ASCII rune that folds into the vocabulary alphabet is `U+0130 İ → i`. But since *every* consumer applies the same `ToLower`, `tls_fingerprint: "fİrefox"` resolves to `ProfileFirefox` on the uTLS dial, `H2FingerprintFirefox` on the H2 send-shape, and `ProfileFirefox` on the resend transport. Coherent end to end — no split identity.
2. **`ToLower` vs `EqualFold` for `"none"`** — zero divergences across all four character slots over the entire scanned range. The sentinel cannot be reached by a spelling the validator rejects, nor missed by one it accepts.

The USK-1021 no-rewrite invariant is preserved on the config-file path (values stored verbatim; `TestProxyConfig_Validate_TLSFingerprintNotMutated` guards it), and padded/mixed-case verbatim values stay coherent because `UTLSProfileFor` returns the untrimmed original and every downstream seam trims.

### Other checks

- **`validTLSFingerprints` mutability** (`config.go:412`) — built by a package-level func literal, unexported, and the only write is inside the initializer. Read-only after init; concurrent map reads from `applyTLSFingerprint` are safe. No race.
- **Error messages** — both new messages quote the operator’s raw input with `%q`, which escapes control characters (no terminal-escape injection into the MCP agent / stderr). This is operator config, not wire data; no path or internal state is leaked.
- **Availability** — the new boot-time rejection is strictly better than the previous behaviour: a typo already hard-failed every live MITM dial (`tlslayer: unsupported uTLS profile`), so the proxy was non-functional anyway. The config file is operator-owned, not attacker-reachable, so this is not a DoS vector.
- **Dependencies** — none added.
- Verified against `origin/main` (66886618): no conflicting changes landed in `internal/config`, `internal/mcpserver`, or the fingerprint path of `internal/mcp` during the 7-week gap, so none of the above is stale-base analysis.

### Security Findings

| S-1 | LOW | internal/config/config.go | 406-418 | TLS/Cert | CWE-1188 (Insecure Default Initialization of Resource) | The new single source of truth covers *acceptance* only, not *resolution*. `config.tlsFingerprintNames`, `transport.ParseBrowserProfile`’s switch, and `tlslayer.utlsProfileIDs` are three independently-maintained tables with no test binding them together. |

**S-1 detail.** Today the three tables agree exactly (`chrome`/`firefox`/`safari`/`edge`/`random`, plus the `none` sentinel handled by `UTLSProfileFor`), which is what makes both fail-soft fallbacks dead code. But nothing pins that. Adding a name to `tlsFingerprintNames` without adding it to `utlsProfileIDs` would make the config file accept a value that hard-errors on every live MITM dial, while the resend axis silently degrades to a Go-native ClientHello through the very `InitTLSTransport` fallback this PR just declared unreachable — i.e. it re-opens CWE-636 by drift. Given the project’s own “mirror the parser, don’t approximate it” lesson (USK-1035), the vocabulary deserves the same treatment.

**Remediation (non-blocking).** Neither `internal/connector/transport` nor `internal/layer/tlslayer` imports `internal/config`, so a test in one of those packages can import `config` without a cycle. Assert that every name in `config.TLSFingerprintNamesList()` except `"none"` is accepted by `transport.ParseBrowserProfile` and is a key of `tlslayer.utlsProfileIDs`, and that the sets are the same size in both directions (so an addition on either side fails the test). A bidirectional guard is what makes this catch real drift rather than just one direction of it.

### STATS

```
CRITICAL: 0
HIGH:     0
MEDIUM:   0
LOW:      1
```

---

usk6666 (commented):
## Code Review: APPROVED ✅

The fix closes the USK-1032 gap correctly and the central claim in the commit message holds under verification: all four named consumption seams (`config.UTLSProfileFor`, `tlslayer.clientUTLS`, `transport.ParseBrowserProfile`, `http2.resolveH2Fingerprint`) apply exactly `TrimSpace` + `ToLower`, so every spelling the new predicate accepts resolves coherently end to end on both the identity and the dial axis. Only LOW/NIT observations, all about polish rather than behaviour.

### What was verified

- **Seam normalization parity** — confirmed `strings.ToLower(strings.TrimSpace(...))` (or `EqualFold` + `TrimSpace`) at all four seams. Traced `"  Firefox  "` and `"  NONE  "` end to end: H2 send-shape ✅, uTLS parrot ✅, `InitTLSTransport` ✅, `StandardTransport` opt-out ✅.
- **Entry-point coverage** — config file (`LoadConfigs` → `ProxyConfig.Validate`), `-tls-fingerprint` / `YP_TLS_FINGERPRINT` (`ApplyTLSFingerprintFlag`), `proxy_start` / `configure` (`applyTLSFingerprint`), and the config → `proxy_start` defaults merge (`applyProxyDefaultTLSStrings`) all now gate on the same set. `config.LoadFile` has exactly one production caller and it validates immediately after.
- **`InitTLSTransport` fail-soft justification** — single caller (`server.go:361`), reached only after both `cfg.Validate()` and `LoadConfigs` + `ApplyTLSFingerprintFlag`. The godoc's "now unreachable" claim is accurate.
- **USK-1021 pool-key invariant** (`EffectiveUTLSProfile` godoc: *"`UTLSProfileFor` must never merge two identities that `resolveH2Fingerprint` distinguishes"*) — not violated. The merged set is still only the `"none"` spellings ∪ `""`, all of which share both the standard-`crypto/tls` handshake and the default H2 shape.
- **gocyclo** — old `Config.Validate` was exactly 15 (`1 + 9 ifs + 3 short-circuit operators`); inlining the new check would have pushed it to 16, over `min-complexity: 15`. The `validateLogging` extraction is genuinely load-bearing, not drive-by. New value is 12.
- **Precedent shape** — `tls_fingerprint %q is not valid (expected %s)` matches `intercept_queue.timeout_behavior %q is not valid (expected %q or %q)` (`internal/config/intercept_queue.go:141`).
- **Reachability of the boot-gate test** — `TestLoadConfigs_InvalidTLSFingerprint` writes a real JSON file and calls the real `LoadConfigs`, which `mcpserver.Run` invokes unconditionally at `server.go:195`. Not a simulated-producer test.
- Locally: `gofmt -l` clean, `go vet` clean (including `-tags e2e` over `internal/mcptest`, `cmd`, `internal/config`, `internal/mcpserver`), and `internal/config` + `internal/mcpserver` + `internal/mcp` tests green. No pre-existing test asserted the old non-trimming behaviour. No conflict with the 22 commits `main` has moved ahead (`internal/config` and `internal/mcpserver` untouched there; `main`'s `proxy_start_tool.go` edits are jsonschema strings far from this diff).

### Findings

| F-1 | LOW | internal/config/config.go | 420-460 | Correctness | Entry-point storage asymmetry: the config-file path stores the fingerprint verbatim while the CLI flag and MCP tool store the normalized spelling |
| F-2 | NIT | internal/config/config.go | 442-447 | GoConventions | `TLSFingerprintNamesList()` returns a joined `string`, but the name reads as a slice accessor |
| F-3 | NIT | internal/mcp/proxy_start_tool.go | 1184-1190 | CodeHealth | `buildTLSTransport` exact-matches `profile == "none"`; its now-load-bearing "caller must pass a normalized name" precondition is undocumented |
| F-4 | NIT | internal/mcp/proxy_start_tool_test.go | 939-997 | TestQuality | `TestApplyTLSFingerprint_Vocabulary` asserts the transport but not the normalized spelling installed via `SetTLSFingerprint` |
| F-5 | NIT | internal/config/config_tls_test.go, internal/mcpserver/init_test.go | 212-229, 826-847 | TestQuality | Raw fingerprint strings used as subtest names produce `#00` / `__NONE__` |

### Suggestions

**F-1 (LOW)** — `ProxyConfig.Validate` deliberately stores verbatim (USK-1021 no-rewrite), while `ApplyTLSFingerprintFlag` (`server.go:485`) and `applyTLSFingerprint` (`proxy_start_tool.go:1162`) store `strings.ToLower(strings.TrimSpace(...))`. Two surfaces observe the difference:

1. `query config` echoes `"  Firefox  "` when it came from the file, `"firefox"` when it came from `-tls-fingerprint` or `configure`.
2. `EffectiveUTLSProfile()` keys `poolKeyForH2` and `ALPNCacheKeyFromConfig`. A boot-time `"  Firefox  "` and a runtime `configure(tls_fingerprint="firefox")` now occupy **separate** pool/ALPN keyspaces despite being the same dial — exactly what the `ALPNCacheKeyFromConfig` godoc says the resolved form exists to avoid (*"so a boot-time `none` and a runtime `none` share one keyspace"*).

This is fail-safe in the only direction that matters: splitting causes an extra dial, never a wrong reuse, and the "`resolveH2Fingerprint`-distinguishing identities must not merge" invariant is intact. It is also not a regression — padded spellings were already accepted before this PR (by the absence of validation). But the PR *blesses* them as officially valid, so the keyspace-splitting spelling becomes a supported input rather than an accident.

Two acceptable resolutions, either is fine:
- Add a `config.NormalizeTLSFingerprint` applied at the load boundary (`LoadConfigs`, after `Validate`) so all three entry points store the same bytes. This does **not** weaken the USK-1021 fail-hard property — `UTLSProfileFor` stays untouched, and a typo is already rejected before normalization would run.
- Or keep verbatim storage and extend the "Load-bearing invariant" block on `EffectiveUTLSProfile` (`internal/connector/stack_builder.go:596-608`) with one sentence: the *splitting* direction is now reachable via config-file whitespace/case padding, is benign, and must stay benign.

**F-2 (NIT)** — consider `TLSFingerprintNames() string` (dropping `List`), or exposing `TLSFingerprintNames() []string` and letting each call site choose its separator — the precedent `rotation.policy` message uses `per_request | per_connection | ...` while this one uses commas.

**F-3 (NIT)** — add to `buildTLSTransport`'s godoc: *"profile must already be trimmed and lowercased; `applyTLSFingerprint` is the only caller and normalizes first."* The `"none"` comparison is exact, so a direct call with `"  NONE  "` would silently build a `UTLSTransport` via the `ParseBrowserProfile` fallback path instead of the intended `StandardTransport`.

**F-4 (NIT)** — add `if got := s.currentTLSFingerprint(); got != tt.wantStored { ... }` (or assert on the `tlsFingerprintSetter`) for the `"  FireFox  "` case. That value is what `query config` returns to the agent, and it is the half of the normalization contract the test currently leaves unpinned.

**F-5 (NIT)** — switch both loops to `[]struct{ name, fingerprint string }` so `""` reads as `"unset"` and `"  NONE  "` as `"padded none"`.

---

PR #118, opened 2026-09-15
fix(layer/grpc): derive send-side :authority / :scheme from the L7 overlay

usk6666:
## Summary
- `internal/layer/grpc/channel.go`: `authorityFromContext` → `authorityForStart(m)` reads `GRPCStartMessage.Authority` **only** (no `Context.TargetHost` fallback); `schemeFromContext` → `schemeForStart(env, m, fallback)` prefers `m.Scheme`, keeping the `Context.TLS` derivation as fallback.
- `internal/mcp/resend_grpc_helpers.go`: new `resendGRPCPlan.scheme` field populated alongside `useTLS`; `buildResendGRPCStartEnvelope` now sets `Authority` / `Scheme` on the message. One line covers `fuzz_grpc` too via the shared builder.
- `internal/pluginv2/convert.go`: preserve-only fix — the `GRPCStartMessage` builder carries `Authority` / `Scheme` / `Path` across the rebuild. **No new dict keys** (the hook surface is unchanged).
- Stale doc comments corrected in `internal/mcp/intercept_typed.go`, `internal/envelope/grpc.go`, and both gRPC Layer helpers.
- ~160 LOC of tests across three tiers (fast / smoke / nightly).

## Root cause
The Send path derived `:authority` from `Envelope.Context.TargetHost` and returned a hardcoded `"https"` for `:scheme`, ignoring the `GRPCStartMessage.Authority` / `.Scheme` L7 overlay that USK-920 added and that `absorbHeaders` already populates from the wire on Receive.

Two defects shared that root cause:

1. **`resend_grpc` / `fuzz_grpc` / `grpc_schema` discover sent no `:authority` at all.** Their synthetic envelopes set `Context{ConnID}` only, so `TargetHost` was empty and `appendRequestPseudoHeaders` (which gates on `evt.Authority != ""`) omitted the field. grpc-go 1.83.2, pulled in by #115, added a `len(mdata[":authority"]) == 0` → `writeEarlyAbort(codes.Internal)` check, which turned 21 nightly e2e tests in `internal/mcp` red. The grpc-go change is the exposure, not the cause — `:authority` is mandatory for clients under RFC 9113 §8.3.1.
2. **Live MITM rewrote the observed `:authority`** to the CONNECT target, and any observed h2c `:scheme: http` to `https`. Both violate MITM Principle #1 and produce different upstream behaviour than the client would get direct.

The sibling `internal/layer/httpaggregator/` has always used `Authority: msg.Authority` and references `TargetHost` nowhere; gRPC was the outlier.

## Scope decisions
- **No `Context.TargetHost` fallback.** RFC 9113 §8.3.1 requires an intermediary to build `:authority` from the *original request's* control data and forbids generating one when the original carried none. A fallback would make the proxy repair a client that omitted `:authority` — succeeding through the proxy while failing direct. A client that sends only `host:` still works: `buildStartMessage` keeps it in `Metadata` and gRPC's A41 rename applies upstream.
- **`:scheme` fixed alongside `:authority`** — same defect, same function block. The asymmetry is deliberate: authority gets no fallback (RFC forbids synthesis), scheme keeps one (RFC requires exactly one valid value, so a derived default is correct).
- **`internal/pluginv2/convert.go` preserve-fix included** — prerequisite for removing the fallback. The builder runs on *any* plugin mutation and previously zeroed the USK-920 triple, which already silently erased `Flow.URL` and would have become a live wire regression (grpc-go rejecting the RPC) once `sendStart` depended on those fields.
- **No change to `internal/job/grpc_source.go`** — `GRPCResendSpec.Start` is already a caller-owned pass-through, and it has no consumer outside `internal/job`.
- **`:path` deferred** to a follow-up. Two verified blockers: `applyGRPCStartModify` sets `Service`/`Method` but never `Path`, so preferring `m.Path` would break the working intercept override; and `GRPCStartMessage.Path` is not byte-verbatim (the H2 assembler splits the query into `RawQuery`, which this type has no field for).
- **`grpc_schema` discover real-wire e2e deferred** — `defaultReflectionDialFunc` has never been executed by any test (`SetReflectionDialForTest` replaces the whole `reflectionDial` var), and building one needs a hand-registered `ServerReflectionInfo` bidi ServiceDesc. The regression is fully pinned without it: all three producers go through the same single `sendStart` emission site.

## Test plan

**Fast tier (merge gate, untagged)**
- [x] `TestChannel_SendStartPseudoHeadersComeFromMessage` (`internal/layer/grpc/channel_test.go`) — 5 table cases driving `grpcChannel.Send` against the `stubInner` double, asserting the captured `*http2.H2HeadersEvent`. Every case sets `Context.TargetHost` to a value that *differs* from the message authority, so a regression cannot pass by coincidence. This is the only way to assert `:scheme` — it is reserved and **not** whitelisted by grpc-go, so it never reaches `metadata.FromIncomingContext`.
- [x] `TestAuthorityForStart_NilMessage` / `TestSchemeForStart_NilArgs` — nil guards preserved from the old helpers.
- [x] `TestBuildResendGRPCStartEnvelope_CarriesAuthorityAndScheme` (`internal/mcp/resend_grpc_helpers_test.go`) — 3 cases incl. `target_addr` override (dialAddr ≠ authority); also asserts `env.Context.TargetHost` stays empty so the Layer has no fallback source.
- [x] `TestGRPCStart_PseudoHeaderOverlaySurvivesMutation` (`internal/pluginv2/convert_test.go`) — 3 mutation kinds (metadata append / service / encoding), each asserting the triple survives the rebuild.
- [x] `TestGRPCStart_PseudoHeadersAreNotPluginMutableKeys` — pins that the hook surface was *not* widened.

**Smoke tier (merge gate, `//go:build e2e`)**
- [x] `TestGRPC_ClientAuthorityReachesUpstreamVerbatim` (`internal/layer/grpc/grpc_integration_test.go`) — real gRPC RPC through the CONNECT-tunneled MITM. The divergence is already built into the harness: `dialGRPCViaProxy` sets `tlsCfg.ServerName = "localhost"` (which grpc-go uses for `cc.authority`) while the upstream listens on `127.0.0.1:0`, so the client sends `:authority: localhost` and the CONNECT target is the literal address. Placing this in the **smoke** tier is what prevents structural recurrence of a nightly-only miss.

**Nightly tier (`e2e && !e2e_smoke`)**
- [x] Direct `:authority` assertions added to `TestResendGRPC_UnaryRoundTrip` and `TestFuzzGRPC_PayloadPositionGeneratesVariants` (both harnesses already capture incoming metadata; no harness changes needed).

### Before/after evidence — the new tests do detect the bug

Each new test was run against the **unfixed** production code first.

| Test | Unfixed | Fixed |
|---|---|---|
| `TestChannel_SendStartPseudoHeadersComeFromMessage` | **FAIL 5/5 subtests** — e.g. `Authority = "127.0.0.1:34483", want "vhost.example:8443"`; `Scheme = "https", want "http"` | PASS |
| `TestGRPC_ClientAuthorityReachesUpstreamVerbatim` | **FAIL** — `upstream observed :authority = "127.0.0.1:45345", want "localhost"` + `the CONNECT target was substituted for the client's value` | PASS |
| `TestBuildResendGRPCStartEnvelope_CarriesAuthorityAndScheme` | **FAIL 3/3 subtests** — `Authority = "", want "api.example.com:443"` | PASS |
| `TestGRPCStart_PseudoHeaderOverlaySurvivesMutation` | **FAIL 3/3 subtests** — Authority/Scheme/Path all `""` after rebuild | PASS |

### Nightly red cleared

```
$ go test -race -tags e2e -run 'TestResendGRPC|TestFuzzGRPC' ./internal/mcp/
```

- Baseline on `main` (before the fix, same worktree): **21 FAIL** — exactly the set listed in USK-1051.
- After the fix: **46 PASS / 0 FAIL**, `ok ... 39.5s`.

### Full verification

- [x] `make lint` — gofmt clean, golangci-lint v2.13.2, **0 issues**
- [x] `make build`
- [x] `make test` (fast tier) — **4208 PASS / 0 FAIL**
- [x] `make test-e2e-smoke` (merge gate) — **0 FAIL**
- [x] `go test -race -tags e2e ./internal/mcp/` (the previously-red package, full nightly tier) — `ok 471s`
- [x] `go test -race -tags e2e` on every other gRPC-touching package — `connector`, `encoding/protoschema`, `job`, `layer/grpc`, `layer/grpcweb`, `pipeline`, `pluginv2`, `proxybuild`, `rules/grpc`, `flow` — all `ok`
- [x] No new external dependencies

Resolves USK-1051
Linear: https://linear.app/usk6666/issue/USK-1051


usk6666 (commented):

No CRITICAL/HIGH/MEDIUM issues. The two values this PR newly threads to the wire (`:authority`, `:scheme`) reach only the HPACK header block — neither influences the dial target, the h2 pool key, or capture-scope evaluation on the live MITM path — and the `internal/pluginv2/convert.go` preserve-only fix does not widen the plugin sandbox (the preserved values come from the captured pre-mutation Go message, and `MessageDict.SetKey` still rejects `authority`/`scheme`/`path` as unknown keys). Three LOW observations below, two of which are pre-existing.

### Authority / Scheme data flow (as traced)

**(a) Live MITM path.** Client HEADERS → `assembler.go` `applyPseudo` (`pf.authority` / `pf.scheme`; duplicates flagged `H2DuplicatePseudoHeader`) → `H2HeadersEvent` → `buildStartMessage(evt, dir)` writes `GRPCStartMessage.Authority/.Scheme` on the request side only, and strips every `:`-prefixed name out of `Metadata` → Pipeline → `sendStart` → `authorityForStart` / `schemeForStart` → `appendRequestPseudoHeaders` → `hpack.HeaderField{Name: ":authority", ...}`.
No Pipeline Step can rewrite the pair: `rules/grpc.applyStartAction` only edits `Metadata`; `applyGRPCStartModify` mutates Service/Method/Encoding/Metadata on an `env.Clone()` whose `CloneMessage` copies all three fields; pluginv2 never registers them as dict keys.
**Dial target: not influenced.** The upstream connection predates any envelope; `poolKeyForH2` keys the h2 pool on the CONNECT target + TLS/fingerprint/proxy config, never on `:authority`. `flow/record_scope.go` `extractScopeFields` takes the gRPC hostname from `Context.TargetHost` / `TLS.SNI` and explicitly *not* from `m.Authority`, so a forged authority cannot move a stream in or out of capture scope.

**(b) resend / fuzz path.** `flow_id` → `extractResendGRPCStartFields` → `Flow.URL.Host` / `.Scheme` (written by `projectGRPCStart` from the wire-observed pseudo-headers); otherwise `input.TargetAddr` / `input.Scheme`. → `plan.authority` / `plan.scheme` → `buildResendGRPCStartEnvelope` → same Layer path. `fuzz_grpc` inherits both through `cloneFuzzGRPCPlan`'s by-value copy.
**Dial target: yes, `authority` determines the dial here** (`dialAuthority = authority` unless `target_addr` overrides), but that is pre-existing and untouched by this diff — the only change on that line is `plan.scheme = scheme`. It is gated by `checkResendGRPCScope`, which checks the canonical URL and separately re-checks the override address whenever `dialAddr != canonicalURL.Host`.

### Cleared explicitly

- **Header injection (CWE-113/93)** — `internal/layer/http2/hpack.Encoder` emits length-prefixed literals, so CR/LF/NUL in `:authority` cannot terminate or add a field, and cannot desync the framer (CONTINUATION splitting is length-driven). Forwarding an anomalous wire authority is intended MITM behaviour; an MCP-supplied one cannot smuggle extra header fields.
- **pluginv2 sandbox** — `out := &GRPCStartMessage{Authority: m.Authority, ...}` reads the closure-captured pre-mutation message (Go strings, no aliasing). `SetKey` returns `msg[%q]: unknown field` for keys not registered via `setKV`, and the new `TestGRPCStart_PseudoHeadersAreNotPluginMutableKeys` pins that.
- **MCP information disclosure** — no new response field; `intercept_typed.go` is comment-only. Output SafetyFilter (`filterOutput*`) coverage is unchanged.
- **Go basics** — pure functions over immutable strings; `-race` unit tests plus the gRPC / resend_grpc / fuzz_grpc e2e tests pass. `gofmt` clean, no `go.mod` / `go.sum` change.

### LOW findings

| S-1 | LOW | internal/mcp/resend_grpc_helpers.go | 70-85, 151-171, 293 | InputValidation | CWE-20 | `scheme` allowlist (`http`/`https`) runs only in `validateResendGRPCFromScratch`, i.e. only when `flow_id == ""`. With `flow_id` set, `input.Scheme` is CRLF-checked and lowercased but otherwise arbitrary, and after this PR it reaches the wire `:scheme` (previously hardcoded `"https"`). Shared by `fuzz_grpc`. |
| S-2 | LOW | internal/mcp/resend_grpc_helpers.go | 259-263, 344-358 | Network | CWE-918 / CWE-319 | Pre-existing (USK-920): the recorded `Flow.URL.Host` / `.Scheme` — i.e. the *client-forged* `:authority` / `:scheme`, not the CONNECT target — drive `plan.dialAddr` and `plan.useTLS` on `resend_grpc {flow_id}`. Note `recScheme` wins over the true `stream.Scheme`. |
| S-3 | LOW | internal/layer/grpc/channel.go | 1434-1469 | GoSecurity | CWE-476 | The new `m != nil` guards in `authorityForStart` / `schemeForStart` are unreachable: `sendStart` already dereferences `m` unconditionally at `buildStartHeaderKVs` (`len(m.Metadata)`) and `buildGRPCPath(m.Service, m.Method)` two lines earlier, so a typed-nil `*GRPCStartMessage` panics before the guards run. |

### Remediation

- **S-1** — move the `scheme != "" && scheme != "http" && scheme != "https"` check from `validateResendGRPCFromScratch` up into `validateResendGRPCStringFields` so both branches share it. Not exploitable as it stands (HPACK is length-prefixed, so an odd `:scheme` is a malformed *message* the upstream rejects, not an injection, and sending one is arguably a legitimate pentest capability), but the divergence is real: `fuzz_grpc_helpers.go:1078-1082` re-derives `canonicalURL`'s scheme from `useTLS`, so the safety/scope URL would read `http` while the wire reads the arbitrary value. Worth a maintainer decision rather than a silent gap.
- **S-2** — out of scope for this PR (the diff adds only `plan.scheme = scheme` on that line). Concrete scenario, for a follow-up: a client behind the proxy sends `:authority: attacker.tld:80` + `:scheme: http` over a TLS gRPC connection; a later `resend_grpc {flow_id}` with no `target_addr` dials `attacker.tld:80` in plaintext h2c and replays that flow's recorded metadata in the clear. Mitigated when TargetScope rules are configured (open mode is the default), and the replayed credentials are the attacker's own, which is why this is LOW. Consider preferring `Stream.Scheme` (the true handshake transport) over the wire-forged `Flow.URL.Scheme` for the `useTLS` decision, or warning when they disagree.
- **S-3** — no action required for safety. If the guards are meant to be load-bearing, the nil check belongs at the top of `sendStart` (or in `Send`'s type switch) rather than in the two leaf helpers; otherwise the two new unit tests assert a property the caller cannot actually rely on.

STATS — CRITICAL: 0 / HIGH: 0 / MEDIUM: 0 / LOW: 3

---

usk6666 (commented):
## Code Review: APPROVED ✅

The fix is correct and complete: every non-test Send-side `GRPCStartMessage` producer that can reach `grpcChannel.sendStart` now sets `Authority`, so removing the `Context.TargetHost` fallback introduces no silent hard failure against grpc-go >= 1.83.2. All four new/extended test groups were mutation-verified to actually fail when the fix is reverted, and the new doc comments check out against the code (including grpc-go v1.83.2's `"no host or :authority header present"` / A41 rename, which I read in the module cache). One MEDIUM test-coverage gap on the newly-introduced `resendGRPCPlan.scheme` field, plus doc/code-health nits.

### Producer table (independently re-derived)

| # | Producer | Reaches `grpc.sendStart`? | Sets `Authority`? |
|---|----------|---------------------------|-------------------|
| 1 | `internal/layer/grpc/channel.go:1036` `buildStartMessage` (sole caller `absorbHeaders:444`) | YES — live MITM | YES for `dir==Send` (`msg.Authority = evt.Authority`); intentionally empty on `Receive` (response uses `:status`) |
| 2 | `internal/mcp/resend_grpc_helpers.go:927` `buildResendGRPCStartEnvelope` | YES (`Wrap` at `resend_grpc_helpers.go:799`) | YES (this PR). `finalizeResendGRPCAuthorityScheme` guarantees non-empty-or-error. Also covers `fuzz_grpc` via `cloneFuzzGRPCPlan` (verified `out := *base` by-value copy) |
| 3 | `internal/encoding/protoschema/reflection.go:630` `buildReflectionStartEnvelope` | YES (`Wrap` at `reflection.go:327`) | YES (pre-existing); `normalizeDiscoverOptions` rejects empty `TargetAddr` |
| 4 | `internal/mcp/grpc_schema_discover.go:266` | NO — budget-accounting envelope, only fed to `pipeline.NewBudgetStep` | YES anyway |
| 5 | `internal/pluginv2/convert.go:394` dict read-back builder | YES — replaces `env.Message` for any `grpc`/`start` hook mutation | YES (this PR) — was the silent eraser |
| 6 | `internal/envelope/grpc.go:87` `CloneMessage` | YES — `applyGRPCStartModify` + `rules/grpc` engines | YES (pre-existing) |
| 7 | `internal/layer/grpcweb/channel.go:331` | NO — `grpcweb.Send` → `buildHTTPMessage` → httpaggregator | NO (see F-6, pre-existing) |
| 8 | `internal/job/grpc_source.go` `GRPCResendSpec.Start` | no non-test consumer (grep-verified) — settled decision #4 holds | N/A |

### Verification performed

- **Mutation testing** (each fix reverted in a scratch copy, then restored — tree is clean):
  - revert `authorityForStart`/`schemeForStart` → 4/5 table cases + `TestSchemeForStart_NilArgs` FAIL.
  - revert to `Authority: env.Context.TargetHost` → `TestGRPC_ClientAuthorityReachesUpstreamVerbatim` FAILs with `:authority = "127.0.0.1:42497", want "localhost"`.
  - revert the pluginv2 builder → all 3 `…OverlaySurvivesMutation` subcases FAIL.
  - add `d.setKV("Authority", …)` → `TestGRPCStart_PseudoHeadersAreNotPluginMutableKeys` FAILs (confirmed `setKV` appends to `d.keys` regardless of read-only, so the guard is real).
  - drop `Authority: plan.authority` → nightly `TestResendGRPC_UnaryRoundTrip` + `TestFuzzGRPC_PayloadPositionGeneratesVariants` FAIL, handler never runs (`req=""`), matching the doc claim.
- `stubInner.Send` genuinely records what production `sendStart` hands to `c.inner.Send` — the table test is not self-asserting.
- Build tag confirmed: `internal/layer/grpc/grpc_integration_test.go` is plain `//go:build e2e` → merge-gate smoke tier.
- Wire fidelity: nothing in the diff normalizes, dedups, or reorders headers/metadata. `buildStartHeaderKVs` untouched; `host` is not in `connectionSpecificHeaders`, so the A41 host-only fallback documented in `authorityForStart` actually works.
- Package boundary: no `net/http` added to `internal/envelope/`, `internal/layer/`, `internal/pluginv2/`.
- `gofmt -l .` clean; `go vet` clean (incl. `-tags e2e`); smoke tier green for `internal/layer/grpc`, `internal/layer/grpcweb`, `internal/encoding/protoschema`; **full** e2e tier green for `internal/mcp` (137s).

### Findings

| F-1 | MEDIUM | internal/mcp/resend_grpc_helpers.go / resend_grpc_helpers_test.go | 265 / 209 | TestQuality | New `plan.scheme` assignment has zero test coverage — mutation-verified |
| F-2 | LOW | internal/layer/grpc/channel.go | 1465-1470, 1587-1614 | CodeHealth | Nil guards (and their two tests) are unreachable — `buildStartHeaderKVs` derefs `m` first |
| F-3 | LOW | internal/layer/grpc/channel.go | 1413-1417, 1472-1479 | GoConventions | `methodOr` / `statusFor` doc comments are factually false, adjacent to the block this PR rewrote |
| F-4 | NIT | internal/envelope/grpc.go | 53-56 | CodeHealth | `Metadata` doc still says pseudo-headers are "observable via Envelope.Context" |
| F-5 | NIT | internal/mcp/fuzz_grpc_helpers.go | 1078-1081 | CodeHealth | Scheme re-derived from `useTLS` next to the new authoritative `variantPlan.scheme` |
| F-6 | NIT | internal/layer/grpcweb/channel.go | 1060-1070 | Architecture | Pre-existing sibling gap: gRPC-Web forward drops `:authority` on the H2 upstream path |

### Suggestions

**F-1 (MEDIUM)** — `resendGRPCPlan.scheme` is assigned in exactly one place (`resend_grpc_helpers.go:265`) and read in exactly one place (`:931`). I removed the assignment and ran the **entire fast tier for `internal/mcp`** (`ok`, 32s) and the nightly gRPC e2e — both stay green, because every existing resend/fuzz test passes `"scheme": "https"`, which is identical to `schemeForStart`'s literal fallback. `:scheme` is also reserved-and-not-whitelisted in grpc-go, so no e2e can observe it. Meanwhile `buildResendGRPCPlan` has **no test anywhere** (grep-verified).

`TestBuildResendGRPCStartEnvelope_CarriesAuthorityAndScheme` builds `resendGRPCPlan` literals, so it proves the copy but nothing about the producer — the simulated-producer pattern the project rules call out. A regression would silently emit `:scheme: https` over an h2c connection: exactly the fidelity defect USK-1051 set out to fix.

Cheap closure — `buildResendGRPCPlan` needs no store when `FlowID == ""`:

```go
func TestBuildResendGRPCPlan_ResolvesSchemeAndAuthority(t *testing.T) {
    for _, tc := range []struct{ name, inScheme, want string }{
        {"h2c", "http", "http"},
        {"tls", "https", "https"},
        {"default", "", "https"},
    } {
        plan, err := (&Server{}).buildResendGRPCPlan(context.Background(), &resendGRPCInput{
            TargetAddr: "127.0.0.1:50051", Scheme: tc.inScheme,
            Service: "S", Method: "M",
            Messages: []resendGRPCMessageInput{{Payload: "x"}},
        })
        // assert plan.scheme == tc.want && plan.authority == "127.0.0.1:50051"
    }
}
```

**F-2 (LOW)** — `sendStart` calls `buildStartHeaderKVs(env, m)` first, which dereferences `m.Metadata` / `m.ContentType` unconditionally; `buildGRPCPath(m.Service, m.Method)` follows. A typed-nil `*GRPCStartMessage` in the interface therefore panics before `authorityForStart(m)` / `schemeForStart(env, m, …)` are ever reached, so the guards cannot fire and `TestAuthorityForStart_NilMessage` / `TestSchemeForStart_NilArgs` assert a state production cannot produce. Not a regression (the old guards were equally decorative), and the second half of `TestSchemeForStart_NilArgs` is load-bearing. Either drop the nil branches, or move the check up to `sendStart`'s entry so it is genuinely reachable and the tests become meaningful.

**F-3 (LOW)** — This PR rewrote the doc block these two sit in, so worth correcting in the same pass:

```go
// methodOr returns evt.Method when set, else fallback. POST is the
// canonical gRPC verb.
func methodOr(_ *envelope.GRPCStartMessage, env *envelope.Envelope, fallback string) string {
	_ = env
	return fallback
}
```

There is no `evt` in scope and neither parameter is read — it always returns `fallback`. `statusFor`'s "returns evt.Status" is the same shape (it returns a hardcoded `200`). Suggest rewording to state the real behaviour and, for `methodOr`, noting *why* `:method` stays fabricated: unlike `:authority`/`:scheme`, `GRPCStartMessage.Method` is the gRPC method name, not the HTTP verb, so there is no overlay field to recover it from — the same structural blocker as the `:path` deferral, so a pointer to USK-1053 would land it in one place.

**F-4 (NIT)** — "HTTP/2 pseudo-headers are NOT included here — they belong to the transport layer and are **observable via Envelope.Context if needed**". Since USK-920 they are observable via the `Authority`/`Scheme`/`Path` fields on this very struct, and USK-1051 makes "not via Context" an explicit design rule. Suggest: "…they are projected onto the Authority / Scheme / Path fields above (request side only); `Envelope.Context` is not a source for them — see `authorityForStart`."

**F-5 (NIT)** — `variantPlan.scheme` is now authoritative and carried by the by-value clone; `scheme := "http"; if variantPlan.useTLS { … }` is a second derivation that must stay in sync by hand. `scheme := variantPlan.scheme` collapses it. (Confirmed harmless today: `applyFuzzGRPCPosition` rejects `"scheme"` as a fuzz path, so it cannot diverge per variant.)

**F-6 (NIT, out of scope — suggest a follow-up Issue)** — Same defect class in the sibling layer, pre-existing and untouched here. `buildHTTPMessage` (`grpcweb/channel.go:1060-1070`) only copies `Authority`/`Scheme` when `srcEnv.Message` is an `*HTTPMessage` — but on the gRPC-Web Send path `srcEnv` is `c.sendCtxEnv`, whose Message is a `*GRPCStartMessage`, so `msg.Authority` stays empty. `httpaggregator/send.go:106` then feeds that empty value to `H2HeadersEvent.Authority`, and `appendRequestPseudoHeaders` omits `:authority` entirely. So gRPC-Web over an HTTP/2 upstream (`WrapH2UpstreamForDispatchFull`, `ProtocolGRPCWeb` branch) forwards no `:authority` to an authority-routed upstream. Worth filing alongside USK-1053/USK-1054 rather than expanding this PR.

### Claim check

- ✅ Helper renames + nil guards preserved — true (guards unreachable, see F-2).
- ✅ `resendGRPCPlan.scheme` set alongside `plan.useTLS`; builder sets `Authority`/`Scheme` — true.
- ✅ `cloneFuzzGRPCPlan` nils `canonicalURL` (`fuzz_grpc_helpers.go:1482`) and copies both new fields via `out := *base` — verified exactly as claimed.
- ✅ New test inventory + `//go:build e2e` (smoke) on `grpc_integration_test.go` — verified.
- ⚠️ `make lint` 0 issues — **not verified**: golangci-lint v2.13.2 (the pinned version in `.golangci-lint-version`) is not on PATH here and CLAUDE.md forbids `go install`-ing it. `gofmt -l .` is clean and `go vet` is clean on all affected packages including `-tags e2e`.
- ⚠️ `make test` 4208 PASS / 0 FAIL — **partially confirmed**. My whole-tree `make test` reported 4009 `--- PASS` and then failed with `panic: test timed out after 4m0s` in `internal/mcp`. Re-running `internal/mcp` alone (`go test -race -timeout 4m`) passes in **221s of the 240s budget**, so the failure was contention from my own concurrent test runs, not a PR regression. Flagging only because the headroom is thin — pre-existing, and the new mcp test costs 0.00s.
- ✅ `make test-e2e-smoke` — ran the smoke tier for the affected packages (all `ok`) and the **full** e2e tier for `internal/mcp` (`ok`, 137s). Did not run the whole-tree smoke tier.

---

PR #123, opened 2026-09-15
fix(layer/grpcweb): carry the observed :authority and :scheme through the L7 overlay

usk6666:
## Summary
- A gRPC-Web request relayed over an **HTTP/2 upstream leg** reached the server with **no `:authority` at all**, and on h2c with `:scheme` **rewritten** from the observed `http` to `https`. USK-1051's defect class, one Layer over.
- Root cause is two-sided and both halves were broken: the grpcweb Layer never *populated* the USK-920 L7 overlay on `GRPCStartMessage`, and its Send half read that overlay from a type assertion that can never succeed.
- **Producer** (`refillFromHTTPMessage` → new `applyRequestOverlay`): project `:authority` / `:scheme` / `:path` off the inbound request `HTTPMessage`, mirroring `internal/layer/grpc.buildStartMessage` verbatim, gated on `dir == Send`.
- **Consumer** (`buildHTTPMessage`, Send branch): read `start.Authority` / `start.Scheme`. **No `Context.TargetHost` fallback**, no TLS-derived `:scheme` fallback.
- Scope widened from `:authority` to `:authority` + `:scheme` by user decision — same root cause, same lines, same test. `Path` is projected too; it moves no wire byte but makes `Flow.URL` exact for gRPC-Web start flows.

## The Issue's proposed fix was a no-op

The Issue body read the defect as "`buildHTTPMessage` only copies `Authority` when the source Envelope is an `HTTPMessage`", implying a one-line fallback to `start.Authority`. That fixes nothing, for two independent reasons:

1. **`start.Authority` is always empty.** USK-920 added `Authority` / `Scheme` / `Path` to `envelope.GRPCStartMessage` but wired only the `layer/grpc` producer. Before this PR, a grep for `.Authority =` / `Authority:` across the non-test files of `internal/layer/grpcweb/` matched *only inside the dead branch below*. Nothing ever wrote the field.
2. **The branch it would have replaced is structurally dead.** `buildHTTPMessage` did `srcEnv.Message.(*envelope.HTTPMessage)`, but `srcEnv` is `c.sendCtxEnv`, which has exactly one write site — `sendStartLocked` (`channel.go:570`) — reachable only from the `case *envelope.GRPCStartMessage:` arm of `Send`'s type switch. `srcEnv.Message` is therefore *always* a `*GRPCStartMessage`. The assertion succeeds 0% of the time, live and under test.

So the dead reads are **deleted**, not demoted to a fallback: an unexercisable precedence rule is worse than none (USK-1051 F-2). The `RawQuery` read that lived on the same dead branch is dropped with them — it was never reachable, dropping it changes no behaviour, and Send-side `RawQuery` adoption belongs to USK-1053.

## Why `:scheme` is in scope

`internal/layer/http2`'s `appendRequestPseudoHeaders` **defaults an empty `:scheme` to `"https"`**. So the empty overlay did not merely drop the value on h2c — it *actively rewrote* an observed `:scheme: http` to `https` on the upstream wire. That is a stronger violation of MITM Principle #1 than the `:authority` drop, it is USK-1051 defect #2 reproduced, and it lives on the same four lines with the same test. h2c gRPC-Web is reachable via `connect_inner_dispatch.go`.

## Mutation proof

Every new assertion was demonstrated red against unfixed code.

| Mutation applied to production code | Result |
|---|---|
| Part A producer call removed | `TestRefill_PopulatesRequestOverlayFromHTTPMessage` **FAIL** — `start.Authority = ""`, `start.Scheme = ""`, `start.Path = ""` |
| `dir != envelope.Send` gate removed from `applyRequestOverlay` | `TestRefill_ResponseSideLeavesOverlayEmpty` **FAIL** — `start.Authority = "api.example.test"` on a Receive-direction message |
| Pre-fix consumer restored (the dead `srcEnv` assertion) | `TestSend_AuthorityFromOverlayReachesHTTPMessage` **FAIL** in 2 of 3 sub-cases — `HTTPMessage.Authority = ""`, `HTTPMessage.Scheme = ""` |
| Forbidden `srcEnv.Context.TargetHost` fallback added | `TestSend_AuthorityFromOverlayReachesHTTPMessage/empty_overlay_stays_empty` **FAIL** — `HTTPMessage.Authority = "connect-target.invalid:443"` |
| Whole production file reverted to `main` | `TestGRPCWeb_BinaryOverHTTP2_RoundTrip` **FAIL** `upstream saw :authority = "", want "127.0.0.1:45699"`; `TestGRPCWeb_Base64OverHTTP2_RoundTrip` **FAIL** likewise; **both HTTP/1.x round-trips still PASS** |

That last row is doing double duty: it proves the H2 assertion is load-bearing *and* executes Decision 6's argument that HTTP/1.x is untouched.

## Test-oracle lesson

The H2 gRPC-Web path has been emitting a non-conformant wire **for its entire life while the suite stayed green**. `x/net/http2`'s server validates only `:method` / `:scheme` / `:path` and silently tolerates a missing `:authority`, and nothing in the suite inspected `r.Host`. Reachability was never in doubt — the oracle simply did not look. The new smoke assertion (`r.Host == upAddr`, ~15 LOC, zero harness surgery) closes that gap, and the HTTP/1.x twin (`Host:` header from the raw request bytes) converts "this cannot regress HTTP/1.x" from prose into an executable guarantee.

## HTTP/1.x is unaffected — two independent reasons

1. On HTTP/1.x the authority is an ordinary `Host` header, so it survives inside `start.Metadata` (`stripStartHeaders` drops only pseudo-headers and the four gRPC control headers) and is re-emitted verbatim by `rebuildStartHeaders`.
2. The http1 send path **never reads `HTTPMessage.Authority`** — `sendRequestSynthetic` and `encodeRequestSynthetic` build the request line from `Path`/`RawQuery` and the header block from `msg.Headers` alone. A grep for `.Authority` across `internal/layer/http1/*.go` hits test files only.

## Out of scope, deliberately untouched

- The Receive-branch `Status: 200` / `"OK"` hardcode — same dead assertion, but a **separate defect** (an upstream 404/502 is relayed to the client as 200). Owned by its own Issue.
- gRPC-Web `:path` / `RawQuery` Send-side adoption — USK-1053's sibling gap.
- `http2.options.scheme` being write-only — separate Issue.

## Test plan
- [x] `make lint` — gofmt clean, golangci-lint v2.13.2, **0 issues** (`applyRequestOverlay` extracted as a helper so `refillFromHTTPMessage` stays under the gocyclo ceiling)
- [x] `make build`
- [x] `make test` (fast tier) — **0 failures**
- [x] `make test-e2e-smoke` (merge gate) — **0 failures**, 49 packages ok
- [x] 3 new fast-tier tests (5 assertions groups incl. 3 table sub-cases), each demonstrated red against unfixed code
- [x] 2 new smoke-tier assertions (`runRoundTripHTTP2` + regression guard in `runRoundTripHTTP1`), covering 4 existing round-trip tests
- [x] `TestRoundTrip_BinaryHTTP1Style` still green — the overlay travels in typed fields, never as a re-injected pseudo-header in `Metadata` (also asserted directly by the new producer test)
- [x] No `Envelope.Context.TargetHost` anywhere in the change

Resolves USK-1057
Linear: https://linear.app/usk6666/issue/USK-1057


usk6666 (commented):
## Security Review: APPROVED

The change forwards an attacker-controlled `:authority` / `:scheme` verbatim, which is the correct MITM behaviour and introduces no injection, panic, or resource exposure. HPACK's length-prefixed string encoding structurally forecloses CRLF smuggling on the HTTP/2 path, and `HTTPMessage.Authority` is never read by the line-delimited HTTP/1.x encoder, so there is no downgrade injection surface. No new validation gap relative to fields already flowing unvalidated on the same path, no new dependencies.

### Injection analysis

**1. HPACK injection surface (settled: not exploitable).**
Traced `start.Authority` (`internal/layer/grpcweb/channel.go:1122`) -> `msg.Authority` -> `internal/layer/httpaggregator/send.go:106` -> `H2HeadersEvent.Authority` -> `internal/layer/http2/channel.go:714` -> the in-repo encoder at `internal/layer/http2/hpack/encoder.go:137`. `encodeString` emits `encodeInteger(dst, 0x00, 7, uint64(len(s)))` followed by the raw octets (or a Huffman literal with the same length prefix). A value containing CR, LF, NUL, or a colon is carried as opaque payload inside a counted field and cannot terminate the field or synthesize a second one. Note the encoder is the repository's own `internal/layer/http2/hpack`, not `x/net` — both are length-prefixed. Response splitting on the H2 upstream wire is therefore not reachable.

**2. Line-delimited (HTTP/1.x) path: `HTTPMessage.Authority` never reaches it.**
A repo-wide grep for `.Authority` outside tests returns zero hits in `internal/layer/http1/`. `sendRequestSynthetic` (`internal/layer/http1/channel.go:1297-1325`) builds the request line from `msg.Method` / `msg.Path` / `msg.RawQuery` and the header block from `msg.Headers` via `serializeHeaders`; `Authority` is not consulted. The PR comment's claim that the HTTP/1.x leg re-emits the preserved `Host` out of `Metadata` is accurate — `isStartStripName` (`channel.go:842`) strips only pseudo-headers and the four gRPC content headers, so `host` survives verbatim. The added e2e assertion converts that into an executable guarantee.

Adjacent check: `msg.Path` *is* read by `serializeRequestLine`, and `parseServiceMethod` does not reject control characters, so an h2-sourced `:path` carrying CRLF would in principle split an HTTP/1.x request line. Per CLAUDE.md MITM Principle 6 I checked the precondition and it is absent — it requires an h2 client paired with an h1 upstream in one stack, which the architecture does not construct: the sniff-first MITM path advertises a single-element ALPN equal to upstream's pick (`mitmAdvertiseFromUpstreamPick`), the legacy fallback uses refresh-on-mismatch redial, and both TCP-forward dispatchers require matched transports (`tcp_forward_h2.go:230` asserts the upstream topmost is `*http2.Layer`; `tcp_forward_l7.go:212` pairs an h1 client with `upstreamH1 *http1.Layer`). Not a finding against this PR, and not reachable today. Re-open trigger: if a mixed h2-client / h1-upstream L7 dispatch is ever added.

**3. Panic safety on pathological values: clean.**
`applyRequestOverlay` (`channel.go:497-502`) is three unconditional string assignments; the consumer at `channel.go:1122-1123` is two more. No indexing, slicing, or splitting is introduced. Downstream, `appendRequestPseudoHeaders` only does `evt.Authority != ""`; `DynamicTable.Add` (`hpack/dynamic_table.go:40`) implements RFC 7541 4.4 correctly for an entry larger than the table (empties and declines to add, no unbounded growth); the writer enforces the peer's `SETTINGS_MAX_HEADER_LIST_SIZE` and splits into CONTINUATION frames (`http2/writer.go:221-237`, `296-316`). Empty is handled by the `hasAuthority` gate. An IPv6 literal, a bare colon, and non-UTF-8 bytes are all inert: Go strings hold arbitrary octets and the only structural consumer is `url.URL{Host: ...}` in `record_step.go:1309`, a struct literal with no parsing. Input length is bounded upstream by the decoder's `SetMaxHeaderListSize` (`http2/layer.go:607-609`), so `HeaderField.Size()`'s uint32 arithmetic cannot overflow.

**4. `hasAuthority` gate and attacker-chosen `:scheme`: correct, no downstream problem.**
`appendRequestPseudoHeaders` (`http2/channel.go:693-735`) omits `:authority` entirely when empty, which is what preserves a client's own omission and is what the "empty overlay stays empty" sub-case pins. `:scheme` can now be any value the client sent — the assembler's `applyPseudo` records pseudo-header values without validating them. That is wire-faithful and harmless: RFC 9113 permits any scheme token, the value is HPACK-encoded as a counted literal, and the only other consumer is `url.URL{Scheme: ...}` on the recording path. On the HTTP/1.x client leg the value is proxy-derived, not attacker-derived (`http1/channel.go:835` sets `Scheme: c.scheme` from `WithScheme`). Confirmed the overlay travels in typed fields only — `stripStartHeaders` defensively drops any name beginning with `:`, so no pseudo-header can be appended after the regular headers, which would otherwise be an HPACK protocol error.

**5. Existing unvalidated flow on the same path (so "adds no new validation" is not a finding).**
`Service` / `Method` (from `parseServiceMethod`, tolerant by design), `ContentType`, `Encoding`, and every `Metadata` value already reach the same HPACK encoder unvalidated via `rebuildStartHeaders`. This PR adds two more strings to a path that carries a dozen. Consistent with MITM Principle 1 and with the already-merged native-gRPC sibling (`internal/layer/grpc/channel.go:1481` `authorityForStart`), which this mirrors exactly.

**6. Memory / retention: no leak.**
`Authority` / `Scheme` / `Path` are Go strings. There is no `unsafe` anywhere under `internal/layer/`, so no string aliases a large read buffer; the hpack decoder and the HTTP/1.x parser both produce fresh allocations. `GRPCStartMessage` was already retained via `c.sendStart` until `sendEndLocked`; three more small strings add bytes, not buffer retention.

### Additional checks

- **Dead-assertion deletion is safe.** `srcEnv` is `c.sendCtxEnv`, assigned in `sendStartLocked` from an envelope whose `Message` is a `*GRPCStartMessage` by construction of the `Send` type switch. The deleted `srcEnv.Message.(*envelope.HTTPMessage)` could never succeed. The surviving Receive-branch assertion is dead for the same reason, which is exactly USK-1062 and consistent with it being tracked separately.
- **No scope or policy bypass.** `HTTPScopeStep` is `*HTTPMessage`-typed and never sees a gRPC-Web envelope; `extractScopeFields`'s `*GRPCStartMessage` branch (`internal/flow/record_scope.go:187-201`) deliberately does not read `Authority`, falling back to connection-level `TargetHost` / SNI. Capture-scope and target-scope decisions are unchanged. No rule engine under `internal/rules/` matches on gRPC authority.
- **Newly-populated `Flow.URL` does not create an SSRF path.** `projectGRPCStart` now fills `Flow.URL` for gRPC-Web as it already did for native gRPC, making `resend_grpc { flow_id }` able to recover a target. That target still passes `checkResendGRPCScope`, which gates both the canonical URL and the dial address against `TargetScope` (`internal/mcp/resend_grpc_helpers.go:682-690`). Net effect is a scope check against the correct host where previously there was no recoverable host at all.
- **Concurrency.** `applyRequestOverlay` mutates a function-local `start` before it is published to `emitQueue` under `c.mu`; the `sendStart` / `sendCtxEnv` mutex discipline is untouched. No new shared state.
- **Plugins.** `internal/pluginv2/convert.go:395` already carries `Authority` / `Scheme` / `Path` across a `GRPCStartMessage` rebuild, so a plugin mutating an unrelated field cannot silently erase the new overlay.
- **Dependencies.** No `go.mod` / `go.sum` change.

### Verification

`make lint` (gofmt + golangci-lint v2.13.2): 0 issues. `make test` did not finish inside the review window — three concurrent `-race` suites from parallel review sessions saturated the machine — so CI remains the authoritative gate for the fast tier. The lint run compiles the default build configuration, which covers both `channel.go` and `channel_test.go`.

**STATS:** CRITICAL 0, HIGH 0, MEDIUM 0, LOW 0

---

usk6666 (commented):
## Code Review: APPROVED ✅

The two-sided fix is correct and I independently re-derived every load-bearing claim in the PR body against the source: `sendEndLocked` sets `srcEnv = c.sendCtxEnv` from the Start envelope, so `srcEnv.Message` is always `*GRPCStartMessage` and the deleted Send-branch assertion was structurally dead; `assembler.go:284-296` really does gate `evt.Authority/Scheme/Path` on `direction == Send`; `appendRequestPseudoHeaders` really does omit `:authority` when empty and default `:scheme` to `"https"`; `http1`'s send path (`sendRequestSynthetic` / `sendRequestOpaque`) never reads `HTTPMessage.Authority`; and `host` is not in `connectionSpecificHeaders`, so the "client sent only `host:`" escape hatch in the new comment is real. Mutation-tested all three fast-tier tests plus the h2 e2e — each fails against the unfixed line — and the whole thing is lint-clean (`gocyclo` included; inlining the helper reproduces `complexity 16 > 15`, so the extraction rationale checks out). Only LOW/NIT findings, none blocking.

### Findings

| F-1 | LOW | internal/layer/grpcweb/grpcweb_integration_test.go | 845-860 | TestQuality | The PR silently flips grpc-web `Flow.URL` from `nil` to populated and `Stream.Scheme` from `""` to the observed scheme; nothing asserts it |
| F-2 | NIT | internal/layer/grpcweb/grpcweb_integration_test.go | 848-857 | TestQuality | The HTTP/1.x `upstreamHost` assertion passes against pre-fix code (invariance guard, not a USK-1057 regression pin) |
| F-3 | NIT | internal/envelope/grpc.go | 27-51 | CodeHealth | `GRPCStartMessage.Authority/Scheme/Path` godoc still describes an HTTP/2-HEADERS-only origin; this PR adds an HTTP/1.x `Host`-header producer |

### Suggestions

**F-1** — Populating the overlay changes recorded output, not just the wire. I probed it both ways in this worktree:

```
with fix:    send-start Flow.URL = https://127.0.0.1:39865/pkg.Echo/Say ; Stream.Scheme = "https"
without fix: send-start Flow.URL = <nil>                                ; Stream.Scheme = ""
```

That runs through `record_step.go:1299` (`if m.Authority != "" || m.Scheme != ""` → `fl.URL`) and `createStream`'s `*GRPCStartMessage` case (`st.Scheme = msg.Scheme`). It is the desirable USK-920 behaviour and it is what makes `resend_grpc { flow_id }`-style recovery and the WebUI URL column work for grpc-web — but it is an MCP-surface change with zero coverage, and CLAUDE.md's e2e subsystem checklist calls out MCP/query retrievability explicitly. The sibling PR did assert it (`TestGRPC_ClientAuthorityReachesUpstreamVerbatim` checks `sendStart[0].URL.Host`). `runRoundTripHTTP1` already has `flows` and a `firstFlowWithEvent(flows, "send", "start")` call in hand, so this is ~3 lines:

```go
if startF := firstFlowWithEvent(flows, "send", "start"); startF == nil {
	t.Errorf("missing send-start flow; flows=%d", len(flows))
} else if startF.URL == nil || startF.URL.Host != target {
	t.Errorf("send-start Flow.URL = %v, want host %q (USK-920 projection)", startF.URL, target)
}
```

**F-2** — I verified this empirically: stubbing out `applyRequestOverlay` + the `buildHTTPMessage` reads leaves both HTTP/1.x round-trips green while both HTTP/2 ones fail with `upstream saw :authority = ""`. So it does not meet the "must fail against unfixed code" bar — by design, since it is asserting that nothing moved. It is still worth keeping: adding `"host"` to `isStartStripName`'s D7 set makes it fail with `upstream saw Host = ""`, so it genuinely pins the `stripStartHeaders → Metadata → rebuildStartHeaders` path the HTTP/1.x leg depends on. Suggestion is wording only — call it an invariance guard rather than a "regression guard", so a future reader does not mistake it for a USK-1057 pin.

**F-3** — Outside the diff, so purely FYI / follow-up material. The new `applyRequestOverlay` godoc gets the dual origin right ("on HTTP/2 `msg.Authority` originates from the client's `:authority`; on HTTP/1.x it originates from the Host header"), but the shared type's field doc still says "from the request-side HEADERS frame" and scopes `Scheme` to `("http" for h2c, "https" for TLS+ALPN h2)`. After this PR `http1.parseRequestURI`'s `headers.Get("Host")` and the http1 Layer's configured `c.scheme` are a second, equally load-bearing source. Worth one sentence when USK-1053 or USK-1062 next touches that file.

### Scrutiny points

**(a) Is `dir == envelope.Send` the right seam?** Yes, and the claim checks out at the source: `buildHeadersEvent` (assembler.go:284-296) only assigns `evt.Method/Scheme/Authority/Path` under `if direction == envelope.Send`, and `http1`'s response constructors (channel.go:921, :995) set neither `Scheme` nor `Authority`, while `parseRequestURI` is called from the request path only (channel.go:777). So today the gate is a no-op on every real producer — which is exactly why gating in the producer is better than gating in `buildHTTPMessage`: it keeps the invariant "a Receive-side `GRPCStartMessage` carries no request pseudo-headers" true for *all* consumers (RecordStep's `Flow.URL` projection, plugins, rules), not just the Send path. And `TestRefill_ResponseSideLeavesOverlayEmpty` makes the gate's deletion visible, which is the pattern that was missing in the earlier unidirectional-guard episodes.

**(b) Dropping the `RawQuery` read.** Nothing real is lost. It sat inside the same `srcEnv.Message.(*envelope.HTTPMessage)` assertion that can never succeed, so it never executed; and even a "repaired" version could not have carried a query, because `GRPCStartMessage` has no `RawQuery` field and `refillFromHTTPMessage` never projects `msg.RawQuery` anywhere. `buildHTTPMessage` rebuilds `msg.Path` from Service/Method and leaves `RawQuery` empty both before and after this PR, so the observable behaviour (query dropped on the upstream leg — USK-1053) is unchanged. Keeping a dead assignment would have re-created exactly the unexercisable-precedence-rule problem USK-1051's F-2 removed.

**(c) Is the h2c `:scheme` case under-tested?** Acceptable as-is. The Layer's own contribution — "forward the observed scheme verbatim, do not rewrite `http` to `https`" — is fully pinned by the `h2c scheme is not rewritten to https` subtest, and I confirmed it fails when the `msg.Scheme = start.Scheme` line is removed. What is *not* covered end-to-end is `appendRequestPseudoHeaders` emitting a non-`https` scheme onto the wire, but that is `internal/layer/http2`'s invariant, not grpcweb's, and it is uncovered there too (every `:scheme` in `internal/layer/http2/*_test.go` is `"https"`). Notably the merged sibling made the identical call — USK-1051's e2e asserts `:authority` on the wire and explicitly defers `:scheme` to a unit test ("grpc-go classifies `:authority` as reserved but whitelisted … unlike `:scheme`, which is reserved and NOT whitelisted — assert that one in the unit test instead"). Closing it properly means an h2c wire test in `internal/layer/http2`, which is nearer USK-1063's territory than this PR's.

**(d) Helper extraction.** Clean and correctly placed. `applyRequestOverlay` is a free function that touches no `channel` state, so it needs no lock — and the call site is right: `start` is a fully-built local that has not yet been published to `emitQueue` (the `c.mu.Lock()` comes ~30 lines later), and it precedes both consumers of `start`, the `emitAnomalyStart` early-return path and the normal `buildEnvelope` path, so the overlay is present on the anomaly Start too. Nil-safety holds without new guards: `start` is a fresh composite literal, and a typed-nil `msg` would already have panicked in `materializeBody` several lines earlier, so no new panic surface is introduced. The gocyclo rationale is factually correct — I inlined the gate and `golangci-lint --enable-only gocyclo` reports `cyclomatic complexity 16 of func (*channel).refillFromHTTPMessage is high (> 15)`.

### Test efficacy (mutation-verified)

| Test | Production line pinned | Result against mutated code |
|---|---|---|
| `TestRefill_PopulatesRequestOverlayFromHTTPMessage` | `applyRequestOverlay(start, msg, dir)` call site (channel.go:340) | FAIL — all three of Authority/Scheme/Path read `""` |
| `TestRefill_ResponseSideLeavesOverlayEmpty` | the `if dir != envelope.Send { return }` gate | FAIL — all three leak request-shaped values onto the response overlay |
| `TestSend_AuthorityFromOverlayReachesHTTPMessage` (subtests 1-2) | `msg.Authority = start.Authority` / `msg.Scheme = start.Scheme` | FAIL |
| `TestSend_AuthorityFromOverlayReachesHTTPMessage` (subtest 3, empty overlay) | *absence* of a `Context.TargetHost` fallback | FAIL when a fallback is injected; passes pre-fix by construction (correct for a negative guard) |
| `TestGRPCWeb_{Binary,Base64}OverHTTP2_RoundTrip` | the full producer→consumer→h2-encoder chain | FAIL — `upstream saw :authority = "", want "127.0.0.1:NNNNN"` |
| `TestGRPCWeb_{Binary,Base64}OverHTTP1_RoundTrip` | `host` surviving `stripStartHeaders` → `rebuildStartHeaders` | PASS pre-fix (see F-2); FAIL when `host` is added to the D7 strip set |

`TestRoundTrip_BinaryHTTP1Style` is still green and still asserts `:authority` is absent from `start.Metadata` — the fix did not satisfy itself by re-injecting a pseudo-header, and the new `TestRefill_*` test independently re-asserts that for the typed-field path.

### Verification run

`PATH=/home/user/go/bin:$PATH make lint` → 0 issues. `internal/layer/grpcweb` green on the fast tier, the smoke tier, and the full e2e tier (`-tags e2e`, 144s); `internal/proxybuild` full e2e green (87s); `internal/mcptest` grpc-web smoke green. `make test` reported `internal/flow` and `internal/mcp` as failures, but both are bare 4-minute package timeouts with zero `--- FAIL` lines anywhere in the run — `internal/flow` passes in isolation at 208s on this machine. Environmental, unrelated to this change.

---

PR #28, opened 2026-05-20
feat(mcp): migrate user-facing playbooks to MCP Prompts

usk6666:
## Summary

Migrates the user-facing vulnerability-verification playbooks from the Claude-Code-only `.claude/skills/yorishiro/` directory into MCP Prompts. Implements stage 1 + stage 2 of USK-949 — both prompts and the legacy `install skills` subcommand coexist; stage 3 (full deletion of `install skills`) is deferred to a follow-up Issue 1 minor version later.

- Adds `prompts/list` + `prompts/get` with **9** playbooks embedded via `//go:embed prompts/*.md` so they version-track the binary (no more version skew).
- Argument substitution uses `{{arg_name}}` double-brace placeholders (distinct from Go `text/template` so the model is not misled). Substitution happens at `GetPrompt` time via `strings.ReplaceAll`. Required-arg validation rejects empty / missing values.
- `install skills` still runs; first line of `installSkills()` now prints a `[DEPRECATED]` notice pointing at `prompts/list` and noting the one-minor-version removal timeline.

### Final prompt set

| Name | Purpose |
|------|---------|
| `verify-idor` | IDOR / privilege-escalation sweep (body / path / header positions) |
| `verify-sqli` | SQL injection (time-based / error-based / UNION) with method-aware safety guardrails |
| `verify-xss` | Reflected XSS via `YP_`-prefixed marker payloads |
| `verify-csrf` | CSRF token validation check (replace / empty / remove) |
| `audit-auth` | Authentication bypass + authorization (role-downgrade) audit |
| `fuzz-endpoint` | General-purpose fuzz workflow across HTTP / WS / gRPC / Raw |
| `replay-with-mods` | Single-shot replay with header / body modifications via `resend_*` |
| `capture-traffic` | playwright-cli capture procedure with scoped recording filter |
| `stateful-fuzz-loop` | Setup → resend → teardown loop for stateful endpoints via macros |

## Decision rationale

**Option 1 (inline-expansion) chosen per draft §5.7** — the model does not actively fetch MCP Resources in current Claude Code usage, so Option 2 (Resource-URI references) and Option 3 (hybrid) rely on a fetch that does not happen. Option 1 ships the full playbook body inline as `TextContent` on `GetPrompt`, paying body cost only at call time.

[...]

PR #124, opened 2026-09-15
test(protoschema): drive the real reflection dialer end to end

usk6666:
## Summary
- Add `internal/encoding/protoschema/reflection_wire_integration_test.go` (`//go:build e2e`) — four tests that drive `Discover` through `defaultReflectionDialFunc` (TCP dial → `http2.New(ClientRole)` → `OpenStream` → `grpclayer.Wrap` → `buildReflectionStartEnvelope`) against a **real grpc-go server-reflection endpoint**. That production dialer had never executed under test: every other test in the package installs a synthetic dialer over the package-level `reflectionDial` seam, which is why the USK-1051 `:authority` omission went undetected on the `grpc_schema discover` path.
- A `reflect.Value.Pointer()` guard (`assertProductionDialer`) asserts the seam still points at `defaultReflectionDialFunc` immediately before each `Discover` call. This is the direct answer to the Issue title — the file cannot silently regress into testing a fake.
- **Test D** (`TestDiscover_RealWire_DialFailureSurfacesTransportError`) covers the production dialer's **transport** error path — a dial against a loopback port that was bound only long enough to learn its number and then released. Tests A–C all reach the server, so `dialReflectionUpstream`'s `dial %s: %w` wrap (and the sibling `http2 layer` / `open stream` / `send Start` / `tls handshake` wraps) had no executing test at all; Test C's UNIMPLEMENTED fallback is a *protocol* error, not a transport one. The assertion pins the two wraps this codebase owns — `reflection (v1): dial <addr>: ` — and explicitly not the OS errno tail, plus a guard that a transport error does **not** trigger the v1alpha fallback.
- **Doc-only** correction to `internal/encoding/protoschema/reflection.go`'s package comment (no code change) — see below.
- No `go.mod` / `go.sum` change: `google.golang.org/grpc` is already a direct require and its root package is already linked into the shipped binary via `grpc_reflection_v1`'s generated service stub.

### Test harness notes
- Real grpc-go server with a custom `reflection.ServiceInfoProvider` + a `protodesc.NewFiles`-resolved synthetic `FileDescriptorProto`. The *descriptor content* is reused verbatim from the existing `buildTestFileDescriptorProto` fixture — no new descriptor is defined — but that helper returns **marshalled bytes** while this harness needs a `*descriptorpb.FileDescriptorProto`, so a small unmarshal wrapper (`newGreeterFileDescriptor`, ~13 lines) was added. `protoregistry.GlobalFiles` stays untouched.
- Default proto codec on purpose: the sibling resend/fuzz harnesses install a raw `[]byte` codec via `grpc.ForceServerCodec`, which is server-wide and would break the reflection service's own protobuf marshalling.
- No `net/http` / `httptest` — `net.Listen` + `grpc.Server.Serve` directly (MITM Principle #4). `internal/mcptest/grpc_upstream.go` was deliberately not copied.
- The listener binds `127.0.0.1:0` while the caller targets `localhost:<port>`, so an `:authority` derived from the socket would read `127.0.0.1:<port>` — the assertion pins "the caller-supplied target", not "whatever the socket happened to be".
- The h2c tap's base listener gets its own `t.Cleanup(func() { _ = base.Close() })` at the call site, because `startReflectionUpstream` registers its Close only after two `t.Fatalf`-capable steps. The resulting double Close on the happy path is harmless — `net.TCPListener.Close` is idempotent and the second `ErrClosed` is discarded — and the code says so, so it does not get "tidied" back into a leak.

#### File size accounting
The design review sized this file at 300–380 LOC. As reviewed it was **575 lines = 142 comment-only + 54 blank + 379 code**; with Test D it is **656 lines = 185 comment-only + 60 blank + 411 code**. The gap against the estimate is therefore *not* explained by comment density alone: the design review's component breakdown sums to ~330 lines, so roughly **49 lines of genuine code** sit outside it. Every one of them is load-bearing — no dead code, no speculative abstraction:

| Component | ~LOC | Why |
|---|---|---|
| `reflectionCallObserver` | 29 | `grpc.StreamInterceptor` is the only seam into grpc-go's own reflection handler |
| `assertGreeterDiscovered` | 45 | shared content assertion for the assembled `FileDescriptorSet` |
| `assertSingleMetadataValue` | 12 | exactly-one-value, byte-for-byte metadata check |
| `newGreeterFileDescriptor` | 13 | unmarshal wrapper over the reused fixture (see above) |
| `formatHeaderFields` | 13 | ordered-field rendering for the Test B failure output |
| import block | 28 | — |

### e2e Subsystem Verification Checklist applicability
Covered: **Communication success**, **Raw bytes / wire fidelity** (h2c byte tap), **Error paths** in both senses — the *protocol* error path (v1 → v1alpha UNIMPLEMENTED fallback, Test C) and the *transport* error path (a refused dial inside `defaultReflectionDialFunc` / `dialReflectionUpstream`, Test D). N/A by design and stated in the file's doc comment: Stream recording, Flow recording, State transitions, Plugin hook firing, Variant recording, MCP `query` retrieval — `grpc_schema discover` is a control-plane outbound dial that runs no Record/Plugin Pipeline step and persists no Flow; its only Pipeline is a lone `pipeline.NewBudgetStep` (`internal/mcp/grpc_schema_discover.go`), and the reflection chatter is deliberately not recorded (USK-928 Resolved #28 / U1).

## Tier promotion (deviation from CLAUDE.md default)
This file uses plain `//go:build e2e` (merge-gate **smoke**) rather than the documented default `e2e && !e2e_smoke`. Justification:

1. A well-formed `:authority` is a **hard precondition** for every outbound gRPC dial against grpc-go >= 1.83.2 — without it the server early-aborts with `codes.Internal` before dispatching the handler.
2. It sits **one line from regression** (`authorityForStart(m)` / `schemeForStart(env, m, "https")` in `internal/layer/grpc/channel.go`).
3. The existing merge-gate wire guard (`internal/layer/grpc/grpc_integration_test.go`, `TestGRPC_ClientAuthorityReachesUpstreamVerbatim`) covers **only the MITM relay path**. All three synthetic producers — resend / fuzz / reflection discover — had **zero** merge-gate wire coverage.
4. Cost is negligible: the four tests total **0.09s** against `ci.yml`'s `-timeout 10m` / job `timeout-minutes: 15`.

## Mutation proof

Both controls were executed by temporarily reverting production lines in `internal/layer/grpc/channel.go`. **Neither mutation is committed** — this PR's diff touches only the two files listed above.

### A. `:authority` — pre-USK-1051 `Authority: env.Context.TargetHost`

`buildReflectionStartEnvelope` sets `Context: envelope.EnvelopeContext{ConnID: connID}` only, so `TargetHost` is empty and `appendRequestPseudoHeaders` omits the field. All three tests fail:

```
=== RUN   TestDiscover_RealWire_HTTPS_AuthorityAndSchema
    reflection_wire_integration_test.go:344: Discover over TLS: reflection (v1): read ListServices response: gRPC status 13: no host or :authority header present
--- FAIL: TestDiscover_RealWire_HTTPS_AuthorityAndSchema (0.01s)
=== RUN   TestDiscover_RealWire_H2C_RequestPseudoHeadersOnTheWire
    reflection_wire_integration_test.go:495: Discover over h2c: reflection (v1): read ListServices response: gRPC status 13: no host or :authority header present
--- FAIL: TestDiscover_RealWire_H2C_RequestPseudoHeadersOnTheWire (0.00s)
=== RUN   TestDiscover_RealWire_FallsBackToV1AlphaOnUnimplemented
    reflection_wire_integration_test.go:559: Discover with only v1alpha registered: reflection (v1): read ListServices response: gRPC status 13: no host or :authority header present
--- FAIL: TestDiscover_RealWire_FallsBackToV1AlphaOnUnimplemented (0.00s)
FAIL
```

### B. `:scheme` — pre-USK-1051 `schemeFromContext(env, "https")`

`env.Context.TLS` is nil for a synthetic producer, so the pre-fix body returns the `"https"` literal. grpc-go **does not validate `:scheme`**, so the h2c RPC still succeeds while lying on the wire — Tests A and C stay green and only the byte tap catches it. This is precisely the blind spot the tap exists to close:

```
=== RUN   TestDiscover_RealWire_HTTPS_AuthorityAndSchema
--- PASS: TestDiscover_RealWire_HTTPS_AuthorityAndSchema (0.02s)
=== RUN   TestDiscover_RealWire_H2C_RequestPseudoHeadersOnTheWire
    reflection_wire_integration_test.go:524: stream-1 HEADERS field[1] = ":scheme": "https", want ":scheme": "http"
         got: :method: POST, :scheme: https, :authority: localhost:33365, :path: /grpc.reflection.v1.ServerReflection/ServerReflectionInfo, content-type: application/grpc+proto, te: trailers
        want: :method: POST, :scheme: http, :authority: localhost:33365, :path: /grpc.reflection.v1.ServerReflection/ServerReflectionInfo, content-type: application/grpc+proto, te: trailers
--- FAIL: TestDiscover_RealWire_H2C_RequestPseudoHeadersOnTheWire (0.01s)
=== RUN   TestDiscover_RealWire_FallsBackToV1AlphaOnUnimplemented
--- PASS: TestDiscover_RealWire_FallsBackToV1AlphaOnUnimplemented (0.01s)
FAIL
```

## Function-var seam audit
All four package-level `var xFunc = defaultX` seams outside tests were audited. `reflectionDial` is the **only** instance of the dangerous class — one that hides a multi-component network/transport path behind a swap. The other three replace trivially-reachable filesystem/OS probes:

| Seam | What it hides |
|------|---------------|
| `internal/mcpserver/serverjson.go:31` `serverJSONPathFunc` | `os.UserHomeDir` + `filepath.Join` |
| `internal/setup/browser_detect.go:78` `fileExistsFunc` | `os.Stat` |
| `internal/setup/browser_detect.go:115` `containerCheckFunc` | container indicator probe |
| `internal/encoding/protoschema/reflection.go:169` `reflectionDial` | **TCP dial + H2 Layer + gRPC Layer + Start envelope** |

No follow-up Issue needed. The seam itself is deliberately left in place — its two `internal/mcp` consumers give the MCP handler fast-tier coverage with zero network, which is legitimate; the new guard is what closes the hole.

## `reflection.go` doc-only correction
The package comment claimed:

> NO runtime dependency on google.golang.org/grpc is introduced; only the generated proto types under google.golang.org/grpc/reflection/grpc_reflection_v1 (and v1alpha) are imported

This is factually wrong and was verified so: `go list -deps ./cmd/yorishiro-proxy | grep -c '^google.golang.org/grpc'` reports **65** packages, including the bare root `google.golang.org/grpc`, pulled in transitively by `grpc_reflection_v1`'s generated service stub. Apache-2.0 and on the allow-list, so there is no licence or policy problem — but the comment misled anyone reasoning about the dependency surface. The corrected text keeps the point the comment was actually making (the *transport* is the in-house Layer stack, which remains true) and states the linkage reality. **No code changed.**

## Test plan
- [x] `make lint` — gofmt clean, golangci-lint v2.13.2, 0 issues
- [x] `make build` — vet + build OK
- [x] `make test` (fast tier) — see note below
- [x] `make test-e2e-smoke` — **EXIT=0**, 49 packages ok, 0 failures. The four new tests: `TestDiscover_RealWire_HTTPS_AuthorityAndSchema` PASS (0.04s), `TestDiscover_RealWire_H2C_RequestPseudoHeadersOnTheWire` PASS (0.01s), `TestDiscover_RealWire_FallsBackToV1AlphaOnUnimplemented` PASS (0.04s), `TestDiscover_RealWire_DialFailureSurfacesTransportError` PASS (0.00s)
- [x] Mutation proof A executed and recorded; mutation reverted
- [x] Mutation proof B executed and recorded; mutation reverted
- [x] Flake check — 15x all four tests, plus 40x the v1alpha fallback test (its teardown races the server's post-trailers `RST_STREAM`) and 40x the new dial-failure test; zero failures
- [x] Test D vacuity control — temporarily left the probe listener **bound** instead of releasing it, confirming the test FAILS (`Discover error = "reflection (v1): read ListServices response: context deadline exceeded", want prefix "reflection (v1): dial 127.0.0.1:40281: "`). It is genuinely detecting a dial failure, not passing on any-error. Control reverted; the final diff touches only the two intended files.
- [x] No `t.Logf` standing in for an assertion; no `t.Parallel()` (the `reflectionDial` seam is global)
- [x] No `go.mod` / `go.sum` change

> **Fast-tier note**: on the dev box `make test` reported `FAIL internal/flow 240.275s` and `FAIL internal/mcp 240.445s`. Both are `panic: test timed out after 4m0s`, not assertion failures — several sibling agent sessions were running `go test ./...` concurrently (load average ~25 on 4 cores; `TestLargeExportImport` alone took 2m3s). Re-run in isolation with a longer timeout, both pass: `ok internal/flow 288.424s`, `ok internal/mcp 473.162s` — i.e. each genuinely needed more than the 240s budget under that load. Neither package depends on `protoschema` (`go list -deps ./internal/flow | grep -c protoschema` -> 0), and the only non-test change in this PR is a comment.

Resolves USK-1054
Linear: https://linear.app/usk6666/issue/USK-1054


usk6666 (commented):

Test-only PR; no findings at any severity. The new file is `//go:build e2e`, confirmed excluded from the default build (`go list` reports it in `TestGoFiles` only under `-tags e2e` / `-tags 'e2e e2e_smoke'`, and not at all without a tag), so none of it is compiled into the shipped binary or reachable from attacker-controlled traffic.

**(a) Do `cert.CA.Generate` / `Issuer.GetCertificate` touch the filesystem?** **No — both are purely in-memory.** `CA.Generate` (`internal/cert/ca.go:41`) generates an ECDSA P-256 key with `crypto/rand`, self-signs, and stores the result in struct fields under a mutex; it never calls into `paths.go`. The only `os.WriteFile` calls in the package are in `CA.Save` (`ca.go:171/176`), which this test never invokes, and `DefaultCADir` (`~/.yorishiro-proxy/ca/`) is never reached. `internal/cert/issuer.go` does not import `os` at all (zero filesystem calls). No key material is written to disk, inside or outside a temp dir — so the absence of `t.TempDir()` here is correct, not an omission.

**(b) Do all listeners bind loopback?** **Yes.** Both `net.Listen` calls in the file (lines 184 and 473) are `"tcp", "127.0.0.1:0"`. The third listener path is `teeListener`, which wraps the line-473 loopback listener. No `0.0.0.0`, bare `":0"`, or `[::]` anywhere in the file.

**(c) Does the doc-only edit weaken a security-relevant claim?** **No — it strengthens the record, and the surviving claim is TRUE.** `git diff` on `reflection.go` is comment lines exclusively; not one line of code changed. The edit removes a *false* "NO runtime dependency on google.golang.org/grpc is introduced" and replaces it with the verified reality (`go list -deps ./cmd/yorishiro-proxy` reports **65** `google.golang.org/grpc*` packages including the bare root). Admitting linked attack surface that was previously denied is a net transparency gain. Critically, the old comment's real point survives and was independently verified: `defaultReflectionDialFunc` (`reflection.go:303`) builds `net.Dialer` → optional `Transport.TLSConnect` → in-house `http2.New(..., http2.ClientRole)` → `OpenStream` → `grpclayer.Wrap`, and a repo-wide grep for `grpc.Dial` / `grpc.NewClient` / `grpc.DialContext` / `grpc.ClientConn` across **all non-test files returns zero hits**. grpc-go's client stack is genuinely never used to speak the wire.

**(d) Did the module graph change?** **No.** Zero `go.mod` / `go.sum` diff. `google.golang.org/grpc v1.83.2` is already a **direct** require, licensed **Apache-2.0** (allow-listed). Every external root reached by `go list -tags e2e -deps -test` on the package maps to a module already present in `go.mod`; no new module enters the graph. The newly imported server-reflection package `google.golang.org/grpc/reflection` is **absent** from `go list -deps ./cmd/yorishiro-proxy` (only `grpc_reflection_v1` is linked, and that was already the case on `main`), so the shipped binary's dependency set is byte-for-byte unaffected — the count is 65 on both `main` and this branch.

### Other criteria checked, all clean

- **`InsecureSkipVerify: true`** (line 342) is a per-instance field on a locally-constructed `&transport.StandardTransport{}` passed as `DiscoverOptions.Transport`. It mutates no package-level state and changes no production default; the pre-existing proxy-side defaults in `connpool.go` / `hosttls.go` are untouched. It is also semantically required — the test generates a throwaway CA no trust store knows. Confined to the `e2e`-tagged file.
- **Goroutine leak**: the single `go func() { _ = gs.Serve(ln) }()` has an explicit termination condition — `t.Cleanup` runs `gs.GracefulStop()` (which closes listeners and unblocks `Serve`) then `ln.Close()`. The client side is already torn down by `discoverOnce`'s `defer cleanup()` before cleanup runs, so `GracefulStop` cannot block.
- **Races**: `wireRecorder` and `reflectionCallObserver` are fully mutex-guarded, and both `snapshot()` methods return deep copies (`append([]byte(nil), ...)` and `md.Copy()`) rather than aliasing shared state. The new tests only *read* the `reflectionDial` package var (via `assertProductionDialer`) and none of them call `t.Parallel()`, so they cannot race the seam-mutating helpers in `reflection_test.go`.
- **Slice bounds** in `firstRequestHeaderFields`: the `len(raw) < len(http2.ClientPreface)` guard precedes both `raw[:24]` and `raw[24:]`, and `t.Fatalf` halts via `runtime.Goexit` with no fallthrough. The `len(raw) == 24` edge yields a valid empty slice. Panic is not reachable. The `frame.Reader` defaults to a 16 KiB `DefaultMaxFrameSize` and `hpack.NewDecoder(4096)` inherits default `maxHeaderListSize` / `maxStringLength` caps, so decoding is size-bounded — and the input is self-produced by the proxy's own HTTP/2 Layer, not attacker-shaped.
- **`crypto/rand` vs `math/rand`**: no `math/rand` import anywhere on the path; `internal/cert` uses `crypto/rand` for both the key and the 128-bit serial. No `unsafe`. No hardcoded credentials, tokens, `os.Getenv`/`Setenv`, `exec.*`, or external network egress.
- **`wireRecorder.append` unbounded growth**: assessed and dismissed — bounded by the test's own few-KB reflection exchange over a single connection with a test-scoped lifetime. Not worth a finding.
- **`MinVersion: tls.VersionTLS12`** is on the *test server*, not an upstream dial, so the CLAUDE.md MITM exception does not apply and a floor here is fine.

FINDINGS: None

STATS:
  CRITICAL: 0
  HIGH: 0
  MEDIUM: 0
  LOW: 0

---

usk6666 (commented):
## Code Review: APPROVED ✅

This PR closes a real structural hole: `defaultReflectionDialFunc` (TCP dial → `http2.New(ClientRole)` → `OpenStream` → `grpclayer.Wrap` → `buildReflectionStartEnvelope`) now executes end to end against a live grpc-go reflection server, and `assertProductionDialer` makes it structurally hard to silently regress back into testing the fake. The two mutation proofs in the PR body are the right kind of evidence — proof B in particular demonstrates that Test A/C alone would *not* have caught a `:scheme` regression, which is exactly why the h2c byte tap earns its place. Findings are 1 LOW + 5 NIT; nothing blocking.

**(a) Exact-ordered HPACK assertion — KEEP, do not relax.** I traced `appendRequestPseudoHeaders` (`internal/layer/http2/channel.go:723-730`, default Chrome-shaped `:method :scheme :authority :path`) and `buildStartHeaderKVs` (`internal/layer/grpc/channel.go:1355-1384`, `content-type` then `te`), and the 6-field `want` block matches the production encoder exactly. The strictness is a feature, not brittleness: this repo's north star is browser parity / fingerprint fidelity (M47/M48), so a `user-agent` silently appearing on a synthetic-producer dial is precisely the class of change that should require a deliberate, reviewed test update. The failure output prints the full ordered got/want lists, so the update is a 10-second diff. The `len != len` → `t.Fatalf` followed by per-index `t.Errorf` is also the right shape. Any relaxation (e.g. ordered-subsequence matching) would drop the "no unexpected header" guarantee, which is the half that actually matters here.

**(b) `reflect.Pointer` seam guard — reliable in this specific use.** The documented caveat ("not necessarily enough to identify a single function uniquely") bites for method values and for two closures minted from the same source func — neither applies. `defaultReflectionDialFunc` is a package-level func assigned to a package-level var of func type, so `Pointer()` returns the function's entry PC on both sides and compares equal. A fake installed via `SetReflectionDialForTest` / `installFakeDial` is an anonymous closure with a different entry PC, so the guard *would* fire. I also checked the ordering hazard: `reflection_test.go`'s fake-installing tests restore via `t.Cleanup` and none of them call `t.Parallel()`; the package's parallel tests (`registry_test.go`, `protoschema_test.go`, `protoc_runner_test.go`) never touch `reflectionDial` and, per `testing`'s semantics, resume only after all sequential top-level tests complete. So no spurious fire and no `-race` report on the unsynchronised var read.

**(c) LOC claim — not accurate as stated, though nothing improper is hiding.** Measured: 575 total = 142 comment-only + 54 blank + **379 code**. The design review's 330-line breakdown is exceeded by ~49 lines of *code*, so the delta is not "entirely comment density". Five components sit outside the named breakdown: `reflectionCallObserver` (the `grpc.StreamInterceptor` observer, ~29 lines), `assertGreeterDiscovered` (~45), `assertSingleMetadataValue` (~12), `newGreeterFileDescriptor` (~13 — which contradicts the "fixture 0" / "zero new fixture code" claim, since the untagged fixture returns marshalled bytes and the harness needs a `*descriptorpb.FileDescriptorProto`), and `formatHeaderFields` (~13), plus a 28-line import block the estimate never accounted for. Every one of these is load-bearing for the three tests — no dead code, no speculative abstraction, no YAGNI — so this is a reporting-accuracy NIT (F-2), not a code problem.

**(d) `localhost` / `::1` flake risk — not a real risk; leave as is.** Go's `net.Dialer` does dual-stack Happy Eyeballs: if `::1` is ordered first it gets an immediate `ECONNREFUSED` (nothing bound there) and the dialer falls straight over to `127.0.0.1` well inside the 10s `reflectionWireTimeout`. It would fail hard only on a host whose `/etc/hosts` gives `localhost` no A record at all — not the case on GitHub Actions ubuntu runners, where `::1` maps to `ip6-localhost`/`ip6-loopback` rather than `localhost`. And the `localhost:<port>` vs `127.0.0.1:<port>` split is load-bearing: I traced `buildReflectionStartEnvelope` (`reflection.go:632-650`, `Authority: opts.TargetAddr`) → `authorityForStart` (`grpc/channel.go:1481`, L7 overlay only, no `TargetHost` fallback per USK-1051) → `appendRequestPseudoHeaders`, and the assertion genuinely pins the caller-supplied target rather than the socket. Substituting a literal IP would destroy the discrimination, so no change is warranted.

**Other verifications:** doc-only change to `reflection.go` confirmed (every `+`/`-` line is `//`-prefixed); `go list -deps ./cmd/yorishiro-proxy | grep -c '^google.golang.org/grpc'` → **65**, with the bare root present, so the corrected comment is factually right; `google.golang.org/grpc v1.83.2` is a direct (non-indirect) require and no `go.mod`/`go.sum` change; no `net/http` / `httptest` import and `internal/mcptest/grpc_upstream.go`'s `*httptest.Server` shell was not copied (MITM Principle #4 satisfied); no `t.Logf`-as-assertion, no `t.Parallel()`, no TODO/FIXME/HACK; `t.Helper()` present on all seven `*testing.T` helpers; `gofmt -l` clean. Tier justification is stated in both the PR body and the file doc comment (lines 16-23). The N/A checklist rationale (lines 25-35) is accurate — I read `internal/mcp/grpc_schema_discover.go` and confirmed the only Pipeline is `pipeline.NewBudgetStep` with no Record/Plugin step and no Flow persisted on the success path. Concurrency: `wireRecorder`/`reflectionCallObserver` are fully mutex-guarded; both `snapshot()` calls happen after `Discover` returns (whose `discoverOnce` `defer cleanup()` has already closed ch/layer/conn), so a concurrent `append` can only add post-HEADERS bytes and cannot truncate the frame the tap parses; `t.Cleanup` LIFO after the body's `defer cancel()` is the correct ordering; and in Test B `ln` is the *wrapped* `teeListener`, whose promoted `Close` does close the base listener, so the double-close with `GracefulStop` is harmless and the base is not leaked on the success path.

### Findings

| F-1 | LOW | internal/encoding/protoschema/reflection_wire_integration_test.go | 325-575 | TestQuality | No production-dialer *transport* error-path test |
| F-2 | NIT | internal/encoding/protoschema/reflection_wire_integration_test.go | 1-575 | CodeHealth | PR-body LOC and "zero new fixture code" claims are inaccurate |
| F-3 | NIT | internal/encoding/protoschema/reflection_wire_integration_test.go | 336, 488, 552 | CodeHealth | Bare `30 * time.Second` repeated 3× beside a named timeout const |
| F-4 | NIT | internal/encoding/protoschema/reflection_wire_integration_test.go | 325 | TestQuality | Test A name reads as `:scheme` in a PR about `:scheme` |
| F-5 | NIT | internal/encoding/protoschema/reflection_wire_integration_test.go | 369-398 | TestQuality | `wireRecorder` concatenates all accepted conns into one buffer |
| F-6 | NIT | internal/encoding/protoschema/reflection_wire_integration_test.go | 473-483 | CodeHealth | `base` listener leaks if `startReflectionUpstream` fatals early |

### Suggestions

**F-1** — The happy path now drives every component the Issue named, but `defaultReflectionDialFunc`'s four error wraps (`dial %s: %w`, `http2 layer: %w`, `open stream: %w`, `send Start: %w`) and `dialReflectionUpstream`'s `tls handshake %s: %w` branch are still unexecuted. Test C covers a *protocol* error path (UNIMPLEMENTED fallback) but not a *transport* one, so the doc comment's "Error paths" row is only half met. Cheap addition (~10 lines), and it is discriminating because a dial failure is not UNIMPLEMENTED and therefore surfaces immediately as `reflection (v1): dial ...` rather than falling through to the both-variants message:

```go
func TestDiscover_RealWire_DialFailureSurfacesVerbatim(t *testing.T) {
	ln, err := net.Listen("tcp", "127.0.0.1:0")
	if err != nil {
		t.Fatalf("listen: %v", err)
	}
	addr := ln.Addr().String()
	_ = ln.Close() // nothing is bound on addr any more

	assertProductionDialer(t)
	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
	defer cancel()

	_, derr := Discover(ctx, DiscoverOptions{TargetAddr: addr, Scheme: "http", Timeout: reflectionWireTimeout})
	if derr == nil {
		t.Fatal("Discover against an unbound port returned nil error")
	}
	// Terminal on v1: a dial failure is not UNIMPLEMENTED, so there is no
	// v1alpha retry and the verbatim dial wrap must survive.
	if !strings.Contains(derr.Error(), "reflection (v1): dial "+addr) {
		t.Errorf("Discover error = %q, want it to carry the verbatim `reflection (v1): dial %s` wrap", derr, addr)
	}
}
```

Returning at all (rather than hanging to the 10s timeout) is itself the assertion that the dialer's `_ = conn.Close()` unwind does not deadlock. If the bind-then-close idiom is judged flaky under port reuse, `127.0.0.1:1` is an equally good refused target.

**F-2** — Reword the PR body: the delta over the design-review estimate is ~49 lines of code plus ~117 lines of in-function comments, not comment density alone, and `newGreeterFileDescriptor` is 13 lines of new fixture *adapter* code (the reused `buildTestFileDescriptorProto` returns marshalled bytes; the harness needs a `*descriptorpb.FileDescriptorProto`). Nothing needs to move in the code — this is only about the claim matching the artefact, per the "design review recommended text is unverified" lesson.

**F-3** — The file already models the right habit with `reflectionWireTimeout` (line 70) and its rationale comment. The three bare `30 * time.Second` outer-context deadlines are the belt to that braces; give them a name so the relationship is explicit:

```go
// reflectionWireCtxTimeout is the outer test-body deadline. It is
// deliberately looser than reflectionWireTimeout so that a hang is
// attributed to DiscoverOptions.Timeout, not to the test harness.
const reflectionWireCtxTimeout = 30 * time.Second
```

**F-4** — `TestDiscover_RealWire_HTTPS_AuthorityAndSchema` asserts `:authority`, `content-type`, `FullMethod` and the discovered *schema* — it asserts no `:scheme` at all (and could not: grpc-go discards it). In a PR whose central subject is `:scheme`, "Schema" is one keystroke from being misread as "Scheme", which would leave a future reader believing `:scheme` has TLS-path coverage it does not have. Suggest `TestDiscover_RealWire_HTTPS_AuthorityAndDiscoveredSchema`.

**F-5** — `teeListener.Accept` hands every accepted conn the *same* `wireRecorder`, so two connections would interleave into one byte slice and `firstRequestHeaderFields` would parse garbage. That is correct today (Test B registers v1 only, so `discoverOnce` dials exactly once), but the sibling Test C path dials twice, and the failure mode would present as a confusing hpack decode error rather than "you reused the tap wrong". A one-line guard on the doc comment is enough:

```go
// teeListener wraps every accepted connection in a teeConn. All accepted
// connections share one wireRecorder, so this tap is only valid for a
// single-connection scenario — the v1 -> v1alpha fallback dials twice and
// would interleave two frame streams into one buffer.
```

Alternatively key the recorder per accepted conn and expose `snapshot(i int)`, but that is more machinery than the single-connection case warrants.

**F-6** — `base` is created at line 473 and only becomes owned by a `t.Cleanup` once `startReflectionUpstream` reaches line 197. Between the two, `protodesc.NewFiles` (line 152) and the `SplitHostPort` (line 189) can `t.Fatalf`, leaking the socket for the remainder of the test binary. One line at the call site closes it:

```go
base, err := net.Listen("tcp", "127.0.0.1:0")
if err != nil {
	t.Fatalf("listen: %v", err)
}
t.Cleanup(func() { _ = base.Close() }) // idempotent w.r.t. GracefulStop's own close
```

---

PR #125, opened 2026-09-15
fix(mcp): take the resend_grpc TLS decision from observed transport, not the declared scheme

usk6666:
## Summary

- `resend_grpc { flow_id }` decided its socket entirely from `Flow.URL`, which `record_step.go projectGRPCStart` projects verbatim from `GRPCStartMessage.Authority` / `.Scheme` — i.e. from the client's HEADERS block. Recording them unvalidated is correct (MITM Principle 1); dialling from them is not. Under the default (open) TargetScope there was no intervening gate, so a `:scheme: http` sent inside an otherwise TLS session replayed the recorded metadata — `authorization` included — in cleartext on port 80.
- The TLS decision now consults `Stream.ConnInfo.TLSVersion`, which `RecordStep.updateStreamTLS` writes from the **upstream leg's** TLS snapshot. `loadResendGRPCFlows` already calls `GetStream`, so this costs no extra I/O. The oracle is **strictly one-sided**: non-empty ⇒ the proxy really completed a TLS handshake with the upstream ⇒ upgrade the dial. Empty means only "TLS was not observed" (it also covers h2c and streams that never got a response) and never downgrades. An explicit `scheme` input always wins.
- **`plan.scheme` is not rewritten.** The resent HEADERS frame still carries the recorded `:scheme`, so a byte-diff of the recorded flow and of the outgoing frame is unchanged — only the TCP/TLS handshake differs. The two facts are separated by *layer* (recorded L7 overlay vs recorded L4 transport), not by overwriting one with the other. Rewriting the scheme would reverse USK-1051.
- Host axis: warned, not changed (see corrections below). Surfaced in the tool result's `warnings[]`; logged at Debug, because it fires on every normal Mode A call and CLAUDE.md's Log Level Guidelines demote a Warn that frequent. The TLS-upgrade advisory gets `warnings[]` **and** `slog.Warn`, which is genuinely abnormal.
- `fuzz_grpc`'s per-variant canonical-URL rebuild re-derived the scheme from `variantPlan.useTLS`. That was equivalent only while those two could not diverge. Extracted as `rebuildFuzzGRPCCanonicalURL` and switched to `variantPlan.scheme`, so the URL the safety/scope filters check cannot drift from the URL the wire expresses (the USK-1051 failure mode).
- Contract documented in the `resend_grpc` tool description, the `target_addr` / `scheme` schema text, and `help_resend_grpc.md` (new "What decides the dial in Mode A" section + `warnings[]` result field).

## Scope: this is a PARTIAL fix

**Three-protocol defect on the host half**: `resend_grpc`, `resend_http`, and `resend_ws` all recover a **client-declared** `:authority` and dial it. Only `resend_grpc` is fixed here; follow-up Issues for `resend_http` and `resend_ws` are being filed in the same session. `resend_raw` is immune — `target_addr` is required (`resend_raw.go:51`).

**The TLS half is HTTP/2-origin-only.** HTTP/1.x's recorded scheme is proxy-observed (the Layer-configured value at `internal/layer/http1/channel.go:833-835`), not client-declared; HTTP/2's comes from the client (`internal/layer/httpaggregator/channel.go:324-327`). `Authority` is client-declared on both.

**Explicitly not modified**: `internal/flow/` (no schema change — `server_addr` exists but this fix does not use it), `resolveResendGRPCDialTarget`, `checkResendGRPCScope`.

## Corrections to the Issue body

1. **Option A ("the `Flow` side has the CONNECT target recorded") is false today.** `createStream` never populates `ConnInfo` (`internal/pipeline/record_step.go:853-877`) and `envelopeToFlow` never records `TargetHost`. The only production writers of `ConnectionInfo.ServerAddr` are the non-MITM TLS-passthrough recorder (`internal/proxybuild/passthrough_recorder.go:121`) and the TLS-handshake-*error* audit Stream (`internal/proxybuild/builder.go:2306-2310`). With no persisted address there is nothing better to dial — hence warn-only on the host axis, with `target_addr` as the pin.
2. **The LOW-severity rationale ("TargetScope mitigates it") only covers the host half.** The sharper finding: `:authority` spoofing is not required at all. With the *correct* authority and only `:scheme: http`, the replay hits the right host on port 80 in cleartext, and a hostname-only TargetScope rule passes it. `checkTargetScopeAddr` is inert here too — it fires only when `dialAddr != canonicalURL.Host`, which in the no-override path happens purely because `resolveResendGRPCDialTarget` appends the default port.
3. **Accurate framing is CWE-918 + CWE-319, not cross-principal credential theft.** In the single-client case the party that spoofed `:scheme` also authored the `authorization` header.

Additionally, `Stream.Scheme` is **not** a usable independent fallback oracle: `createStream` sets it from the same client-declared `GRPCStartMessage.Scheme`, so the existing USK-920 fallback inherits the defect rather than correcting it. Commented in place.

## Mutation proof

Every new assertion was demonstrated to fail against unfixed code. Seven mutations, each applied in isolation to the production files (tests untouched):

| # | Mutation | Result |
|---|----------|--------|
| M1 | Delete the `applyResendGRPCDialGroundTruth(ctx, input, plan)` call | `useTLS = false, want true`; `dialAddr = "api.example.com:80", want "api.example.com:443"`; both warnings absent. 2 tests / 3 subtests FAIL |
| M2 | Revert `rebuildFuzzGRPCCanonicalURL` to derive the scheme from `useTLS` | `variant canonicalURL.Scheme = "https", want "http"` + base/variant URL disagreement. FAIL |
| M3 | Drop the `input.Scheme == ""` and `!plan.useTLS` guards | `explicit_http_override_is_never_upgraded`: `useTLS = true, want false`; `dialAddr = ...:443, want ...:80`. FAIL |
| M4 | Also set `plan.scheme = "https"` on upgrade (the anti-distortion guard) | `plan.scheme = "https", want "http"`; `canonicalURL.Scheme` likewise; `GRPCStartMessage.Scheme = "https", want "http"`. FAIL |
| M5 | Make the oracle two-sided (`useTLS = observedTLSVersion != ""`) | `declared_https_with_no_conninfo_stays_tls`, `empty_recovered_scheme_still_defaults_to_https`, `uppercase_declared_scheme_is_normalised_by_the_store` all FAIL with `useTLS = false, want true` |
| M6 | Move the reconciliation *after* `resolveResendGRPCDialTarget` | `dialAddr = "api.example.com:80", want "api.example.com:443"` — pins the ordering hazard (the default port follows `useTLS`). FAIL |
| M7 | Never read `Stream.ConnInfo` (`if false && stream.ConnInfo != nil`) | `useTLS = false, want true` + missing TLS warning. FAIL |

M1 and M4 were re-run against the final test file after the store consolidation described below; both still FAIL.

## Test plan

- [x] `TestBuildResendGRPCPlan_DialTLSFromObservedTransport` — 7 cases through the **real SQLite store**: the upgrade case (asserting `useTLS`, *and* that `plan.scheme` / `canonicalURL.Scheme` / `GRPCStartMessage.Scheme` all stay `"http"`, *and* that the recovered `authorization` is still replayed); nil `ConnInfo`; explicit-`http` override; empty recovered scheme still defaulting to https (decision B); recovered https with no observation (never downgrade); uppercase `HTTPS` normalised by the store; `target_addr` redirect still upgrading the transport.
- [x] The real store is load-bearing, not incidental: `Flow.URL` is persisted as `f.URL.String()` and read back through `url.Parse`, and `net/url` lowercases the scheme on that read. A mock store would bypass the normalisation and assert behaviour production never exhibits. The uppercase case pins exactly that, which is why an `:scheme: HTTPS` bypass of the exact-match `scheme == "https"` is not live.
- [x] `TestBuildResendGRPCPlan_WarnsOnClientDeclaredDialTarget` — 3 cases: `flow_id` without `target_addr` warns; with `target_addr` silent; from-scratch silent.
- [x] `TestRebuildFuzzGRPCCanonicalURL_FollowsPlanSchemeNotUseTLS` — 3 cases through `cloneFuzzGRPCPlan` + the rebuild, including the diverged `scheme="http" / useTLS=true` plan.
- [x] USK-920 acceptance gate: `go test -race -tags e2e -run TestResendGRPCListenerCapture ./internal/mcp/` → `ok, 5.106s`. Its upstream leg is plaintext h2c, so `ConnInfo` is nil and nothing upgrades.
- [x] Wider gRPC e2e: `-tags e2e -run GRPC ./internal/mcp/ ./internal/mcptest/` → both `ok`.
- [x] `make lint` (gofmt + golangci-lint v2.13.2) → `0 issues`. `make build` → clean. `make test` → all packages pass, `internal/mcp` 221.8s (under the Makefile's 4m budget).

### Runtime note

The first draft created a fresh `newTestStore` per subtest (11 SQLite schema migrations). Under `-race` that pushed `internal/mcp` past the Makefile's `-timeout 4m`. Consolidated to one store per table — each case still gets its own stream id, and the parent's `t.Cleanup(Close)` runs only after all parallel subtests finish.

## Known limitation

In a `tcp_forwards` topology with a **plaintext client leg** and `upstream_tls: true`, the recorded `:authority` is the forward listener while the observed TLS belongs to the upstream beyond it, so the upgrade attempts TLS against a plaintext listener. It is loudly warned in `warnings[]` + `slog.Warn`, fails with a clean handshake error rather than silent corruption, and `scheme: "http"` overrides it. Gating on a client-side signal instead (e.g. `ConnInfo.TLSClientJA3`) would under-detect the actual vulnerability, which is the worse trade. Documented in `help_resend_grpc.md`.

Resolves USK-1056
Linear: https://linear.app/usk6666/issue/USK-1056


usk6666 (commented):
## Review Gate: APPROVED ✅ (1 fix round)

**Code Review: APPROVED** (1 MEDIUM, 3 LOW, 2 NIT) · **Security Review: APPROVED** (2 LOW) · **MITM-rejected: none**

All 7 in-scope findings fixed in `d2908d14`. F-6 (`internal/flow/types.go` godoc) deliberately excluded as out-of-diff-scope → follow-up.

### The central invariant was re-verified, not assumed

`plan.scheme` is written exactly once (before the oracle) and never again; `plan.useTLS` has exactly two writes, the base assignment and a bare `= true`. **Downgrade is impossible by construction.**

I re-applied the M4 mutation (also set `plan.scheme = "https"` on upgrade) against the **fixed** head and observed it still fail on both upgrade sub-cases across all three anti-distortion assertions (`plan.scheme`, `canonicalURL.Scheme`, `GRPCStartMessage.Scheme`). The guard on this PR's whole design is live.

### Mutations re-run by hand by the gate

| # | Reverted | Observed |
|---|----------|----------|
| M4 | rewrite `plan.scheme` on upgrade | FAIL ×2 subtests × 3 assertions |
| M5 | two-sided oracle | FAIL on exactly the 3 named sub-cases |
| M2 | fuzz rebuild via `useTLS` | FAIL — variant URL disagrees with base |
| M6 | oracle after dial-target resolution | FAIL — `:80` not `:443` |

### Gates, as observed

`make lint` 0 issues · `make build` clean · `make test` **EXIT=0, 49 ok, 0 FAIL** (`internal/mcp` 160.9s, `internal/flow` 109.8s — inside budget) · **USK-920 gate** `TestResendGRPCListenerCapture` 4/4 PASS, including the new `warnings[]` assertion.

### Notes on the judgement calls

- **Log levels**: the implementer's deviation from the design review is **correct**. `slog.Debug` for a notice that fires on every normal Mode A call matches CLAUDE.md's own "intercept rule match → Debug" precedent, and the actionable party still receives it via `warnings[]`.
- **Accepted `tcp_forwards` limitation**: failure is clean — conn closed, wrapped handshake error, **no cleartext fallback, no retry**, nothing written to the socket before `http2.New`. F-1 made it self-diagnosing rather than pinning it with a disproportionate topology test.
- **Residual gap, on the record**: `updateStreamTLS` fires only on `Direction == Receive`, so a client controlling the recorded session can suppress the oracle by never letting a response be observed. Documented in-code; follow-up filed. Never worse than `main`.

---
Automated review gate by yorishiro-proxy Review Gate

PR #126, opened 2026-09-15
fix(layer/grpc): make the observed :path authoritative when it is unparseable

usk6666:
## Summary

`sendStart` rebuilt `:path` from `GRPCStartMessage.Service`/`.Method` and never read the USK-920 `Path` overlay, so a malformed `:path` observed on the wire was **normalised on its way to the upstream**. The verified chain:

| Step | Site | Behaviour |
|---|---|---|
| 1 | `internal/layer/http2/assembler.go` `applyPseudo` | accepts `:path: /NoSlash` — no content validation, no anomaly |
| 2 | `internal/layer/http2/assembler.go` `splitPath` | → `Path="/NoSlash"`, `RawQuery=""` |
| 3 | `internal/layer/grpc/channel.go` `buildStartMessage` | `msg.Path="/NoSlash"`; `parseGRPCPath` → `("","")` |
| 4 | `internal/layer/grpc/channel.go` `sendStart` | `buildGRPCPath("","")` → `"/"` |
| 5 | `internal/layer/http2/channel.go` `appendRequestPseudoHeaders` | emits `:path: /` |

The client sent `/NoSlash`; the upstream received `/`. Every query-bearing `:path` also lost its query, because `GRPCStartMessage` had no home for it.

Affected path is **the live MITM relay only** — `resend_grpc` / `fuzz_grpc` never populate `Path`, so they could not be affected. This is the mirror image of USK-1051, where resend/fuzz *were* the risk.

## The precedence rule

> The derived `Service`/`Method` view is authoritative whenever the wire's `:path` was parseable into it; when the parse failed, the wire overlay is the only faithful representation and wins.

```go
func pathForStart(m *envelope.GRPCStartMessage) string {
	if m.Path != "" {
		if svc, method := parseGRPCPath(m.Path); svc == "" && method == "" {
			return m.Path
		}
	}
	return buildGRPCPath(m.Service, m.Method)
}
```

This is Architecture Principle 3 ("L7 parsing is an overlay on raw bytes") applied to the Send direction, and every branch is exercisable.

**`applyGRPCStartModify` needs no change — the Issue's blocker #1 dissolves.** On any well-formed `/S/M` the path is parseable, so `Service`/`Method` always win and the intercept `service`/`method` override behaves exactly as before. The blocker was an artefact of assuming "Path always wins". `internal/mcp/intercept_typed.go` is untouched, and no `path` parameter is added to the MCP intercept schema (new MCP surface, own CRLF guard, no evidenced demand).

`pathForStart` deliberately carries **no nil guard**, unlike its `authorityForStart` / `schemeForStart` siblings: `sendStart` is the sole call site and rejects a typed-nil `*GRPCStartMessage` before reaching it (pinned by the existing `TestChannel_SendTypedNilStartMessageReturnsError`). An extra guard would be dead defence — USK-1051 review F-2.

## Why no producer changes are needed

The rule makes this structural rather than incidental:

| Producer | Sets `Path`? | Outcome |
|---|---|---|
| `layer/grpc/channel.go` `buildStartMessage` | Yes — `evt.Path` | the fixed path |
| `layer/grpcweb/channel.go` | No | unchanged (→ USK-1069) |
| `mcp/resend_grpc_helpers.go` (`fuzz_grpc` shares it) | **No** | falls through to `buildGRPCPath` — byte-identical to today |
| `mcp/grpc_schema_discover.go` | Yes — `"/" + ReflectionV1Service + "/" + ReflectionMethod` | parseable and consistent with its Service/Method → rebuild reproduces it |
| `encoding/protoschema/reflection.go` | Yes — same shape | same |

A naive "Path always wins" rule **would** have regressed resend/fuzz to `:path: /` — the exact USK-1051 defect class. The empty-`Path` fall-through is what prevents it.

`internal/job/grpc_source.go`'s `GRPCResendSpec.Start` is assigned verbatim and needs no change.

## Mutation proof (verbatim)

Every new assertion has a named production line whose reversion kills it. **None of the mutations are committed** — verified with `git diff main --stat` (8 files, all intended).

**1 — revert `sendStart` to `pathPseudo := buildGRPCPath(m.Service, m.Method)`**

```
--- FAIL: TestChannel_SendStartPathRoundTrip (0.00s)
    --- FAIL: TestChannel_SendStartPathRoundTrip/trailing_slash_survives (0.00s)
        channel_test.go:1819: H2HeadersEvent.Path = "/", want "/pkg.Svc/Do/" (parseGRPCPath rejects an empty method segment)
    --- FAIL: TestChannel_SendStartPathRoundTrip/leading_double_slash_survives (0.00s)
        channel_test.go:1819: H2HeadersEvent.Path = "/", want "//Do" (leading "//" leaves an empty service segment, which parseGRPCPath rejects)
    --- FAIL: TestChannel_SendStartPathRoundTrip/no_inner_slash_survives (0.00s)
        channel_test.go:1819: H2HeadersEvent.Path = "/", want "/NoSlash" (core fix: parseGRPCPath cannot represent this, so the wire overlay wins)
    --- FAIL: TestChannel_SendStartPathRoundTrip/no_leading_slash_survives (0.00s)
        channel_test.go:1819: H2HeadersEvent.Path = "/", want "noslash" (parseGRPCPath rejects a missing leading slash)
FAIL
FAIL	github.com/usk6666/yorishiro-proxy/internal/layer/grpc	0.014s
```

(Run against the scaffolding-only build *before* the fix, the same mutation also fails both `TestPathForStart` unparseable rows — that was the initial live-bug reproduction.)

**2 — drop `RawQuery: m.RawQuery` from the `H2HeadersEvent` literal in `sendStart`**

```
--- FAIL: TestChannel_SendStartPathRoundTrip (0.00s)
    --- FAIL: TestChannel_SendStartPathRoundTrip/query_is_carried_across (0.00s)
        channel_test.go:1822: H2HeadersEvent.RawQuery = "", want "trace=1" (GRPCStartMessage.RawQuery is the new carrier for the query string)
FAIL
FAIL	github.com/usk6666/yorishiro-proxy/internal/layer/grpc	0.007s
```

**3 — drop `msg.RawQuery = evt.RawQuery` from `buildStartMessage`**

```
--- FAIL: TestChannel_SendStartPathRoundTrip (0.00s)
    --- FAIL: TestChannel_SendStartPathRoundTrip/query_is_carried_across (0.00s)
        channel_test.go:1798: Receive overlay = (Path "/pkg.Svc/Do", RawQuery ""), want ("/pkg.Svc/Do", "trace=1")
FAIL
FAIL	github.com/usk6666/yorishiro-proxy/internal/layer/grpc	0.006s
```

**4 — drop `RawQuery: m.RawQuery` from the `url.URL` literal in `projectGRPCStart`**

```
--- FAIL: TestRecordStep_GRPCStartProjectsURLAndScheme (0.00s)
    --- FAIL: TestRecordStep_GRPCStartProjectsURLAndScheme/query_bearing_path_projects_raw_query (0.00s)
        record_step_test.go:3369: flow URL.RawQuery = "", want "trace=1&debug"
FAIL
FAIL	github.com/usk6666/yorishiro-proxy/internal/pipeline	0.009s
```

**5 — drop `RawQuery: m.RawQuery` from the pluginv2 rebuild closure**

```
--- FAIL: TestGRPCStart_PseudoHeaderOverlaySurvivesMutation (0.00s)
    --- FAIL: TestGRPCStart_PseudoHeaderOverlaySurvivesMutation/metadata_mutation (0.00s)
        convert_test.go:262: RawQuery = "", want "trace=1" (must survive rebuild)
    --- FAIL: TestGRPCStart_PseudoHeaderOverlaySurvivesMutation/service_mutation (0.00s)
        convert_test.go:262: RawQuery = "", want "trace=1" (must survive rebuild)
    --- FAIL: TestGRPCStart_PseudoHeaderOverlaySurvivesMutation/encoding_mutation (0.00s)
        convert_test.go:262: RawQuery = "", want "trace=1" (must survive rebuild)
FAIL
FAIL	github.com/usk6666/yorishiro-proxy/internal/pluginv2	0.008s
```

## Residual losses (accepted, documented in RFC-001 §11)

Both live at the HTTP/2 Layer's `Path`/`RawQuery` split and are **shared verbatim with `HTTPMessage`** — not gRPC-specific, not fixable here:

1. **`:path: ""` is re-emitted as `/`** — `appendRequestPseudoHeaders` substitutes `/` for an empty path.
2. **A bare trailing `?` is destroyed** — `splitPath("/x?")` yields `RawQuery=""`, and the rejoin only re-adds `?` when `RawQuery` is non-empty.

Fixing either means changing the H2 event representation for *all* HTTP traffic. `Envelope.Raw` keeps the wire bytes in both cases. A re-open trigger is recorded in §11.

Percent-encoding and non-UTF-8 bytes round-trip fine: Go strings are byte slices and the in-house hpack `encodeString` is length-counted.

## No validation added on the relay path

Control characters already survive today through Service/Method with no help from this change — `parseGRPCPath("/a\r\nX: y/c")` → `Service="a\r\nX: y"`, `Method="c"`, and `buildGRPCPath` reproduces it exactly. Adopting `Path` on Send only extends that to the *no-inner-slash* subset.

The `:authority` / `:path` asymmetry does not widen here either. `:path` is the one field that reaches `serializeRequestLine` (via `HTTPMessage.Path`), but "an h2 client paired with an h1 upstream in one stack" is never constructed: `mitmAdvertiseFromUpstreamPick` advertises the upstream's *selected* protocol back to the client as a single-element list, `runHTTP1ExchangeLoop` pairs `clientH1`/`upstreamH1`, and tcp-forward asserts http1 Layers on both sides. Validating here would normalise what the wire did not normalise (MITM Principle 1) and would close no reachable hole.

The one real gap — `applyGRPCStartModify` lacking the CRLF guard its `resend_grpc` sibling has — is pre-existing, independent, and filed as **USK-1068**.

## Adjacent sites deliberately untouched

- **`internal/flow/record_scope.go` `extractScopeFields`** synthesises its path from Service/Method and never reads `Path`, so for exactly the malformed-`:path` flows this Issue targets, `path` is `""` during capture-scope matching. Changing it alters *what gets recorded* — a deliberate behaviour change deserving its own decision.
- **`internal/mcp/query_tool.go` `projectGRPCStartView`** exposes only Service/Method/Encoding/ContentType/Metadata on the intercept tool's held-envelope view, so an AI agent pausing a flow still cannot see the malformed path. Out of scope.
- **`grpcStartModified` in `internal/pipeline/record_step.go`** — `RawQuery` is **not** added to the variant-recording diff. That comparison deliberately omits Authority/Scheme/Path because they are wire-derived overlay, not user-mutated; consistency with its three siblings is the point. (`EncodeWireBytes` returns `nil, nil` for Start, so there is no variant-recording impact either way.)
- **`internal/layer/grpcweb/`** — never populates `Path`, so the rule has nothing to act on there. Deferred to **USK-1069**, blocked on USK-1057 merging (same file). Recorded in RFC-001 §11 with a re-open trigger.

Also confirmed safe: `createStream` reads `msg.Scheme` only; `resend_grpc_helpers.go` recovers service/method by splitting `Flow.URL.Path` and touches only `.Host`/`.Scheme`/`.Path`, so the added `.RawQuery` is ignored there.

## Test plan

Fast tier only. **No e2e test**: grpc-go always builds `/Service/Method`, so provoking a malformed one needs a hand-rolled HEADERS frame *and* a raw h2 upstream — disproportionate when the fast-tier invariant pins the property exactly.

`TestChannel_SendStartPathRoundTrip` (`internal/layer/grpc/channel_test.go`) drives each case in through `stubInner.pushHeaders` and back out through `Send`. The oracle is a round-trip invariant, not an enumeration:

| `evt.Path` / `RawQuery` | expected `(Path, RawQuery)` | pins |
|---|---|---|
| `/pkg.Svc/Do`, `""` | `/pkg.Svc/Do`, `""` | no regression, well-formed |
| `/pkg.Svc/Do`, `trace=1` | `/pkg.Svc/Do`, `trace=1` | **new `RawQuery` carriage** |
| `/NoSlash`, `""` | `/NoSlash`, `""` | **core fix** (today `/`) |
| `noslash`, `""` | `noslash`, `""` | no leading slash |
| `//Do`, `""` | `//Do`, `""` | leading `//`, one segment |
| `/pkg.Svc/Do/`, `""` | `/pkg.Svc/Do/`, `""` | trailing slash |
| `/a/b/c`, `""` | `/a/b/c`, `""` | multi-slash stays on the rebuild branch |
| `//pkg/Do`, `""` | `//pkg/Do`, `""` | leading `//`, two segments — parseable, rebuild branch |
| `""`, `""` | `/`, `""` | documented residual loss |
| Service/Method overridden post-`Next` | `/new.Svc/NewDo` | **the intercept-override guard — blocker #1** |
| Service+Method both cleared | `/` | U2's unchanged half |

Assertions are on the `H2HeadersEvent` `(Path, RawQuery)` pair, not a joined string: the `?` rejoin belongs to `appendRequestPseudoHeaders`, and re-implementing it in the test would only test the test.

Plus `TestPathForStart`, a direct unit table mirroring `TestAuthorityForStart_NilMessage` / `TestSchemeForStart_NilArgs` (no nil row, for the F-2 reason above); a `RawQuery` assertion and two new rows on `TestRecordStep_GRPCStartProjectsURLAndScheme`; `raw_query` added to the `TestGRPCStart_PseudoHeadersAreNotPluginMutableKeys` guard list (it must stay **off** the plugin hook surface, unlike `HTTPMessage`'s); and a `RawQuery` assertion on `TestGRPCStart_PseudoHeaderOverlaySurvivesMutation`.

`make lint` (v2.13.2, 0 issues) / `make build` / `make test` all pass. `internal/mcp` hit the documented USK-1048 load artefact in the full run (`panic: test timed out after 4m0s` at 240.271s with **zero** `--- FAIL:` lines); re-run in isolation it passes in **299.453s, 0 failures**. 48 packages ok, no real failures.

## Note on a stale design-review citation

The design review listed `//pkg/Do` (leading `//`) in the broken set. **The code disagrees and the code is authoritative**: `parseGRPCPath` splits at the *last* slash, so `//pkg/Do` → `Service="/pkg"`, `Method="Do"`, and `buildGRPCPath` rebuilds it identically — it is on the *rebuild* branch, not the overlay branch. A leading `//` is only unparseable when a single segment remains (`//Do`). Both shapes are now pinned as separate, contrasting table rows so the distinction cannot be lost again. The `parseGRPCPath` doc comment's own example list (`"/foo" → ("","")`) never claimed otherwise; the review over-generalised it.

Resolves USK-1053
Linear: https://linear.app/usk6666/issue/USK-1053


usk6666 (commented):
## Code Review: APPROVED

The precedence rule is the right shape and it is genuinely bidirectionally tested — I verified by mutation that inverting it, removing it, and removing its empty-`Path` fall-through each kill a distinct, named set of rows. `pathForStart`'s missing nil guard is correct (its single call site is dominated by `sendStart`'s typed-nil rejection), the `?` rejoin is properly left to the HTTP/2 Layer, and `Flow.URL.RawQuery` reaches no consumer that HTTP had not already been exercising through the identical construction. The findings are all documentation-accuracy and test-reach items; nothing blocks merge.

`make test` (full fast tier), `make lint` (golangci-lint v2.13.2, 0 issues) and `gofmt -l` all pass at `dee53163`.

### Mutation evidence (test efficacy)

| Mutation on production code | Result |
|---|---|
| `pathForStart` -> `return buildGRPCPath(m.Service, m.Method)` (rule **removed**) | KILLED — `TestPathForStart/{unparseable_path_wins, unparseable_path_wins_even_against_a_populated_derived_view}`, `TestChannel_SendStartPathRoundTrip/{no_inner_slash,no_leading_slash,leading_double_slash,trailing_slash}_survives` |
| `if m.Path != "" { return m.Path }` (rule **inverted**) | KILLED — `TestPathForStart/parseable_path_loses_to_the_derived_view`, `.../service_method_override_beats_a_parseable_path`, `.../clearing_service_and_method_on_a_parseable_path_emits_root` |
| drop the `m.Path != ""` guard (the **USK-1051 regression shape**) | KILLED — 8 rows incl. the pre-existing synthetic-producer test `TestChannel_SendStartPseudoHeadersComeFromMessage` |
| drop `RawQuery` from the Send-side `H2HeadersEvent` literal | KILLED — `query_is_carried_across` |
| drop `msg.RawQuery = evt.RawQuery` in `buildStartMessage` | KILLED — `query_is_carried_across` |
| drop `RawQuery` from `projectGRPCStart` | KILLED — `query_bearing_path_projects_raw_query` |
| drop `RawQuery` from the pluginv2 rebuild closure | KILLED — `TestGRPCStart_PseudoHeaderOverlaySurvivesMutation` (3 subtests) |
| drop `RawQuery` from `CloneMessage` | **SURVIVED** (see F-3) |
| `parseGRPCPath`: `strings.LastIndex` -> `strings.Index` | **SURVIVED** (see F-4) |

### Findings

| ID | Severity | File:Line | Category | Description |
|----|----------|-----------|----------|-------------|
| F-1 | MEDIUM | `internal/envelope/grpc.go:46-48`, `internal/pluginv2/convert.go:394-397` | Correctness (doc accuracy) | The override-compatibility guarantee is stated without its exception |
| F-2 | LOW | `internal/envelope/grpc.go:45-46` | Correctness (doc accuracy) | `leading "//"` listed as unparseable without the qualifier |
| F-3 | LOW | `internal/envelope/grpc.go:112` | TestQuality | New `CloneMessage` line is not covered by any test |
| F-4 | LOW | `internal/layer/grpc/channel_test.go:1733,1742` | TestQuality | Two rows narrate `parseGRPCPath` semantics that nothing asserts |
| F-5 | NIT | `docs/rfc/envelope.md:1481` | CodeHealth | Blank line splits the deferred-decisions list |

---

**F-1 — MEDIUM — the "exactly as before" claim omits its exception.**

`internal/envelope/grpc.go:46-48` says the rule "keeps the intercept / plugin service+method override working **exactly as before**", and `internal/pluginv2/convert.go:397` concludes "so there is **no stale-Path hazard**". Both are true only on the parseable subset.

On an unparseable `:path` the override is now silently discarded, where pre-USK-1053 it applied. Wire `:path: noslash` -> `Service=""`, `Method=""`; an intercept setting `service=pkg.Svc`, `method=Do` previously emitted `/pkg.Svc/Do` via `buildGRPCPath`, and now emits `noslash`. The PR's own row `unparseable_path_wins_even_against_a_populated_derived_view` (`internal/layer/grpc/channel_test.go:1865-1868`) pins exactly that. For `convert.go` the wording is the more pointed miss: carrying `Path` across the rebuild *is* the mechanism by which the plugin's `service`/`method` rewrite gets dropped there, so "no stale-Path hazard" describes the opposite of what happens.

This is **not** a request to change the behaviour — dropping the override on a malformed path is defensible MITM behaviour and `applyGRPCStartModify` is explicitly out of scope. It is a request to document it, so the next maintainer debugging "my intercept override did not apply" finds the answer in the comment rather than in a test row. Note `docs/rfc/envelope.md:243-246` already gets this right ("on any parseable `:path` the derived Service/Method view still wins, which is what keeps the ... override working") — it drops "exactly as before" and scopes the claim. Suggest aligning both Go comments with the RFC's wording and adding one sentence: *on an unparseable `:path` the overlay wins, so a `service`/`method` override is not applied there.*

**F-2 — LOW — `leading "//"` is not by itself unparseable.**

`internal/envelope/grpc.go:45-46` enumerates the unparseable shapes as `(no inner slash, leading "//", trailing "/", no leading "/")`. `//pkg/Do` has a leading `//` and is parseable: `parseGRPCPath` splits at the **last** slash, yielding `Service="/pkg"`, `Method="Do"`, which `buildGRPCPath` rebuilds identically — it sits on the rebuild branch, and the PR adds a table row saying so. `pathForStart`'s own comment (`internal/layer/grpc/channel.go:1500`) has the correct qualifier, "leading `//` **with a single remaining segment**". This is the same imprecision the commit message credits itself with correcting in the design review, surviving in the one comment on the public envelope type — the same comment USK-1051 already had to correct once. Suggest copying the qualifier from `pathForStart`.

**F-3 — LOW — the new `CloneMessage` line has no test behind it.**

Removing `RawQuery: m.RawQuery,` at `internal/envelope/grpc.go:112` leaves the entire suite green. This is a pre-existing class gap rather than a new one — `TestGRPCStartMessage_CloneMessage_DeepCopy` (`internal/envelope/grpc_test.go:29-79`) predates USK-920 and never covered `Authority` / `Scheme` / `Path` either — so the PR is consistency-preserving. It still misses the project's stated efficacy bar for a newly added line, and it matters: `CloneMessage` produces the *original* variant, so a dropped `RawQuery` would silently strip the query from the recorded original while `grpcStartModified` (correctly) never compares it, meaning no variant would flag the difference. Suggest adding all four overlay fields to that test's fixture and assertions in one pass.

**F-4 — LOW — the two "stays_on_the_rebuild_branch" rows are documentary, not discriminating.**

`multi_slash_stays_on_the_rebuild_branch` (`:1733`) and `leading_double_slash_with_two_segments_stays_on_the_rebuild_branch` (`:1742`) assert only the emitted `:path`, which for `/a/b/c` and `//pkg/Do` is identical on both branches — so neither the rule-removed nor the rule-inverted mutation touches them. That alone would be acceptable (they are cheap regression guards, and the names honestly say which branch they take; the commit message's mutation list correctly does not claim otherwise).

What makes it worth raising is that their `why` text asserts intermediate output nothing verifies: `Service="a/b"` and `Service="/pkg" Method="Do"`. Changing `strings.LastIndex` to `strings.Index` at `internal/layer/grpc/channel.go:1156` — which inverts precisely the last-slash semantics the whole `//Do` vs `//pkg/Do` asymmetry rests on, and which the commit message flags as the PR's key correction to the design review — leaves the whole `internal/layer/grpc` package green, `TestParseGRPCPath` included: that table (`internal/layer/grpc/channel_test.go:427-435`) has no row with two inner slashes, so it cannot tell the two implementations apart. Two rows there would close it: `{"/a/b/c", "a/b", "c"}` and `{"//pkg/Do", "/pkg", "Do"}`.

**F-5 — NIT — list formatting in RFC-001 section 11.**

The two new bullets at `docs/rfc/envelope.md:1481-1482` land after the blank line that previously terminated the "Post-N9 deferred design decisions" list, so the list now has a blank line in the middle. In CommonMark that makes the whole list loose (every item `<p>`-wrapped) and visually splits it into two groups for no semantic reason. Move them directly under the HTTP/3 / QUIC bullet.

### Verified, no finding

- **`pathForStart` nil guard asymmetry is correct.** `pathForStart` has exactly one production call site (`internal/layer/grpc/channel.go:922`), reached only after the `m == nil` rejection at `:911`; `sendStart` in turn has exactly one call site (`:886`, inside `Send`'s type switch, itself behind an `env.Message == nil` check). The guard strictly dominates. No nil-`m` path exists, so the sibling-symmetry complaint would be wrong.
- **The `?` rejoin belongs to the HTTP/2 Layer** and is correctly left there (`internal/layer/http2/channel.go:706-707`); the test asserts the `(Path, RawQuery)` pair rather than a pre-joined string, which is the right seam.
- **Both documented residual losses are real and accurately described**: `splitPath` (`internal/layer/http2/assembler.go:274-279`) destroys a bare trailing `?`, and `appendRequestPseudoHeaders:702-704` substitutes `/` for an empty path.
- **`RawQuery` is correctly absent from `grpcStartModified`**, consistent with its three overlay siblings.
- **No downstream `Flow.URL` consumer regresses.** HAR export — the only consumer that would change shape — skips gRPC streams entirely (`internal/flow/har.go:223-229`, pinned by `har_test.go:531`). `resend_grpc` splits only `URL.Path` (`internal/mcp/resend_grpc_helpers.go:433`). The query/manage tools stringify only. The one anchored URL filter (`internal/flow/sqlite.go:444-453`) already carries a `%://host?%` alternative for the HTTP case. And `envelopeToFlow`'s HTTP branch (`internal/pipeline/record_step.go:1113-1120`) has always built the identical `url.URL` with a populated `RawQuery`, so this is a well-trodden path.

*Out of scope, noted only in passing:* `internal/flow/types.go:171` still says `URL` is "Only set for HTTP send flows" — stale since USK-920, not touched by this PR.

---

usk6666 (commented):
## Security Review: APPROVED

**PR**: #126 — `fix(layer/grpc): make the observed :path authoritative when it is unparseable` (USK-1053)
**Head**: `dee53163987a7b9bede11ff8100f829706423099`

I traced every sink the newly-forwarded `:path` / `RawQuery` can reach. There is no reachable injection sink, no panic path, and no unbounded retention. Two LOW findings, both about **SafetyFilter target coverage** rather than a boundary bypass.

---

### 1. The HPACK path — confirmed safe by reading, and Huffman makes it *stricter*

`internal/layer/http2/hpack/encoder.go:136-149`:

```go
func (e *Encoder) encodeString(dst []byte, s string) []byte {
	if e.useHuffman {
		huffLen := huffmanEncodedLen([]byte(s))
		if huffLen < len(s) {
			dst = encodeInteger(dst, 0x80, 7, uint64(huffLen))
			dst = huffmanEncode(dst, []byte(s))
			return dst
		}
	}
	dst = encodeInteger(dst, 0x00, 7, uint64(len(s)))
	dst = append(dst, s...)
	return dst
}
```

Both branches emit an explicit length prefix (H-bit 1 or 0) followed by exactly that many bytes. There is no in-band delimiter, so CR / LF / NUL inside a `:path` is opaque payload and cannot terminate the field or the header block. HPACK stays binary and self-delimiting end to end — `writeHeaderBlock` (`internal/layer/http2/writer.go:296`) then splits the already-encoded block on frame boundaries, never on content.

Huffman **is** enabled: `hpack.NewEncoder(local.HeaderTableSize, true)` at `internal/layer/http2/layer.go:611`. That does not weaken the analysis, it reinforces it. RFC 7541 Appendix B assigns a code to every octet 0-255, so `0x0D` / `0x0A` are emitted as Huffman symbols and never appear as literal CR/LF bytes on the wire; the prefix is the Huffman byte count. The non-Huffman branch (`huffLen >= len(s)`, which a long random or binary path hits) is equally safe for the same length-prefix reason.

The only consumer of `H2HeadersEvent.Path` is `appendRequestPseudoHeaders` (`internal/layer/http2/channel.go:693-711`), which places it in the **value** position of `hpack.HeaderField{Name: ":path"}`. The `evt.RawQuery` rejoin at `channel.go:706-708` feeds that same value.

Oversized-entry handling in the dynamic table is also correct (`internal/layer/http2/hpack/dynamic_table.go:40-47`): an entry larger than `maxSize` evicts everything and is not stored, per RFC 7541 §4.4. No panic, no unbounded growth.

### 2. The line-delimited path (`serializeRequestLine`) — not reachable, for three independent reasons

`serializeRequestLine` (`internal/layer/http1/serialize.go:12-18`) does indeed perform zero validation, so this was the finding worth chasing. A `GRPCStartMessage.Path` set by this PR cannot reach it:

**(a) Type-level.** `grpcChannel.sendStart` writes an `*http2.H2HeadersEvent` to `c.inner.Send` (`internal/layer/grpc/channel.go:928-945`). The `http1` channel rejects anything that is not an `*envelope.HTTPMessage` outright:

```go
// internal/layer/http1/channel.go:616-619
msg, ok := env.Message.(*envelope.HTTPMessage)
if !ok {
    return fmt.Errorf("http1: Send requires *HTTPMessage, got %T", env.Message)
}
```

So even a hypothetical mis-wiring yields an error, never a serialized request line.

**(b) Wiring-level.** `grpclayer.Wrap` is only ever called with an HTTP/2 stream channel — `internal/connector/h2_dispatch.go:150` (client side) and `:214` (upstream side). The HTTP/1.x sibling `WrapH1UpstreamForDispatch` (`internal/connector/h1_dispatch.go:125-131`) wraps **only** for `envelope.ProtocolGRPCWeb`; there is no `grpclayer.Wrap` on any h1 channel anywhere in the tree.

**(c) The gRPC-Web bridge — the one place a `GRPCStartMessage` *does* become an `HTTPMessage` — ignores `Path` entirely.** `buildHTTPMessage` (`internal/layer/grpcweb/channel.go:1060`) does `msg.Path = "/" + start.Service + "/" + start.Method`, and only picks up `RawQuery` from a `srcEnv` whose `Message` is an `*HTTPMessage` (`:1067-1072`) — which the Start envelope never is. So even if a native-gRPC Start were somehow routed into the gRPC-Web Send path, `Path` / `RawQuery` are dropped, not serialized. This is also why USK-1069 is correctly scoped as a separate follow-up rather than a blocker.

I verified the ALPN argument independently and it holds, though it is not load-bearing given (a)-(c): `mitmAdvertiseFromUpstreamPick` (`internal/connector/alpn_routing.go:87-92`) narrows the client-facing advertisement to a single element equal to upstream's pick; the legacy fallback paths reconcile via `clientALPNMatchesUpstream` (`:195`) and `canonicalRedialALPNOffer` (`:211`) so the inner stack stays single-protocol end to end; and `tcp_forward`'s http1 arm asserts `*http1.Layer` on **both** `ClientTopmost()` and `UpstreamTopmost()` (`internal/proxybuild/tcp_forward.go:596-602`), routing `"http2"` / `"grpc"` to the h2 handler instead (`:314-321`). h2c (`internal/connector/h2c_handler.go:71-73`) builds an `http2.New` client Layer, and the CONNECT tunnel / h2 pool are transports beneath an unchanged L7 pairing.

**No h2-in / h1-out path exists. No HIGH finding here.**

### 3. Does it widen anything? Only the *shape* subset, never the *character* class

Confirmed by reading `parseGRPCPath` / `buildGRPCPath` (`internal/layer/grpc/channel.go:1146-1166`). `parseGRPCPath("/a\r\nX: y/c")` splits at the **last** slash → `Service="a\r\nX: y"`, `Method="c"`, and `buildGRPCPath` rebuilds `"/a\r\nX: y/c"` byte-identically. Every byte value this PR can now forward was already forwardable pre-PR through the Service/Method branch.

The newly reachable set is exactly the `:path` values `P` (post-`splitPath`) where `P != ""` **and** `parseGRPCPath(P) == ("", "")`:

- no leading `/`;
- no inner `/` after the leading one;
- leading `//` with a single remaining segment (`idx == 0`);
- trailing `/` (`idx == len(rest)-1`).

These previously emitted `/`. So this is a **shape** widening within an already-reachable character class, and the sink (HPACK value position) is length-prefixed. Correct call.

One genuinely new data channel, distinct from the shape question and worth stating plainly: `sendStart` now sets `RawQuery: m.RawQuery` (`channel.go:933`), so the query component of a gRPC `:path` reaches the upstream **at all**, where it was previously stripped for every gRPC request regardless of path shape. That is the intended wire-fidelity fix, and the same bytes were already reachable by stuffing them into the method segment — but it is the input to finding S-1 below.

I also swept for synthetic producers that *do* populate `Path` and could therefore have been flipped onto the new branch. Exactly one reaches a Layer `Send`: `buildReflectionStartEnvelope` (`internal/encoding/protoschema/reflection.go:634`) sets `Path: "/" + service + "/" + ReflectionMethod`, where both components are compile-time constants (`reflection.go:52,56,60`). That is parseable, so `pathForStart` takes the rebuild branch and emits a byte-identical string — no behavioural change. The other `Path`-setting producer, `internal/mcp/grpc_schema_discover.go:271`, feeds only `pipeline.NewBudgetStep` for the budget gate and never reaches `sendStart`. `buildResendGRPCStartEnvelope` (`internal/mcp/resend_grpc_helpers.go:941-949`), which `fuzz_grpc` shares via `cloneFuzzGRPCPlan`, leaves `Path` unset and so falls through to the rebuild exactly as before.

### 4. `RawQuery` → `Flow.URL`

**(a) Nothing downstream assumes it is empty.**

- `internal/mcp/resend_grpc_helpers.go:433` splits **only** `f.URL.Path`: `strings.SplitN(strings.TrimPrefix(f.URL.Path, "/"), "/", 2)`. `RawQuery` is a separate `url.URL` component and cannot contaminate it. An unparseable path degrades gracefully into the explicit `"service and method are required"` error at `:405`.
- `internal/job/grpc_source.go` consumes a `*GRPCStartMessage` directly and never reads a URL.
- `internal/flow/sqlite.go:769-771`, `har.go:352-356`, and `export.go:264-265` all go through `f.URL.String()`, which HTTP flows have been feeding attacker-controlled `RawQuery` since `record_step.go:1118`. Read-back (`sqlite.go:1313-1317`) is fail-soft: a `url.Parse` error leaves `f.URL == nil` rather than erroring the row.
  - Side note, not filed: Go's `net/url` rejects ASCII control bytes, so a raw CR/LF in the query round-trips to `f.URL == nil` and `resend_grpc { flow_id }` then reports `"authority is empty"`. `Envelope.Raw` still holds the truth, and this is byte-identical to the pre-existing HTTP behaviour.
- The per-message MCP view `grpcStartEntryView` (`internal/mcp/query_tool.go:2609-2616`) exposes neither `Path` nor `RawQuery`, so that surface is unchanged.
- `internal/flow/record_scope.go:189-200` deliberately derives the gRPC scope path from `Service`+`Method` (guarded on both being non-empty), *not* from `m.Path`, so capture-scope matching is unaffected by the unparseable case. `HTTPScopeStep` only handles `*HTTPMessage` (`internal/pipeline/http_scope_step.go:36-41`) and was never path-scoping gRPC.

**(b) No log injection (CWE-117).** No log line anywhere interpolates `RawQuery`. The only path-bearing log is the pre-existing `slog.Warn("grpc: malformed :path…", "path", evt.Path)` at `internal/layer/grpc/channel.go:503`, and both configured handlers (`internal/logging/logging.go:57-59`, `slog.NewJSONHandler` / `slog.NewTextHandler`) quote or escape control characters in attribute values. Not a finding.

Unrelated observation, pre-existing and outside this diff: `Warn` on a client-triggerable malformed request is louder than CLAUDE.md's own table suggests ("client sends invalid request" is listed as `Debug`), making it a log-volume amplifier reachable from the wire. Noted only so it is on the record.

**(c) Credentials — see S-2.**

### 5. Plugin surface — correctly closed

`buildGRPCStartDict` (`internal/pluginv2/convert.go:372-427`) registers dict keys for `Service`, `Method`, `Metadata`, `Timeout`, `ContentType`, `Encoding`, `AcceptEncoding`, and a read-only `Anomalies`. `Authority` / `Scheme` / `Path` / `RawQuery` are carried across the rebuild by closure capture only — they are never `setKV`'d, so `SetKey` (`internal/pluginv2/raw_field.go:250`) rejects them as unknown keys. The guard test now covers `raw_query` alongside `authority` / `scheme` / `path` (`internal/pluginv2/convert_test.go:291`).

No other route exists: `applyGRPCStartModify` (`internal/mcp/intercept_typed.go:418-442`) touches only Service / Method / Encoding / Metadata; `applyStartAction` (`internal/rules/grpc/transform.go:267-299`) only Metadata; and `Envelope.Raw` is informational on Start (the Layer always rebuilds — `internal/layer/grpc/channel.go:872-875`), so `msg["raw"]` mutation cannot reach `:path` either.

For contrast, `HTTPMessage` *does* expose `path` and `raw_query` as mutable plugin keys (`convert_test.go:388-403`) and they *do* reach `serializeRequestLine` unvalidated (`internal/layer/http1/channel.go:1301-1302`). That asymmetry is deliberate and pre-existing; this PR does not touch it.

Corollary worth noting for `grpcStartModified` (`internal/pipeline/record_step.go:1675-1687`): it does not compare `Path` / `RawQuery`, so a Step that mutated either would produce no modified variant. Since no Step *can* mutate them today, this is currently sound — but it becomes a variant-recording gap the moment either field is made mutable. Worth a comment there if that ever changes.

### 6. DoS / resource — bounded, and effectively free

`:path` is bounded on receipt at three levels before it ever becomes `Path` / `RawQuery`:

- `defaultMaxStringLength = 16 KiB` per decoded string literal (`internal/layer/http2/hpack/decoder.go:17`, enforced at `:241`);
- `defaultMaxHeaderListSize = 64 KiB` per decoded header list (`decoder.go:11`, enforced at `:94`);
- `maxHeaderFragmentBytes = 1 MiB` and `maxContinuationFrames = 32` on the raw HPACK accumulator (`internal/layer/http2/assembler.go:16,27`) — the CONTINUATION-flood guard.

`splitPath` (`internal/layer/http2/assembler.go:274-279`) returns two substrings of the *same* backing array, so retaining `RawQuery` adds no allocation beyond the `Path` string that `H2HeadersEvent` and (since USK-920) `GRPCStartMessage` already held. One Start message per stream. No new growth.

### 7. Panic safety — the guard dominates

`pathForStart` (`internal/layer/grpc/channel.go:1514-1521`) has no nil guard, and `sendStart` is its sole call site (grep across `internal/` returns exactly one). The typed-nil rejection at `channel.go:911-913` precedes both `buildStartHeaderKVs` (`:914`) and `pathForStart` (`:922`), and `Send`'s type switch (`:880-892`) is the only way to reach `sendStart`. A typed-nil `(*GRPCStartMessage)(nil)` carried in a non-nil interface therefore returns an error, not a nil-deref. `TestChannel_SendTypedNilStartMessageReturnsError` pins it. The omitted guard is correct, not a gap.

---

## FINDINGS

### S-1 · LOW · InputValidation · CWE-693 (Protection Mechanism Failure)

**File**: `internal/rules/grpc/safety.go:169-179` (with `internal/layer/grpc/channel.go:922,933`)

The gRPC Input SafetyFilter's `extractTarget` exposes only `metadata`, `service`, and `method` for a `*GRPCStartMessage`. The shipped presets declare `Targets: []Target{TargetBody, TargetURL, TargetQuery}` (`internal/rules/common/preset.go:27,33,39,…,90`), and `LoadPreset` maps only `TargetBody` → `TargetPayload` (`safety.go:62-74`) — `url` and `query` have no gRPC mapping and silently match nothing on a Start envelope.

Before this PR that was inert: an unparseable `:path` was normalised to `/` by `buildGRPCPath`, and the query component was dropped outright by `sendStart`, so neither ever reached the upstream. After this PR both are forwarded verbatim, opening a Send-direction channel to the upstream server that the Input Filter does not inspect.

**Impact**: guardrail completeness, not a boundary bypass. An agent driving the proxy can place a destructive payload in a gRPC `:path` or query and `destructive-sql` / `destructive-os-command` will not block it. Real-world exploitability is low — a gRPC `:path` routes to an RPC handler and is rarely an injection sink — which is why this is LOW and not MEDIUM.

**Remediation**: add `TargetURL` / `TargetQuery` cases to `extractTarget` for `*GRPCStartMessage`, returning `m.Path` and `m.RawQuery`. This is **scan-only** and does not conflict with the pre-settled "no `:path` validation on the relay path" decision — the SafetyEngine's own contract already forbids mutation (`safety.go:26-32`), so this adds detection without normalising anything. A follow-up Issue is equally acceptable; it need not block this PR.

### S-2 · LOW · MCP/API · CWE-200 (Exposure of Sensitive Information to an Unauthorized Actor)

**File**: `internal/pipeline/record_step.go:1312-1318`

`projectGRPCStart` now populates `Flow.URL.RawQuery`, so a query-bearing gRPC RPC target is recorded where it previously was not. The Output SafetyFilter cannot mask it: `filterOutput*` (`internal/mcp/safety_helper.go:63-104`) covers bodies, headers, metadata, and raw bytes only — **no caller masks `Flow.URL`**. Every consumer emits it raw via `msg.URL.String()` (`internal/mcp/query_tool.go:867`, `:1511`, `:1725`, `:1884`), as do the HAR (`internal/flow/har.go:355`) and JSONL (`internal/flow/export.go:265`) exports. Note that `url.URL.String()` percent-encodes `Path` via `EscapedPath()` but writes `"?" + RawQuery` **verbatim**.

**Impact**: a credential carried in a gRPC query string (`?access_token=…`) is returned to the AI agent and written to export artefacts unmasked.

**Remediation**: the underlying gap is pre-existing and symmetric — HTTP flows have populated `Flow.URL.RawQuery` since `record_step.go:1118` and are equally unmasked — so accepting this as consistent-with-HTTP is a defensible disposition. If you want it closed, the fix is protocol-agnostic (mask `Flow.URL` at the MCP boundary for all protocols) and belongs in its own Issue, not here.

---

## STATS

| Severity | Count |
|---|---|
| CRITICAL | 0 |
| HIGH | 0 |
| MEDIUM | 0 |
| LOW | 2 |

Neither LOW is exploitable in the proxy threat model — not triggerable as a boundary bypass from attacker-controlled traffic, does not touch CA keys, does not cause DoS — so per the verdict rules this is **APPROVED**. Both findings are reasonable as follow-up Issues rather than changes to this PR.

Also checked and clean: no new dependencies; no `unsafe`; no new goroutines, channels, or `defer`-in-loop; no TLS or certificate surface touched; `CloneMessage` correctly deep-copies the new field (`internal/envelope/grpc.go:112`).

**Test note**: every package this PR touches passes `make test` in my worktree — `internal/envelope`, `internal/layer/grpc`, `internal/pipeline`, `internal/pluginv2`. Three untouched packages (`internal/flow`, `internal/layer/grpcweb`, `internal/mcp`) hit the 4-minute `-timeout` panic on long-running cases (`TestLargeExportImport`, `TestGzipDecompressionCap`); those are load artefacts from concurrent agent worktrees on this machine, not assertion failures, and the code review reports the same commit green.

PR #127, opened 2026-09-15
ci: run the full e2e tier on dependency-update PRs

usk6666:
## Summary

The weekly dependency PR was gated on the **smoke** tier only, so a dependency could break the exhaustive tier and still merge green. Not hypothetical — this is the USK-1051 story, and I reproduced the timeline from the run history:

| when | what |
|---|---|
| 2026-09-14 09:04 UTC | PR #115 merges as `66886618` — `grpc 1.83.1 → 1.83.2` plus `x/net`, `x/sync`, `x/crypto`, `x/sys`, `x/text`, all pulled along by one `go mod tidy`. Smoke tier green. |
| 2026-09-14 17:00 UTC | nightly run `34872096294`, head_sha `66886618` → **21 `--- FAIL:` lines, all in `internal/mcp`** |

All 21 live in files tagged `//go:build e2e && !e2e_smoke` (`fuzz_grpc_integration_test.go`, `fuzz_grpc_macro_integration_test.go`, …) — precisely the tier the merge gate skips. Dependency PRs exist to prove behavioural compatibility, and a dependency can only break what its callers exercise; most of those callers are in the exhaustive tier.

This adds a `test-e2e-full` job to `ci.yml` that runs `make test-e2e` on dependency-update PRs, plus a post-merge backstop on `nightly-e2e.yml`.

## Why `head_ref` and not a label

`dependabot-security.yml` applies **no labels at all**. Confirmed against every merged dep PR — #83, #86, #92, #94, #95, #108, #115 — each one `labels: []`, each on `automated/security-updates`, each authored by `app/depsec-auto-vulns-fix-ci-app[bot]`.

Adding a label would also mean adding `types: [opened, synchronize, reopened, labeled]` to `ci.yml`'s `pull_request:` trigger, changing when **all six existing jobs** fire in order to condition one. The branch name is a fixed literal (`dependabot-security.yml`, force-reset to the base on every run), so it is the stable marker. A comment at that line now points back at this job, so a rename cannot silently drop the dep PR to smoke-only.

## Why a job-level `if:` and not a separate workflow

This is the one way the change could brick the repo, so spelling it out. GitHub documents both halves:

- A workflow **skipped by path/branch filtering** leaves its checks in `Pending`, and *"associated checks stay in a 'Pending' state and block merging."* A separate `test-e2e-full.yml` with `branches: [automated/security-updates]` would therefore sit pending forever on every ordinary PR — and if it were ever made required, block all of them.
- A **job skipped by a conditional** *"will report its status as 'Success'"* and *"will not prevent a pull request from merging, even if it is a required check."*

So the job lives inside `ci.yml`, which triggers on every PR. The workflow always runs; the job materialises a check run either way. That is what makes it safe to promote to required later. (I did not use a reusable-workflow `uses:` job either — whether the nested context materialises when the caller skips it is not something this repo's history can answer, and guessing wrong is unrecoverable.)

## Details

- **`timeout-minutes: 45`, not 15.** `make test-e2e` passes `-timeout 30m` to `go test`, and this repo deliberately lets the *inner* timeout fire first so a hang reports `test timed out` with a goroutine dump rather than an opaque job kill — the rationale is already written down for `lint` at `ci.yml:41-43`. Measured runtime (6–11 min) is beside the point; the number only matters when something hangs, which is exactly when the dump is wanted. No collision with USK-1048, which is about `-timeout 4m` per package under local contention inside one VM.
- **`Install protoc`**, carried over from `nightly-e2e.yml`. Four tests in `internal/mcp/grpc_schema_file_source_integration_test.go` `t.Skip` when `exec.LookPath("protoc")` fails. Without the step this job would cover strictly *less* than the nightly — the same silent under-coverage that USK-1058 just fixed.
- **`make test-e2e`, not a hand-copied `go test` line.** The Makefile is the only place `GODEBUG=http2xconnect=1` is recorded; duplicating the command is what produced USK-1058's 4-month gap. `ensure-ui` no-ops because the `ui-dist` artifact already populated `dist/` — same pattern as `lint`.
- **`test-e2e-smoke` is untouched** and still runs on these PRs. It is a *required* check, so making it conditional would add a failure point for zero gain; the two jobs get separate runner VMs, and smoke still reports first.
- **No data-path-only conditioning.** `tools/depsec` writes `plan.json` to `$RUNNER_TEMP` so it is never committed, and it would not discriminate anyway: PR #115's `go.mod` diff moved six modules at once. Essentially every weekly PR touches a data-path module.

## Post-merge backstop (`nightly-e2e.yml`)

`head_ref` structurally cannot see three cases: a human-opened bump on another branch, a `go.mod` change riding inside a feature PR, and — because ruleset 16469182 sets `strict_required_status_checks_policy: false` — a PR that merges green onto a base it never tested. Adding `push: branches: [main], paths: [go.mod, go.sum]` turns "up to 24 h" into "~10 min" for all three. A `paths:` filter is safe *there* precisely because that workflow is not a required status check.

## Cost

Public repo → free Actions minutes. Dep-PR wall clock ~6 min → ~11 min (the full tier becomes the critical path, running in parallel with the others). Every non-dep PR is unchanged — a skipped job is ≈ 0 s. Cadence ≈ 0.7 PRs/week.

## Follow-up action for the maintainer (deliberately not in this PR)

**The job lands advisory.** Making it a required check is an out-of-band GitHub settings change, sequenced after observation:

1. ~~Watch one ordinary PR — `test-e2e-full` should report **Skipped** and must not block the merge.~~ **Already demonstrated by this PR** — see the Test plan: the check reports `SKIPPED` and `mergeStateStatus` is `CLEAN`.
2. Watch one real dep PR — it should actually run the full tier. First opportunity is the next Monday `17 6 * * 1` run.
3. Then add `test-e2e-full` to ruleset **16469182**'s `required_status_checks` contexts.

Until step 3, a dep PR can still merge with this job red. That is intentional: the required-check flip is the irreversible half, and it should follow evidence rather than precede it.

## Test plan

GitHub Actions cannot be run from here, so this is what I could verify locally:

- [x] Both workflow files parse as valid YAML; the job graph resolves (`needs: setup` → an existing job), `timeout-minutes: 45`, last step is exactly `run: make test-e2e`.
- [x] Job order in `ci.yml`: `setup, lint, build, test, test-e2e-smoke, test-e2e-full, depsec-tool`.
- [x] `ci.yml` diff is **+52 / −0** — `test-e2e-smoke` provably untouched.
- [x] Every `uses:` SHA copied verbatim from an existing pin in `ci.yml`, version comment included; no unpinned `uses:` remains in the file.
- [x] Branch literal cross-checked against `dependabot-security.yml` and against all 7 merged dep PRs (`headRefName`, `labels: []`).
- [x] The skipped-job behaviour confirmed against GitHub's own documentation rather than assumed — quoted above.
- [x] **And then confirmed empirically on this very PR**, which is the negative control (`head_ref` is `ci/USK-1055-...`, so the `if:` is false):

  ```
  $ gh api .../actions/runs/34937105085/jobs
  setup            completed  success
  lint             completed  success
  build            completed  success
  test             completed  success
  test-e2e-smoke   completed  success
  depsec-tool      completed  success
  test-e2e-full    completed  skipped   <-- materialises, does not hang

  $ gh pr checks 127
  test-e2e-full    SKIPPED    skipping

  $ gh pr view 127 --json mergeable,mergeStateStatus
  {"mergeStateStatus": "CLEAN", "mergeable": "MERGEABLE"}
  ```

  So the job does surface as a PR check run when skipped, and does not block the merge. The trap this design avoids is real, and this design avoids it.
- [x] `make lint` (0 issues), `make build`, `make test` green. (One `make test` run went red under contention with a bare `FAIL` and **zero `--- FAIL:` lines** — the known USK-1048 load artefact. Re-run in isolation: exit 0, 0 failures.)
- [ ] **Not verifiable until a dep PR exists**: that the `if:` evaluates *true* on a real depsec PR — the positive control. First observable on the next Monday run, which is step 2 below.

Resolves USK-1055
Linear: https://linear.app/usk6666/issue/USK-1055


usk6666 (commented):
## Code Review: APPROVED ✅

CI-only change, and the risky part (a conditional job that must not wedge the merge queue) is implemented the one way that is safe — and it is backed by empirical evidence on this very PR: `test-e2e-full` materialises as a check reporting `skipping` and `mergeStateStatus` is `CLEAN`. Every factual claim in the new comments checks out against the repo, the run history, and the branch ruleset; the only findings are three LOW documentation/health nits and one out-of-scope observation.

### Verified

| Item | Result |
|---|---|
| SHA pins | All 3 `uses:` in the new job are byte-identical to pins already in `ci.yml` (checkout 6→7, setup-go 5→6, download-artifact 4→5 occurrences), version comment included. A sweep for any `uses:` without a 40-hex SHA across `.github/workflows/` returns empty: no unpinned action anywhere. |
| `test-e2e-smoke` untouched | `ci.yml` is **+52 / −0**; zero lines starting with `-` in the `ci.yml` hunk. |
| Artifact reuse | `needs: setup` resolves; `setup` uploads `ui-dist` from `internal/mcp/webui/dist/`; the new `download-artifact` uses the same `name:`/`path:` as `lint`/`build`/`test`/`test-e2e-smoke`. Job key at 2 spaces, steps at 4 — identical to siblings. Job order: `setup, lint, build, test, test-e2e-smoke, test-e2e-full, depsec-tool`. |
| YAML | All five workflow files parse cleanly. `nightly-e2e.yml`'s `on:` resolves to `{schedule, workflow_dispatch: None, push: {branches:[main], paths:[go.mod, go.sum]}}` — the interleaved comment is harmless, `workflow_dispatch` survives, the cron is unchanged. |
| Comment accuracy | PR #115: `headRefName: automated/security-updates`, `labels: []`, merged `2026-09-14T09:04:33Z` as `66886618`; its `go.mod` moved `grpc 1.83.1→1.83.2` plus exactly five `x/` modules. Nightly run `34872096294` (`event: schedule`, `headSha 66886618`, 17:00Z ≈ +7.9 h) → exactly **21** `--- FAIL:` lines, all in `internal/mcp/{fuzz_grpc,fuzz_grpc_macro,resend_grpc}_integration_test.go`, each tagged `//go:build e2e && !e2e_smoke`. `grpc_schema_file_source_integration_test.go` gates on `exec.LookPath("protoc")` in exactly **four** tests. PR #120 is the USK-1058 `make test-e2e` change. Ruleset 16469182 really does set `strict_required_status_checks_policy: false`, and its required contexts are `lint, setup, build, test, test-e2e-smoke, depsec-tool` — `test-e2e-full` is correctly *not* among them. `ci.yml:41-43` is indeed the lint-timeout rationale. |
| `if:` semantics | `github.head_ref` is empty on `push`, so the `event_name` guard is redundant-but-defensively-correct (it also keeps the job inert if `merge_group` / `pull_request_target` is ever added). `synchronize` re-fires it, which matters because the depsec workflow force-updates the branch in place on a re-run. |

### Findings (LOW / NIT only)

**F-1 · LOW · Documentation · `.github/workflows/ci.yml:154`**

`Keep this literal in sync with .github/workflows/dependabot-security.yml:169.` — the reference is stale on arrival. Pre-PR, `branch="automated/security-updates"` *was* at `:169`; this PR's own 3-line comment insertion pushed it to `:172`, so `:169` now lands on the first line of the new cross-reference comment. Harmless (the right line is 3 below) but factually off.

*Suggestion*: drop the line number — "…in sync with `.github/workflows/dependabot-security.yml` (the `branch=` assignment)" — so it cannot drift again.

**F-2 · LOW · Documentation · `CLAUDE.md:188`**

Describes the new job as `ci.yml`'s `test-e2e-full` **merge-gate** job. It is deliberately *advisory* until the maintainer edits ruleset 16469182 (confirmed: the required contexts do not include it), and the PR body is explicit that "a dep PR can still merge with this job red." `merge-gate` is a loaded term in this exact block: the line above calls smoke the "merge-gate subset", and `Makefile:58` says the full tier is "not part of the per-PR gate".

*Suggestion*: "…as `ci.yml`'s `test-e2e-full` job (advisory until it is added to the branch ruleset's required checks)…"

**F-3 · LOW · CodeHealth · `ci.yml:165` / `dependabot-security.yml:172` / `CLAUDE.md:188`**

The branch literal is now duplicated in **three** places (CLAUDE.md too, not just the two workflows), bound only by prose. Rename → the gate silently stops firing and nothing goes red; that failure mode re-creates exactly the bug USK-1055 fixes. I **agree** it is out of scope for a 4-file CI PR — the bidirectional comments are a reasonable compensating control, and a grep would only prove the literals match, not that the gate fires (the positive control is unverifiable until the next Monday run either way). If you want it mechanised later, the cheapest home is a second step in the existing `depsec-tool` job (runs on every PR, ~5 s):

```yaml
      - name: Branch literal stays in sync with ci.yml's test-e2e-full gate
        working-directory: ${{ github.workspace }}
        run: |
          set -euo pipefail
          b='automated/security-updates'
          grep -qF "branch=\"$b\"" .github/workflows/dependabot-security.yml
          grep -qF "github.head_ref == '$b'" .github/workflows/ci.yml
```

A follow-up Issue is fine; not a blocker for this PR.

**F-4 · NIT · Observation · `ci.yml:143` vs `Makefile:54`** — no action requested in this PR.

Correcting the design survey: `ci.yml:121` (`go test -race -v -timeout 4m ./...`) **is** byte-identical to `Makefile:41`. `ci.yml:143` is **not** byte-identical to `Makefile:54` — flag order differs (`-tags 'e2e e2e_smoke' -timeout 10m` vs `-timeout 10m -tags 'e2e e2e_smoke'`) — though semantically identical. Converting the older jobs to `make` targets is explicitly out of scope here; noted only so the record is right.

### On `nightly-e2e.yml` — in scope, and no `concurrency:` needed

The backstop covers three cases `head_ref` structurally cannot see, and it is cheap; I read it as in scope rather than creep. On runs piling up: **no change recommended.** `scorecard.yml` is the exact precedent — `schedule` + `push: branches:[main]`, no `concurrency:` — and the only workflow in the repo that uses one is `dependabot-security.yml` (`cancel-in-progress: false`, for a workflow that mutates a branch ref). Two 10-minute read-only runs overlapping on free public runners costs nothing, and the collision window is a merge landing within ~10 min of 17:00 UTC. If one is ever added it **must** be `cancel-in-progress: false`; `true` would let a merge push cancel the nightly and silently drop the very signal this PR exists to protect.

Two observations, neither a finding: `paths: [go.mod, go.sum]` matches root-level files only (GitHub path patterns without `**` do not descend), so `tools/depsec/go.mod` will not trigger it — correct, since that module is stdlib-only and not shipped. And an npm-only depsec week (`web/pnpm-lock.yaml` only) will not trip the push backstop — that case is covered by the `head_ref` half on the PR itself.

---

usk6666 (commented):

**This PR contains no Go code** — it is GitHub Actions YAML plus one Markdown line. The usual TLS/cert, parser, wire-fidelity, and Go memory-safety sections of the MITM checklist are inapplicable. This review is scoped to **CI supply-chain and CI-permissions** territory, which is where the entire risk surface of this change lives.

Verdict: no CRITICAL/HIGH/MEDIUM findings. One LOW documentation nit.

### 1. Script injection (CWE-94/CWE-78) — the check that mattered

`github.head_ref` is attacker-influenceable: this is a **public** repo (`visibility: PUBLIC`), so anyone can open a fork PR with an arbitrary branch name containing shell metacharacters. The classic sink is interpolation into a `run:` block, where `${{ }}` expands before the shell parses the text.

```
$ grep -rn -E 'head_ref|pull_request\.head|event\.issue|event\.comment|event\.review|github\.actor|event\.pull_request' .github/
.github/workflows/ci.yml:165:    if: github.event_name == 'pull_request' && github.head_ref == 'automated/security-updates'
```

**Exactly one occurrence in the whole tree, and it is an `if:` expression** — an expression-context string comparison evaluated by the Actions expression engine, which never reaches a shell. A structured YAML walk over every job and step confirms the stronger property:

> **Zero `run:` blocks in any of the five workflows contain `${{` at all.** Every interpolation in the repo sits in an `env:` mapping (`release.yml`, `dependabot-security.yml`) — the pattern GitHub explicitly recommends.

Independently corroborated: CodeQL `Analyze (actions)` — whose query pack includes `actions/code-injection` — **passes** on this head (run `34937103229`).

### 2. Token permissions

- `ci.yml` workflow-level `permissions: contents: read` (lines 9-10). ✅
- The new `test-e2e-full` job adds **no** job-level `permissions:` block, so it inherits the read-only token. ✅ (`ci.yml` has no job-level permissions block anywhere.)
- Trigger is the existing `pull_request`, **not** `pull_request_target` — a grep for `pull_request_target` across `.github/` returns nothing. ✅
- The new job touches no `secrets.*`, no `environment:`, no `container:`, no `env:`. ✅
- `nightly-e2e.yml` keeps `permissions: contents: read`; the added `push` trigger widens nothing, and the `e2e-full` job uses no secrets.

### 3. Action pinning

Every added `uses:` is SHA-pinned **and byte-identical to a pin already present in `ci.yml`** — this PR introduces no new action and no new version into the trust boundary:

| Added | SHA | Pre-existing in ci.yml |
|---|---|---|
| `actions/checkout` | `de0fac2e…ce83dd` # v6.0.2 | ✅ (now 7×) |
| `actions/setup-go` | `4b73464b…27bd417` # v6.3.0 | ✅ (now 6×) |
| `actions/download-artifact` | `3e5f45b2…5461e7c` # v8.0.1 | ✅ (now 5×) |

Filtering every `uses:` line in `.github/` for ones lacking a 40-hex SHA returns empty. No unpinned or tag-pinned action anywhere.

### 4. Fork-PR reasoning

A fork **can** name its branch `automated/security-updates` and trigger the job. Consequence, reasoned through rather than assumed:

- The token is `contents: read`; for a public repo, a fork-originated `pull_request` run receives **no repository secrets**.
- `download-artifact` without `run-id`/`github-token` reads only the current run's artifact, so the `ui-dist` a fork consumes is the one its own `setup` job just built — no cross-run poisoning vector.
- Fork code **already** executes on every PR today via `setup` (`make build-ui` runs the fork's package scripts) and via `test` / `test-e2e-smoke`. The full tier runs *more* of that same untrusted code at the *same* privilege — a quantitative delta (~15 min → ~45 min of runner time, hard-capped by `timeout-minutes: 45`), not a privilege delta.

Net: no escalation. The residual is bounded free-tier runner minutes, further gated by GitHub's default "require approval for first-time contributors". Not a finding.

### 5. Security-patching trade-off — PR body is honest

Verified against the live ruleset rather than taking the body's word:

- `strict_required_status_checks_policy: false` on ruleset 16469182 — **claim is true**.
- Required contexts are `[lint, setup, build, test, test-e2e-smoke, depsec-tool, Analyze (actions|go|javascript-typescript)]` — **`test-e2e-full` is absent**, so "the job lands advisory" is **true**. A full-tier flake therefore cannot block a security bump today.
- Admin bypass (`RepositoryRole 5`, `bypass_mode: always`) remains as the release valve.
- The body states plainly: *"Until step 3, a dep PR can still merge with this job red. That is intentional."* No claim is overstated.

One nuance the body leaves implicit: the three most recent commits on the default branch (`885258d3`, `865afbb9`, `566090fc`) are all full-tier e2e flake fixes, so the tier has a live flake history. That is an argument for keeping this advisory for a while before step 3 — which is exactly the sequencing the PR already proposes. No change requested.

### 6. `Install protoc`

Not new surface: the identical `sudo apt-get update && sudo apt-get install -y protobuf-compiler` step already exists at `nightly-e2e.yml:60`. The repo pins no apt packages anywhere — normal for `ubuntu-latest`, and at most informational rather than a finding.

---

### LOW findings

**S-1 — stale cross-reference line number** (`.github/workflows/ci.yml:154`, Category: SupplyChain, CWE-1059)

The comment reads *"Keep this literal in sync with `.github/workflows/dependabot-security.yml:169`"*, but the `branch="automated/security-updates"` literal now lives at **line 172** — the three comment lines this PR adds to `dependabot-security.yml` shifted it by exactly 3. The PR body repeats `:169` in two places.

The purpose of that comment is to stop the branch-name coupling from silently degrading the gate to smoke-only, so a pointer landing three lines off partially defeats it. No exploitability; purely a doc-accuracy nit. Optional fix: say `:172`, or drop the line number and reference the `branch=` assignment by name so it cannot drift again — a grep for the literal already finds both sites.

---

**STATS** — CRITICAL: 0 · HIGH: 0 · MEDIUM: 0 · LOW: 1

---

PR #130, opened 2026-09-15
fix(layer/grpc,layer/grpcweb): demote client-triggerable malformed :path logs to Debug

usk6666:
## Summary

- **Demote two client-triggerable `slog.Warn` calls to `slog.Debug`** — the malformed-`:path` log in `internal/layer/grpc/channel.go` and its sibling in `internal/layer/grpcweb/channel.go`. Both are driven entirely by untrusted input: any peer can send an unparseable `:path` at will, the log fires once per stream with no dedup or sampling, and HTTP/2 multiplexes many streams per connection — an unbounded log-amplification vector. CLAUDE.md's Log Level Guidelines put "client sends invalid request" at `Debug` ("client-side issue; not a proxy anomaly").
- **Retire the `D1 — tolerant` rationale in the comments.** `Warn` was justified when a malformed `:path` was *lost* by the Layer, making the log the operator's only view of it. USK-1053 removed that premise: the observed path now survives to the wire and into `Envelope.Raw`, so it stays recoverable from the flow record without the log line. The structured attributes (`stream_id`, `path`) and the outside-the-mutex placement are unchanged.
- **Record the override behaviour as spec in RFC-001 §11** — an intercept / plugin `service` / `method` override is *deliberately inert* on a flow whose observed `:path` is unparseable, because `pathForStart` keeps the wire overlay, which is the only faithful representation of that path (MITM principle #1). This is a consequence of the USK-1053 precedence rule, not a regression: before USK-1053 the override "worked" only by silently re-normalizing the malformed path. The bullet names the pinning test, records why clearing `Path` on an explicit `service` / `method` edit was rejected, and carries the re-open trigger for adding `path` / `raw_query` to `grpcStartMessageModify`.

This is the option (c) disposition for USK-1075: close as spec, plus the log-level demotion. Options (a) and (b) were not implemented.

## Scope notes

- **No behavioural change.** `applyGRPCStartModify`, `pathForStart`, and all precedence logic are untouched; the diff is 3 files / +15 / -5.
- **No schema changes** — no `path` parameter added anywhere.
- **No test changes.** No existing test asserts these log lines, and the behavioural pin `TestPathForStart/unparseable_path_wins_even_against_a_populated_derived_view` must stay as-is because option (c) ratifies it as spec.
- Zero line overlap with sibling Issues USK-1068 (CRLF guards in `applyGRPCStartModify`) and USK-1069 (gRPC-Web `:path` overlay) — only a `slog` line and a comment were touched in `grpcweb/channel.go`, no `:path` handling logic.

## Test plan

- [x] `make lint` — pass (gofmt clean, golangci-lint v2.13.2, **0 issues**)
- [x] `make build` — pass (`go vet ./...` clean, binary built)
- [x] `make test` — `internal/layer/grpc` and `internal/layer/grpcweb` both **ok** on a forced uncached run (`GOFLAGS=-count=1`); `TestPathForStart/unparseable_path_wins_even_against_a_populated_derived_view` PASS
- [x] No behavioural change: the `pathForStart` precedence rule and its pinning test are untouched

### Pre-existing failure, unrelated to this PR

`make test` fails locally in `internal/mcp` with `panic: test timed out after 4m0s` (the Makefile's explicit `-timeout 4m`). Verified as **pre-existing**: an identical run on unmodified `main` (`d532c99f`) produces the same single failure (`FAIL internal/mcp 240.286s`). The tests executing at the timeout were at 0s–2s, i.e. the package as a whole exceeds the wall-clock limit under `-race` on this machine rather than deadlocking. `internal/mcp` is not touched by this PR.

Resolves USK-1075
Linear: https://linear.app/usk6666/issue/USK-1075

https://claude.ai/code/session_018iPcj6sQatgB1A3rwRJjwg


usk6666 (commented):

A 15-line, log-level-and-comment-only diff with no data-path behavioural change. I verified against the code that the demotion removes **no** recoverable signal, and that it in fact **closes** a small default-level log-amplification surface on untrusted input. Zero security findings.

### Central question: does the demotion remove signal an operator needs?

**Conclusion: no.** The `Warn` was a *push* notification; the *pull* record is complete and byte-faithful, which is the channel this proxy is actually built around. Verified end to end:

| Claim | Evidence |
|---|---|
| Observed `:path` survives in the L7 message | `buildStartMessage` sets `msg.Path = evt.Path` verbatim on the Send side (`internal/layer/grpc/channel.go`); gRPC-Web's `applyRequestOverlay` sets `start.Path = msg.Path` (`internal/layer/grpcweb/channel.go:496-503`) |
| Observed `:path` survives in `Envelope.Raw` | `absorbHeaders` sets `Raw: cloneBytes(ev.Raw)` on the Start envelope; the H2 assembler's Raw-for-event contract populates `Raw = rawBlock` (the complete HPACK block as observed on the wire, `internal/layer/http2/assembler.go:88`, `:472`/`:506`) |
| It reaches the flow record | `projectGRPCStart` (`internal/pipeline/record_step.go:1283`) writes `m.Path` verbatim into `fl.URL.Path`; `projectRawBytes` (`:1207`) writes `env.Raw` into `Flow.RawBytes` |
| The malformed *condition* is itself queryable | `projectGRPCStart` always emits `grpc_service` / `grpc_method` — both empty is exactly the parse-failure signal, filterable via the `query` MCP tool |
| The precedence rule is pinned | `TestPathForStart/unparseable_path_wins_even_against_a_populated_derived_view` exists (`internal/layer/grpc/channel_test.go:1871`) |

This matches CLAUDE.md's own decision criterion verbatim — *"If an operator can confirm normal operation without the log, use Debug"* — and the `Debug` row for "client sends invalid request / client-side issue; not a proxy anomaly".

### The amplification precondition is real, and the demotion is a net security improvement

I confirmed the vector the PR cites rather than taking it on faith:

- The log fires **once per stream** in `absorbHeaders` with no dedup, sampling, or rate limit.
- `defaultMaxConcurrentStreams = 500` (`internal/layer/http2/connstate.go:25`) bounds *concurrency* but not cumulative streams over a connection lifetime — a client can recycle stream IDs indefinitely.
- The logged `path` attribute can be large: `defaultMaxHeaderListSize = 0` (unlimited decoded list) and `maxHeaderFragmentBytes = 1 << 20` caps the raw HPACK block at 1 MiB.

So prior to this PR, a peer could drive attacker-controlled bytes into the operator's **default-level** log at unbounded volume — a CWE-779 / CWE-400 class write-amplification and disk-exhaustion surface reachable from any client that can open an h2 stream. Moving it behind `-log-level debug` removes it from the default path. **The demotion is the security-positive direction here, not a trade-off.**

### Residual `Debug`-level exposure: acceptable

The two attributes are `stream_id` (int32) and the raw client-supplied `path`.

- **Log forging / CRLF injection (CWE-117) is mitigated at the handler.** `internal/logging/logging.go:52-62` uses stdlib `slog.NewJSONHandler` / `slog.NewTextHandler`; both quote or escape control characters in attribute values, so a `:path` carrying `\r\n` or ANSI escapes cannot fabricate log lines.
- **No credential exposure.** `evt.Path` is the path component only — the query string is split off into `RawQuery` at the H2 assembler and is not logged.
- The attribute set is unchanged by this PR in any case.

### Scope check

No TLS/certificate, CA-key, MCP-tool-input, dependency, or concurrency surface is touched. The mutex discipline in `absorbHeaders` (log emitted after `c.mu.Unlock()`) is preserved exactly.

### Observations (not findings)

1. `internal/layer/grpcweb/service_method.go:9` — the `parseServiceMethod` doc comment still reads *"Callers are expected to log a slog.Warn at the call site"*. Now stale for its sole caller. Cosmetic doc drift with no security impact; code-review territory.
2. `projectGRPCStart` only constructs `fl.URL` when `Authority != "" || Scheme != ""`, so a peer omitting both `:scheme` and `:authority` would leave `Flow.URL` nil and the path absent from that projection. `Envelope.Raw` still carries the HPACK bytes, so the anomaly stays recoverable. Pre-existing; not introduced or worsened here.
3. CRLF guards on `service` / `method` in `applyGRPCStartModify` are already tracked as **USK-1068**, and the gRPC-Web `:path` overlay as **USK-1069**. Noted only for completeness — correctly out of scope for this diff.

**FINDINGS: None** (CRITICAL 0 / HIGH 0 / MEDIUM 0 / LOW 0)

---

usk6666 (commented):
## Code Review: APPROVED ✅

The log-level demotion itself is correct and well-justified: `slog` stays used in both files, `gofmt` is clean, no test or doc asserts on these lines (and `grpc_integration_test.go:834` already installs a `LevelDebug` handler, so the demoted lines remain visible in test output). The RFC §11 bullet matches its neighbours' `- [ ] <topic> — **<disposition> <date> (<Issue>)** … **Re-open trigger:** …` shape, and its `USK-920` / `USK-1053` attributions, its `internal/mcp/intercept_typed.go` + `internal/pluginv2/convert.go` references, and the `TestPathForStart/...` pin all check out against source. Four documentation-accuracy issues, none behavioural — two of them are doc drift the PR's own change introduced.

### Findings

| F-1 | MEDIUM | internal/layer/grpc/doc.go | 52-53 | CodeHealth | Package godoc still documents the D1 contract as emitting a **Warn** log |
| F-2 | MEDIUM | docs/rfc/envelope.md | 1484 | Correctness | The bullet's "unreachable from `service`/`method`" characterization is wrong for three shapes |
| F-3 | LOW | internal/layer/grpcweb/service_method.go | 9 | CodeHealth | `parseServiceMethod` godoc still tells callers to log `slog.Warn` |
| F-4 | NIT | internal/layer/grpc/channel.go | 505-506 | Correctness | `Envelope.Raw` / flow-record survival is attributed to USK-1053, but both predate it |

### Suggestions

**F-1** — `internal/layer/grpc/doc.go` is the canonical statement of the D1 decision and still reads:

```
//     leading slash, no separator, single segment) yield Service="" and
//     Method="" together with a Warn log. The malformed path is surfaced
//     to Pipeline rather than hidden, for diagnostic purposes — the
//     Envelope.Raw still contains the original wire bytes.
```

This now contradicts the code, and specifically contradicts the new inline comment in `channel.go`, which argues the D1 "surface to operator" rationale has lapsed. Since this PR is the one that changes the level, updating D1's own documentation belongs here. Suggest `… together with a Debug log`, plus a short clause carrying the new rationale (client-controlled trigger, once per stream, no dedup) so the two comments agree.

**F-2** — Checked against `buildGRPCPath` / `parseGRPCPath` (`internal/layer/grpc/channel.go:1156-1176`). The three *reachable* examples are all correct: `buildGRPCPath("a/b","c")=="/a/b/c"`, `buildGRPCPath("/pkg","Do")=="//pkg/Do"`, `buildGRPCPath("x","")=="/x/"`. The generalization around them is not:

- **A path without a leading `/` is unreachable regardless of slash count.** `buildGRPCPath` unconditionally prefixes `/`, so `a/b/c` — two slashes — is unreachable and inert (`parseGRPCPath` rejects it for the missing leading slash). "Fewer than two slashes" excludes it.
- **`//` is unreachable despite having two slashes.** Producing it needs `"/"+s+"/"+m == "//"`, i.e. `s==""&&m==""`, which hits the `return "/"` early branch instead. `parseGRPCPath("//")` also fails (`idx==0`), so an observed `//` is inert *and* unrepresentable.
- **`/` is reachable despite having one slash**, via `service==""`/`method==""`. Benign — `parseGRPCPath("/")` fails too, so the overlay yields the same `/` — but the sentence lists it as unreachable.

Accurate form: *the unreachable set is the empty path, any path without a leading `/`, any single-slash path other than `/`, and `//`.* Worth fixing because this sentence is what bounds the deferred gap for the re-open trigger.

**F-3** — This file's godoc says "Callers are expected to log a `slog.Warn` at the call site rather than failing the stream." `refillFromHTTPMessage` (`channel.go:319`) is the sole caller and this PR changed it to `slog.Debug`. Suggest either naming `slog.Debug` or making it level-agnostic ("log at the call site rather than failing the stream") so it cannot drift again.

**F-4** — The new comment reads "The D1 'surface to operator' rationale lapsed at USK-1053: the observed `:path` now survives into `Envelope.Raw` and the flow record either way." The conclusion is sound but the causality is mis-dated: `Raw: cloneBytes(ev.Raw)` on the Start envelope dates to `fd33a3ca` (PR #635), and the `Path` → `Flow.URL` projection to USK-920 (`buildStartMessage` comment + `projectGRPCStart`). What USK-1053 actually added is that an unparseable `:path` survives into the **re-emitted upstream wire** (`pathForStart`). Suggest "… survives into `Envelope.Raw` and the flow record (USK-920) and, since USK-1053, into the re-emitted wire as well."

### Observations (not findings)

- `docs/rfc/envelope-ja.md` remains out of sync — its §11 still lacks the entire "Post-N9 deferred design decisions" subsection. Consistent with USK-1053's English-only precedent, so not raised as a finding, but the drift is now four bullets deep.
- `internal/layer/http2/reader.go:132` (`http2: peer sent GOAWAY`) is the only remaining peer-triggerable `slog.Warn` in the surrounding Layers. It is out of USK-1075's scope and arguably legitimate at Warn, noted only for completeness.

---

PR #131, opened 2026-09-16
fix(rules/grpc): scan :path and query in the Input SafetyFilter

usk6666:
## Summary

`internal/rules/grpc/safety.go`'s `extractTarget` had three targets for a `*envelope.GRPCStartMessage` — `metadata`, `service`, `method` — and nothing corresponding to url or query. All 11 shipped destructive preset rules declare `Targets{TargetBody, TargetURL, TargetQuery}`; on a Start, `TargetBody` yields nothing (it aliases to payload, which lives on Data messages) and url/query fell through to `("", "")`, which `checkRule` skips. **Every preset rule evaluated zero targets and could not fire on a gRPC Start.**

This was inert before #126 (USK-1053): a malformed `:path` was normalized to `/Service/Method` on Send and the query was dropped outright (`GRPCStartMessage` had no `RawQuery`). USK-1053 made both wire-faithful — that is the correct fix, but its side effect flipped this Input SafetyFilter hole from inert to live.

**Scan-only. No wire value is rewritten.**

- `extractTarget`'s `*GRPCStartMessage` arm gains inline `case common.TargetURL` / `case common.TargetQuery`. No new exported constants and **no `LoadPreset` change** — `LoadPreset` does no target remapping today (it is byte-identical to `http.LoadPreset` / `ws.LoadPreset`) and must not start. `"url"` and `"query"` are already in `config.validInputTargets`, so this makes an already-expressible operator config actually work; a new `"path"` spelling would have required editing `internal/safety/rule.go` — reaching into the Output Filter package to make an Input Filter feature expressible, against MITM Principle 7, for zero added expressiveness.
- New `rules/grpc.reconstructURL` mirrors the `rules/http` sibling byte-for-byte (`scheme://authority` + path + `?query`, `strings.Builder`, no `net/url`), so a preset authored against HTTP semantics means the same thing on gRPC. Duplicated rather than shared per MITM Principle 2 — it reads a different struct and carries the Service/Method fallback that `projectGRPCStart` already established for `Flow.URL` on a Path-less synthetic Start.
- The Data and End arms keep the existing fallthrough. Neither message type has `Path` / `RawQuery` / `Authority` / `Scheme`, and adding no-op cases would push `extractTarget` past the gocyclo ceiling.
- Three doc sites corrected: the package doc, `Violation.Target`'s enumerated list (which was already missing `metadata:<name>`), and `extractTarget`'s godoc. `LoadPreset`'s godoc also now says explicitly that it performs no remapping — its previous wording is what led the Issue to propose a `LoadPreset` mapping that does not and should not exist.
- `internal/safety/` and `internal/config/safety.go` are untouched.

## Severity honesty — why this is LOW, and why LOW is not "theoretical"

Three things bound the real exposure. None of them makes the bug unreal; all three are needed to read the rating correctly.

1. **The gRPC safety engine is nil in every default configuration.** It requires a config file with `safety_filter.enabled: true` **and** a non-nil `input:` block. The `-safety-filter` flag / `YP_SAFETY_FILTER_ENABLED=1` env var alone is **insufficient** — it flips `enabled` but leaves `sfCfg == nil`, so `InitPerProtocolSafetyEngines` still short-circuits and no engine is built.
2. **The AI-agent resend/fuzz surface was never affected.** It is covered by a different engine (`internal/safety.Engine`, via `checkSafetyInput` at `resend_grpc.go` / `fuzz_grpc_helpers.go`), which does implement url/query. The live exposure here is **client-originated gRPC / gRPC-web traffic transiting the proxy**.
3. **Percent-encoded payloads are still not caught**, matching HTTP. The query is matched raw and undecoded exactly as `rules/http` matches `HTTPMessage.RawQuery`; decoding would normalize what the wire did not normalize (MITM Principle 1). This fix buys exactly HTTP-parity coverage, no more — `?q=DROP%20TABLE%20users` goes through, on gRPC and on HTTP alike. A test pins that as deliberate so a future change has to move both protocols together.

## Incidental coverage and known residual

- **grpc-web gains the path scan for free.** `grpcweb`'s `applyRequestOverlay` emits the same `*envelope.GRPCStartMessage` with `Path` set, so its RPC target becomes scannable through the same arm. It never populates `RawQuery`, so the query target stays a no-op there. Not a scope expansion — no grpc-web code is touched.
- **USK-1078 (residual, recorded inline, not fixed here):** under the USK-702 `modify_and_forward` recheck, if an operator overrides Service/Method on a *parseable* original `:path`, `pathForStart` emits `/newSvc/newMethod` while `m.Path` still holds the observed pre-override path — so the url target scans the pre-override path. This is not a regression introduced here (before this PR the url target scanned nothing at all), and closing it needs `parseGRPCPath`, which is unexported in `internal/layer/grpc`.

## Test plan

- [x] `make lint` — 0 issues (gocyclo included; `extractTarget` stays under the 15 ceiling)
- [x] `make build`
- [x] `make test` — 4245 tests pass, exit 0
- [x] **Mutation-verified**: removing the two new `case` arms fails 7 of the new top-level tests (`TestExtractTarget_StartURLQuery`, `..._OnStartPath`, `..._OnStartQuery`, `TestSafetyEngine_TargetQuery_OnStart`, `TestSafetyEngine_TargetURL_OnStart`, `..._CheckInputAll_...`, `..._DoesNotMutateStartMessage_OnMatch`). The absence-asserting tests correctly stay green.
- [x] Bidirectional: destructive `:path` (`/;DROP TABLE users--`, wholly malformed) and destructive query (`?q=DROP TABLE users` on a well-formed path) both blocked; benign RPC targets across 5 shapes and both shipped presets are not.
- [x] Non-mutation contract repaired. The existing Start test snapshotted only `Metadata`, never `env.Raw`, and — unlike the Data path — had no `_OnMatch` variant at all, so the new arms would have shipped under a contract nothing enforced. The snapshot now covers `Scheme` / `Authority` / `Path` / `RawQuery` / `Service` / `Method` / `env.Raw` (including the backing arrays), and the missing `_OnMatch` variant is added.
- [x] MITM Principle 1 guard on the new builder: no lowercasing, default-port stripping, percent-decoding, or slash collapsing.

### On the optional e2e

The exhaustive-tier e2e was scoped as optional. I dropped it deliberately, because **the obvious oracle would have been a false positive.** `mcptest`'s `GRPCObservedHits` is recorded *inside* the registered `Echo` handler (`grpc_upstream.go`), so a malformed `:path` or a query-bearing `:path` never reaches the handler regardless of whether the SafetyFilter blocks it — grpc-go's server routes on `:path` and would answer `Unimplemented`. A hit-counter assertion would therefore pass identically with and without this fix. A sound e2e needs a differential oracle (status code with vs. without the filter, or a transport-level stream counter on the upstream), which means changing `internal/mcptest` infrastructure. Flagging rather than silently shipping a green-but-meaningless test.

Resolves USK-1073
Linear: https://linear.app/usk6666/issue/USK-1073

https://claude.ai/code/session_013ToA1skCGjJGjfhcbAiB6N


usk6666 (commented):

Scan-only change with a verified non-mutation contract: `reconstructURL` builds into a function-local `strings.Builder` that is never appended to after `String()`, `m.RawQuery` is returned as an immutable Go string, and `grep` confirms zero assignments to `Path` / `RawQuery` / `Scheme` / `Authority` / `Service` / `Method` anywhere in non-test `internal/rules/grpc/`. The brief's ReDoS/DoS premise does not hold in this codebase (details below) and no follow-up header-cap Issue is warranted. Three LOW observations, two of which deserve follow-up Issues rather than changes here.

### Verification notes

**1. Non-mutation contract — holds.**
`checkRule` does `rule.Pattern.Find([]byte(data))`; `[]byte(string)` allocates a fresh copy (the value is passed to a method that could retain it, so the compiler cannot elide it), and `string(match)` copies again — `Violation.Match` cannot alias message memory or `Envelope.Raw`. `reconstructURL`'s returned string aliases only a builder that has gone out of scope. The two new tests assert `Envelope.Raw` plus all seven read fields are byte-identical on **both** the match and no-match paths, which is a genuine widening: the Start path previously had neither variant.

**2. ReDoS / DoS — not a concern, on two independent counts.**

- `common.CompilePattern` (`internal/rules/common/compiler.go:17`) uses `regexp.Compile` — Go's RE2. RE2 does not backtrack, and `Find` is linear in input length regardless of pattern shape. The nested quantifier `(\s|/\*.*?\*/)+` in `sqlWS` (`preset.go:16`) would be catastrophic under a PCRE-style engine but is linear here; Go's small-input `backtrack` engine is a *memoized* backtracker with a visited bitmap, also linear. Catastrophic backtracking is architecturally impossible in this codebase.
- The `:path` is **not** bounded at ~1 MiB. `maxHeaderFragmentBytes = 1 << 20` (`assembler.go:20`) bounds the *raw HPACK bytes* across HEADERS+CONTINUATION, before decode. Decode is then bounded by `hpack.NewDecoder`'s own defaults, which the production path never overrides: `defaultMaxStringLength = 16 KB` per decoded string literal (`decoder.go:17`, enforced at `decoder.go:241`) and `defaultMaxHeaderListSize = 64 KB` per header list (`decoder.go:12`, enforced at `decoder.go:94`). `WithMaxHeaderListSize` appears only in tests, but it only *raises* the decoder cap — omitting it leaves the 64 KB default in force, not "unlimited". (`connstate.go:28`'s `defaultMaxHeaderListSize = 0` is the SETTINGS value advertised to the peer — a different thing.) gRPC-Web over HTTP/1.x is bounded tighter still at `maxRequestLineSize = 8192` (`parser/parser.go:29`).

  Worst case is therefore ~11 preset rules × (one ≤64 KB builder + one ≤64 KB `[]byte` copy) ≈ 1.4 MB transient allocation and ~700 KB of linear scanning per gRPC Start — which is exactly what `rules/http` already does on every HTTP/2 request through this proxy. **No follow-up HTTP/2 header-size-cap Issue is needed; the cap exists and is 16× tighter than assumed.**

**3. Response-side degenerate URL — claim verified.**
`buildStartMessage` (`layer/grpc/channel.go:1089-1095`) sets the four fields only under `if dir == envelope.Send`, and `grpcweb.applyRequestOverlay` (`layer/grpcweb/channel.go:497-503`) early-returns on `dir != envelope.Send`. A Receive Start has all four empty, `reconstructURL` returns `""`, and `checkRule` skips it. Doubly guarded by `SafetyStep.Process`'s `if env.Direction != envelope.Send` (`safety_step.go:63`); `SafetyStep` is the only production caller of `grpcrules.SafetyEngine.CheckInput`. Separately, a *Send* Start with `:scheme` set and `:authority` omitted (legal per RFC 9113 §8.3.1 when `Host` is used) does yield `https:///pkg.Svc/Method` — harmless, cannot cause a false negative against any preset pattern, byte-identical to `rules/http.reconstructURL`, and pinned by a test.

**4. Credential leakage into logs — assessed, not a finding.**
`Violation.Match` is the regex match, not the whole URL. Against the 11 preset patterns (all anchored on SQL/OS keywords), a query like `?token=SECRET&q=DROP+TABLE+users` yields `Match = \"DROP TABLE \"` — the credential is outside the match. No new exposure class either: the gRPC engine already logged `metadata`-target matches (which cover `Authorization`), and `rules/http` already logs url-target matches.

**5. Typed-nil dereference — preconditions absent (MITM Principle 6).**
`case common.TargetQuery: return m.RawQuery` derefs `m` without the nil guard `reconstructURL` carries, and `CheckInput`'s `msg == nil` check does not catch a typed-nil interface. But the three pre-existing arms have the same shape, and I could not construct a reachable producer: `layer/grpc.buildStartMessage`, `layer/grpcweb`, `mcp.buildResendGRPCStartEnvelope` and `mcp/grpc_schema_discover` all allocate a struct, and `grpcweb.Send` explicitly rejects `env.Message == nil`. Not a finding. Re-open trigger: a producer that can place a typed-nil `*envelope.GRPCStartMessage` in `env.Message`.

### LOW findings

| S-1 | LOW | internal/pipeline/safety_step.go | 163-168 | GoSecurity | CWE-532 | `Violation.Match` is emitted verbatim into an Info-level slog field with no length bound. `destructive-sql:update-all` contains `.*WHERE`, so one match can span most of the scanned string (≤64 KB here). **Pre-existing and strictly worse on the HTTP `body` target, where it is unbounded.** Not introduced by this PR. Follow-up: truncate at the log seam across all four protocol arms. |
| S-2 | LOW | internal/layer/grpcweb/channel.go | 497-503 | InputValidation | CWE-693 | gRPC-Web never populates `GRPCStartMessage.RawQuery` — the identifier does not appear anywhere in `internal/layer/grpcweb/`. `applyRequestOverlay` copies Authority/Scheme/Path but not RawQuery, so for gRPC-Web the new `query` target is always empty and `reconstructURL` omits the query. **Confirmed not exploitable**: the gRPC-Web Send path rebuilds the request as `msg.Path = \"/\" + start.Service + \"/\" + start.Method` (`channel.go:1097`) with no RawQuery, and the HTTP/1 encoder builds the request-URI from `msg.Path + \"?\" + msg.RawQuery` (`channel.go:1301`, `wireencode.go:148`) — so the client's query is dropped before reaching upstream. The unscanned payload is also unforwarded. |
| S-3 | LOW | internal/rules/common/preset.go | 77 | InputValidation | CWE-436 | New false-positive surface for gRPC: `destructive-os:mkfs` is `mkfs[.\s]`, which matches any service in a package named `mkfs` — `:path = /mkfs.ConfigService/Get` now matches the `url` target and the RPC is dropped. Novel to gRPC (before this PR neither `service` nor `url` was a preset target), but inherent to regex path scanning, identical to existing HTTP behaviour, and the filter is opt-in. |

### Recommendation

Merge as-is. S-1 and S-2 are worth follow-up Issues; S-3 is worth a documentation line rather than a code change.

S-2 also surfaces an adjacent, out-of-scope defect worth its own Issue: the gRPC-Web upstream leg **destroys the client's query string** (rebuilt from `Service`/`Method` only), which is a wire-fidelity gap in the USK-1053 / USK-1057 class independent of safety scanning.

Local `make vet` and `make lint` (gocyclo included) pass; PR CI is fully green including CodeQL.

---

usk6666 (commented):
## Code Review: APPROVED ✅

Correct, narrowly-scoped, and genuinely scan-only. I independently re-verified the load-bearing claims rather than taking them from the PR body: the two new `case` arms cannot write back (everything they touch is a `string` or a fresh `[]byte(data)` copy, so `Pattern.Find` never aliases message memory); the `Service`/`Method` fallback provably mirrors `layer/grpc.pathForStart` for the `Path == ""` shape that `resend_grpc_helpers.go:1131` actually produces; the response side is doubly guarded (`buildStartMessage` gates all six fields on `dir == envelope.Send`, and `SafetyStep.Process` early-returns on non-Send — and `processGRPC` is the sole production call site of `grpc.SafetyEngine.CheckInput`); and the USK-702 recheck the inline USK-1078 note describes is real (`intercept_step.go:242` runs `s.safety.Process` on the `applyGRPCStartModify` clone, which preserves `Path` while overriding `Service`/`Method`), with USK-1078 filed in Backlog. Findings are 2 LOW + 3 NIT, none blocking.

### Verification performed

- **Mutation test on the fix** — deleting the two new `case` arms fails exactly the 7 claimed top-level tests (`..._DoesNotMutateStartMessage_OnMatch`, `..._OnStartPath`, `..._OnStartQuery`, `TestSafetyEngine_TargetQuery_OnStart`, `TestSafetyEngine_TargetURL_OnStart`, `..._CheckInputAll_StartURLQuery_...`, `TestExtractTarget_StartURLQuery`). The three absence-asserting tests correctly stay green. Claim confirmed, not taken on trust.
- **Mutation test on the fallback** — disabling the `path == "" && (Service != "" || Method != "")` branch fails `TestExtractTarget_StartURLQuery` + `TestReconstructURL_GRPCStart`. The fallback is not untested surface.
- `golangci-lint run ./internal/rules/...` (pinned v2.13.2) → **0 issues**, gocyclo included. `gofmt -l internal/rules/grpc/` clean.
- `make test`: `internal/rules/grpc` ok. `internal/flow` and `internal/mcp` hit `panic: test timed out after 4m0s` in this sandbox — reproducible standalone, both packages untouched by this diff, so environment slowness under `-race`, not a regression here.

### Findings

| F-1 | LOW | `internal/rules/grpc/doc.go` / `safety.go` | 20-22 / 180-181 | GoConventions | Both summaries describe `TargetURL` as the `:path`; it is actually the full URL including scheme and authority |
| F-2 | LOW | `internal/rules/grpc/safety.go` | 210-213 | Correctness | The gRPC-Web query is invisible to both new targets, and the gap is recorded only in the PR body — no code note, no Issue |
| F-3 | NIT | `internal/rules/grpc/safety_test.go` | 301-315 | TestQuality | `startFieldSnapshot` covers 7 of `GRPCStartMessage`'s 11 fields |
| F-4 | NIT | `internal/rules/grpc/helpers.go` | 80-94 | Correctness | `reconstructURL` diverges from `pathForStart` in the all-three-empty shape |
| F-5 | NIT | `internal/rules/grpc/safety_test.go` | (`TestExtractTarget_URLQueryNotApplicableToDataOrEnd`) | TestQuality | Map iteration without `t.Run` subtests |

### Suggestions

**F-1 — `TargetURL` is the full URL, but two of the three docs say `:path`.** `doc.go` says the shared targets "map to the request-side :path (with its query rejoined)"; `extractTarget`'s godoc says they "read the request-side :path and its query". The scanned string is `scheme://authority + path + ?query` — the PR's own `TestSafetyEngine_TargetURL_OnStart` proves it by matching `https://internal\.admin\.example\.com/`. `helpers.go`'s `reconstructURL` godoc is accurate; only these two summaries are not. This is exactly the kind of imprecision that produces a rule that silently never fires: an operator who reads either line and writes a path-anchored `^/pkg\.Svc/` gets zero matches forever, with no error. Suggest both read something like "the full reconstructed request URL (`scheme://authority` + `:path` + query)".

**F-2 — gRPC-Web query is unscanned, and nothing in the tree says so.** `internal/layer/grpcweb/channel.go:496` `applyRequestOverlay` copies `Authority` / `Scheme` / `Path` but not `RawQuery` (`RawQuery` appears nowhere in `internal/layer/grpcweb`). Concretely: a gRPC-Web `POST /pkg.Svc/Method?q=DROP TABLE users` has `start.RawQuery == ""`, so the `query` target returns `""` and `reconstructURL` omits the query — both new targets miss it, while the byte-equivalent native-gRPC request is blocked. Nothing in the diff is wrong; the issue is bookkeeping. USK-1078 got a filed Issue *and* an inline reference; this residual got a PR-body paragraph that disappears from the codebase on merge. Suggest filing a follow-up (the fix is plausibly `start.RawQuery = msg.RawQuery`, but it belongs in a grpcweb-scoped PR with its own overlay tests, since `applyRequestOverlay`'s godoc reasons carefully about which producers populate what) and referencing it from the new arm alongside the USK-1078 note.

**F-3 — widen the non-mutation snapshot with a whole-struct backstop.** `startFieldSnapshot` covers `Metadata` / `Scheme` / `Authority` / `Path` / `RawQuery` / `Service` / `Method` / `env.Raw`, which is exactly the read surface — the godoc is honest about that. But `safety.go`'s contract is broader ("MUST NOT mutate ... or any other message field"), and `Timeout` / `ContentType` / `Encoding` / `AcceptEncoding` / `Anomalies` would not be caught. `GRPCStartMessage.CloneMessage()` already exists, so `want := msg.CloneMessage()` + a `reflect.DeepEqual(want, msg)` assertion is a two-line backstop that covers all 11 fields and auto-extends when a field is added. Keep the per-field `t.Errorf`s for readable diagnostics; add DeepEqual underneath them.

**F-4 — one shape where `reconstructURL` does not mirror `pathForStart`.** When `Path`, `Service` and `Method` are all empty, `pathForStart` → `buildGRPCPath("", "")` → `"/"`, while `reconstructURL` leaves the path component empty and scans `https://a` where the wire carries `https://a/`. It is reachable (a client sending `:path: ""`; `GRPCStartMessage`'s own godoc records that `":path: \"\""` is re-emitted as `/`) but has no security consequence — a rule would have to hinge on a bare trailing slash. The only thing slightly off is the comment, which claims a `pathForStart` mirror. Either emit `"/"` in that case, or narrow the comment to `projectGRPCStart` (`record_step.go:1301`), which the branch matches byte-for-byte.

**F-5 — `TestExtractTarget_URLQueryNotApplicableToDataOrEnd` iterates a map without subtests.** Map order is randomized, so failures interleave and are not reproducible via `-run`. The `kind` is in the message so it stays diagnosable, but a `t.Run(kind+"/"+string(target), ...)` (or a slice of structs) would match the table-driven pattern the rest of the file uses.

### Out of scope, but surfaced

`safety_filter.input.action: "log_only"` is not honoured on the live wire path: `common.CompiledRule` carries no `Action` field, `compileCustomInputRule` never propagates one, and `SafetyStep` returns `Drop` / `Respond` on any violation. Pre-existing and identical for HTTP and WS, so not a defect in this PR — but this PR widens the set of gRPC traffic subject to unconditional blocking under a config an operator may believe is log-only. Worth its own Issue.

---

PR #132, opened 2026-09-16
fix(pipeline): derive the HostScopeStep scheme from the connection's TLS state

usk6666:
## Summary

`HostScopeStep` handed `TargetScope.CheckTarget` a **blank scheme**, on the reasoning stated in the comment it replaced — that a connection-level check "only validates hostname and port". `matchTargetRule` is AND logic, so every rule carrying a `Schemes` condition silently stopped matching at that Step.

- **Allow side — live bug.** A `Schemes`-bearing allow rule never matched, so `CheckTarget` blocked the connection (`Respond` + synthetic 403 for an `HTTPMessage`, silent `Drop` otherwise). The allow rule published **verbatim** in `internal/mcp/resources/help_security.md` is `Schemes`-bearing, so an operator who copied our own documentation got a 403 on every request. It trips step 4 ("not in agent allow list"), since `set_target_scope` writes the Agent layer.
- **Deny side.** A `Schemes`-bearing deny rule never matched here. HTTP is covered downstream by `HTTPScopeStep`, so for HTTP this was defence-in-depth.

The scheme is now derived from `Envelope.Context.TLS`: non-nil ⇒ `https`, nil ⇒ `http` (new unexported `hostScopeScheme`). `TargetRule.Schemes` means **transport confidentiality** — plaintext versus TLS, spelled with the `http`/`https` tokens — not the L7 application protocol. That is already the meaning `connectPolicyAllow` (`"https"`) and `http1_forward_handler` (`"http"`) encode with their literals. The derivation reads no `Message`, so the Step keeps its protocol-agnostic contract.

`Process` now also **early-returns on `env.Direction != envelope.Send`** — see the review-round commit below for why this is load-bearing, not a tidy-up.

Also: corrected the now-wrong call-site comment, the type godoc, an acceptance record in RFC-001, and a cross-reference comment in `internal/connector/socks5.go` recording exactly how much of the handshake-time blank-scheme gap the post-handshake path closes, and the two routes where it does not.

## Commits

1. `03e98481` — derive the scheme from the connection's TLS state.
2. `5f4a1427` — review round 1: gate on the Send direction, and correct the scope docs.

## Why the Direction gate (review round 1, HIGH)

Before this PR, both legs of a connection always derived the identical scheme (`""`), so `HostScopeStep` could not disagree with itself across directions. Deriving from the **per-Layer** `Context.TLS` made per-direction divergence reachable for the first time — and the losing leg's verdict was mishandled:

- `*HTTPMessage` → `Respond`, but `session.upstreamToClient` branches only on `pipeline.Drop`, so the block fell through to `client.Send` and the upstream response was relayed anyway. `onDrop` also fires only under `Drop`, so **no audit Stream was written either** — fail-open and silent.
- Non-HTTP → `Drop` → inbound frames silently discarded, stranding an established session one-way. Reachable for SSE (`sse.Wrap` copies the upstream Receive envelope's `Context`, which carries nil TLS) and gRPC (Receive events come straight off the upstream HTTP/2 Layer).

Gating on `Send` restores the invariant that a connection-level gate yields one verdict per connection, and **loses nothing**: both legs carry the *same* `TargetHost` (`stack_builder.go` and `stack_builder_target_override.go` set `TargetHost: target` on both contexts; only `TLS` differs), so the Receive-leg check was already redundant on hostname and port. The only thing it added was the scheme axis — exactly the axis that diverges. `HTTPScopeStep.Process` already does the same early-return.

`Send Direction = iota` is `0`, so existing cases that omit `Direction` keep exercising the Send path unchanged.

## Acceptance note

**This changes live proxy behaviour for every operator who has `schemes`-bearing scope rules.** Rules without a `schemes` condition are completely unaffected — `len(rule.Schemes) == 0` short-circuits the scheme test.

- **Deny rules newly take effect.** A `schemes`-bearing deny that matched nothing at this Step will now block matching traffic. That is the authored intent, but it is a live behaviour change operators should hear about before upgrading.
- **Allow rules become more permissive — but only to exactly what the operator authored.** A `schemes`-bearing allow that previously matched nothing (and therefore blocked everything) can now match.

One documented counterexample to "nothing that works today stops working": a rule whose `Schemes` list contains the **empty string** did match the old blank scheme, because `containsStringFold` is `strings.EqualFold` and `EqualFold("", "")` is true. For a deny rule that is a fail-open reversal. It is reachable only because the **Policy layer is unvalidated** (`config.LoadPolicyFile` only unmarshals; `mcpserver.ConvertTargetRules` copies `Schemes` verbatim), while `validateTargetRules` guards the Agent layer. Tracked as USK-1084.

Deliberately **not** changed, recorded in RFC-001 §11 with re-open triggers:

- `matchTargetRule`'s blank-scheme semantics. For an unknown scheme, deny wants "matches everything" and allow wants "matches nothing", so any *single* rule is fail-open in exactly one direction.
- `internal/connector/socks5.go`'s own inline check — at handshake time the tunneled protocol is genuinely unknown (`REP=0x02` must precede any tunnel byte).
- `connect_handler.go`'s hardcoded `"https"` for a CONNECT tunnel that may carry plaintext. This PR makes the downstream answer correct regardless.

## Known limitations (surfaced by this work, not introduced by it)

- **Raw TCP is ungated at the Pipeline level.** `bytechunk` envelopes carry an empty `EnvelopeContext` — `internal/layer/bytechunk/channel.go` stamps only `ReceivedAt`, and `bytechunk.New` has no envelope-context parameter — so `HostScopeStep` short-circuits on `TargetHost == ""` for the raw route, both before and after this change. **USK-1083.**
- **TLS passthrough never reaches the Pipeline at all.** `socks5Passthrough` runs `runPassthroughRelay` before any stack is built, so no `ConnectionStack` exists. The only scope-shaped evaluation there feeds `flow.RecordScope` capture filtering, not `HostScopeStep`.
- **`HTTPScopeStep` passes `msg.Scheme` verbatim**, so an HTTP/2 request that omits `:scheme` (recorded with no anomaly outside extended CONNECT) reaches it blank — the same defect class, one Step over. **USK-1086.**

## Test plan

- [x] `TestHostScopeStep_SchemeDerivedFromTLS` — 13 table cases, every one driving the real `HostScopeStep.Process`; no terminal state constructed directly. Covers the `help_security.md` allow rule, an `http` deny against a plaintext envelope, both negative controls (https-only allow must still block plaintext; https-only deny must not fire on plaintext), `http` allow / `https` deny, case-insensitive tokens, rules with no `Schemes`, bare hostname without a port, the Agent layer lists, and the `HTTPMessage` 403 vs non-HTTP `Drop` terminator split. `Continue` cases assert the whole zero `Result`.
- [x] `TestHostScopeStep_SchemeDerivation_ShortCircuits` — nil scope and empty `TargetHost` continue regardless of TLS state.
- [x] `TestHostScopeStep_DirectionGate` — 5 cases: Send+TLS ⇒ Continue; Receive+`*HTTPMessage` ⇒ Continue (was `Respond`, discarded by `upstreamToClient`); Receive+`*SSEMessage` ⇒ Continue (was a silent `Drop`); Receive under an outright hostname deny ⇒ Continue (the gate is unconditional, not verdict-dependent); Send under the same deny ⇒ Respond (control, proving the deny genuinely fires).
- [x] **Verified failing on `main`**: 8 of the 13 derivation cases fail pre-fix. The 5 that pass are the negative controls and no-`Schemes` baselines, each labelled as such.
- [x] **Mutation check, re-verified after the Direction gate**: `hostScopeScheme` → `""` kills 9; always-`https` kills 6; always-`http` kills 6; deleting the Direction gate kills 3. The 8 pre-existing kills are byte-identical to the pre-gate run, so no case short-circuits on Direction before reaching the derivation.
- [x] No e2e or integration test in the tree configures a `Schemes`-bearing scope rule, so no integration behaviour legitimately changes. Nothing was weakened.
- [x] `make lint` (gofmt + golangci-lint v2.13.2) 0 issues; `make build`; `make test`; `make test-e2e-smoke`.

Resolves USK-1081
Linear: https://linear.app/usk6666/issue/USK-1081

https://claude.ai/code/session_01XbQ5WujVS8rms5dBvkSYVH


usk6666 (commented):
## Code Review: CHANGES REQUESTED ❌ → 修正済み (round 1, commit 5f4a1427)

導出そのもの、その配置、`Message` 非依存契約、回帰スイートはいずれも妥当。実装者の「main で 13 中 8 が落ちる / 変異 3 種を殺す」という主張もケース単位で独立に確認済み。

**ブロッカーは、この変更が `HostScopeStep` に「1 本の接続の 2 つの leg で異なる判定に到達しうる」状態を初めて作ったこと**、そして負けた側の判定が HTTP では無言で破棄され、非 HTTP ではセッションが片側開放のままハングすること。いずれも未文書・未テストだった。

### Findings

| ID | Severity | File | Category | 内容 |
|----|----------|------|----------|------|
| F-1 | **HIGH** | `host_scope_step.go` | Correctness | per-leg 乖離が初到達可能に。`*HTTPMessage` → `Respond` だが `upstreamToClient` は `Drop` しか分岐せず `client.Send` へ素通り(`onDrop` も `Drop` 限定なので**監査記録も残らない**)。非 HTTP → 無言 `Drop`。main では同じ設定がリクエスト時点で綺麗に 403 を返していた |
| F-2 | MEDIUM | `socks5.go` | CodeHealth | 「The residue is the bytechunk route」が網羅的主張として不完全。`socks5Passthrough` → `runPassthroughRelay` は `ConnectionStack` を構築せず Pipeline に到達しない — **2 つ目の residue**。「http1 / h2 stacks」の列挙も post-Upgrade の ws/sse/grpc を過小評価 |
| F-3 | LOW | `host_scope_step_test.go` | TestQuality | raw deny ケースのコメントが live coverage を主張しているが、production raw envelope は `TargetHost` を持たず短絡する。同一コミット内の `socks5.go` コメントとも矛盾。USK-1083 の greppable アンカーも不在 |
| F-4 | LOW | `docs/rfc/envelope.md` | Architecture | 「No configuration that works today stops working」が直前の文と矛盾。deny の match 集合が増えることは、今流れているトラフィックが止まるということ |
| F-5 | NIT | `docs/rfc/envelope.md` | Architecture | re-open trigger 付きの deferral が §3.5.1 にインライン。RFC の慣習(USK-900 / USK-1016 / USK-1053)では §11 の `- [ ]` bullet |
| F-6 | NIT | `docs/rfc/envelope.md` | CodeHealth | `help_security.md:58` の例は `set_target_scope` = Agent 層なので `CheckTarget` **ステップ 4**、ステップ 3 ではない |

### Security Review: APPROVED ✅(LOW 2)

`Envelope.Context.TLS` は **leg ごとに stack 構築時、完了したハンドシェイクからただ一度だけ**書かれる。`internal/pluginv2/` / `internal/rules/` / `internal/pipeline/` を横断しても read しか無く、runtime の書き込み経路は存在しない。`PluginStepPre` は `HostScopeStep` より後段で、intercept の `modify_and_forward` 解放も scope Step を再実行しない。クライアントは CONNECT + inner `0x16` で leg を TLS に**選ぶ**ことはできるが、そのとき leg は実際に暗号化されている — ルール自身の述語であって迂回ではない。

blank → derived の変更は `matchTargetRule` 上で**単調**であることを証明済み(`len(rule.Schemes) == 0` は短絡し、`hostScopeScheme` は `"http"` / `"https"` しか返さない)。deny はマッチが増えるだけ、allow はオペレータが記述した範囲ちょうどに収束する。

* **S-1 (LOW)** — 単調性の唯一の例外: `Schemes` に**空文字**を含むルール。`EqualFold("", "")` が true なので旧 blank scheme にマッチしていた。deny なら fail-open に反転。Policy 層が未検証(`LoadPolicyFile` / `ConvertTargetRules`)なので到達可能 → **USK-1084**
* **S-2 (LOW)** — per-leg 非対称。F-1 の Direction ゲートで解消

---
Automated review by yorishiro-proxy Code Review + Security Review Agents

